PR #12, opened 2020-02-04
Fixed issue with publicPath

scottrumsby:
Fixed issue with sass import that prevented app from working when running not off of the server root and reconfigured the .env files to hold the API_URLs.  Re-committed public/config/configuration.json and set the app's public path to /namerequest in all environments


LJTrent:
Did not work.

PR #38, opened 2020-04-21
entity 3427 & entity 3327

scottrumsby:
##### entity 3427 - Moved dev dependencies out of regular dependencies
- Multiple packages required only for development were installed as regular dependencies in error.  moved them in package.json appropriately.

##### Entity 3327 - Added language checkbox to landing page
- Added checkbox for person's name type search to specify name is English name and logic to route such requests to auto-analysis
- refactored store logic for initiating name analysis, GET to name-analysis and name sanitizing


LJTrent:
The English checkbox should say English,. also, when you hover over it we wanted a modal (blurb) like you have for entity types , and a person's name that says: 'The name request is assumed to be English if the box is checked."

PR #48, opened 2020-05-19
Entity 3520

scottrumsby:
#### Entity 3520
- Wrote front end logic to POST to /namerequests for DRAGT, COND-RESERVE and RESERVED situations with support for basic case o
f BC Corporation entity type.
- also fixed PR-#46 issue where configuration.json was not consistantly loading the API url into App causing the need to refresh
- changed the way configuration.json settings are loaded so they are consistantly and reliably applied at launch.  also fixed 
debouncing on call to canadapost address search API.


scottrumsby:
requested changes are made.  have tested COND_RESRVE with conflict, consent words, and conflict + consent words, as well as RESERVED and DRAFT.  everything seems to work now. git checks have passed.  this should be good to go.

PR #50, opened 2020-05-28
entity-3756: fix the response search bar

scottrumsby:
fixed issue with caret jumping around during editing of search term and fixed issue where clicking on the search bar did not result in caret going to clicked spot.


LJTrent:
Look for a new solution.

PR #76, opened 2020-08-19
3416 Affiliation/draft business integration

cameron-freshworks:
*Issue #:* /bcgov/entity#3416

*Description of changes:*

* Implements 2 common components to enable SignIn with Sbc Auth

* Once authenticated, allows for an affiliation request to be made to link an NR to an Account.

* Once affiliated, allows for a request to Legal Api to create a draft business (temp reg #) to begin the flow of an 
Incorporation Application.

* Minor error handling

* Updates configuration file for URLS 

** Notes: ** 

* A dependency for this work is to update the `config_map` in openshift with the environment variables seen in the configuation file as per environment.
* Note: this has been actioned and Lorna will be updating these in OS.

* Another update will be required to the keycloak.json in openshift for Test/Prod env's. Will touch base with Sumesh on this, to sort out all the details of work.
* Note: We have discussed this with Sumesh and have the proper `keyloak.json` file. Lorna will handle the update in OS.

* A temporary `Incorporate Now` button has been implemented for testing purposes only. This button has no logic behind when it is shown and should be removed before promoting this work beyond a DEV environment. 
* Note: This will remain in for testing purposes as per the teams request.


client/package.json
```diff
@@ -17,7 +17,7 @@
     "material-design-icons-iconfont": "^5.0.1",
     "moment": "^2.27.0",
     "remove-accents": "^0.4.2",
-    "sbc-common-components": "^2.2.1",
+    "sbc-common-components": "^2.3.1",
```
nitheesh-aot, line 20: please update the version to latest: 2.3.2
which handles the url issues that we had few days ago

cameron-freshworks: Absolutely, will do. Thanks!

Will update in next commit.

severinbeauvais: It's now 2.3.3 (if you want to use the version info tooltip in the footer).

client/public/config/configuration.json
```diff
@@ -1,6 +1,10 @@
 [
   {
     "URL":"https://namex-dev.pathfinder.gov.bc.ca",
-    "PAYMENT_PORTAL_URL": "https://dev.bcregistry.ca/business/auth/makepayment"
+    "PAYMENT_PORTAL_URL": "https://dev.bcregistry.ca/business/auth/makepayment",
+    "BUSINESSES_URL": "https://dev.bcregistry.ca/business/auth/",
+    "KEYCLOAK_CONFIG_PATH": "/local-keycloak-config-url/keycloak.json",
+    "VUE_APP_AUTH_ROOT_API": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1",
+    "VUE_APP_LEGAL_ROOT_API": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1"
```
severinbeauvais, line 8: Why name there VUE_APP_xxx ? It suggests they should be environment variables (in .env file).

cameron-freshworks: Thats a fair point, i was following the naming convention used elsewhere when using an .env. Will refactor these names for simplicity in the next commit!

client/package.json
cameron-freshworks: Can't keep up lol. Will update

client/src/components/modals/incorporate-login.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <v-dialog v-model="showModal" max-width="60%">
+    <v-card class="pa-4">
```
severinbeauvais, line 3: Why use a v-card?

client/src/components/modals/incorporate-login.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <v-dialog v-model="showModal" max-width="60%">
```
severinbeauvais, line 2: You might want to attach this to the parent page for ease of testing.
You might want to make this persistent (ie, can't close when clicking off of it)?

client/public/config/configuration.json
cameron-freshworks: Updated in next commit

client/src/components/modals/incorporate-login.vue
cameron-freshworks: All of the dialogs are mounted in the `App.vue` and to change the pattern would require refactoring all of them. The same goes for the persistent prop. 
In this specific case, the common component we're using within this modal doesn't have a close action and it looked a little out of place when layered on top of the common component.

client/src/main.ts
```diff
@@ -17,6 +18,16 @@ Vue.config.devtools = true
 async function startVue () {
   Vue.prototype.$PAYMENT_PORTAL_URL = await getConfig()
 
+  // configure Keycloak Service
+  await KeyCloakService.setKeycloakConfigUrl(sessionStorage.getItem('KEYCLOAK_CONFIG_PATH'))
+  if (!window.location.pathname.includes('/signin')) {
+    await KeyCloakService.initializeToken(null).then(() => {}).catch(err => {
```
severinbeauvais, line 24: You may not need the "then" function.

client/src/components/modals/incorporate-login.vue
cameron-freshworks: Good catch, it is redundant here! Updated in next commit

client/src/main.ts
```diff
@@ -17,6 +18,16 @@ Vue.config.devtools = true
...
+      if (err?.message !== 'NOT_AUTHENTICATED') {
+        throw err
```
severinbeauvais, line 26: So there's no page reload mechanism here if this fails? (Or user error message.)

cameron-freshworks: The scope of the ticket grew a lot here, so some of the error handling isn't entirely fleshed. Some design work was required here. That being said, Scott did mention we should/could use whatever was in place in our other ui's. 
Implementing the expanded error handling / ie dialogs, was going to come under another ticket.

client/src/router/index.ts
```diff
@@ -9,6 +10,7 @@ const routes = [
     name: 'Landing',
     component: () => import('@/views/landing.vue')
```
severinbeauvais, line 11: I think this is for lazy-loading. We should try this in Filings UI for the maintenance filings pages.

cameron-freshworks: Good idea, what is method referred to as. Dynamic importing? 😆 Just a wild guess

client/src/main.ts
cameron-freshworks: Updated in next commit!

client/src/router/index.ts
```diff
@@ -9,6 +10,7 @@ const routes = [
...
   },
+  { path: '/signin/:idpHint', name: 'signin', component: Signin, props: true, meta: { requiresAuth: false } },
```
severinbeauvais, line 13: In Signin component, you don't actually use the idpHint prop, so you can probably simplify this.

client/src/components/modals/affiliation-error.vue
```diff
@@ -0,0 +1,55 @@
...
+      <v-card-title>
+        <slot name="icon">
+          <v-icon large color="error">error</v-icon>
+        </slot>
+        <span>
+          <slot name="title">Error Adding Name Request</slot>
+        </span>
+      </v-card-title>
+      <v-card-text>
+        <slot name="text">The specified name request has already been affiliated.</slot>
```
severinbeauvais, line 13: This is a hard-coded error message, but surely there can be other errors if "createNRAffiliation" failed?

client/src/components/existing-request/existing-request-display.vue
```diff
@@ -185,6 +186,12 @@ export default class ExistingRequestDisplay extends Vue {
       newReqModule.patchNameRequestsKV(payload)
     }
   }
+
+  /** Open Incorporate Now Login Modal and apply NR Data to Session */
+  activateILModal () {
+    newReqModule.mutateIncorporateLoginModalVisible(true)
+    sessionStorage.setItem('NR_DATA', JSON.stringify(this.nr))
```
severinbeauvais, line 193: I finally found where this data gets used... but please add a comment here indicating this.

client/src/router/index.ts
cameron-freshworks: We use this idpHint to determine the path for login. It is passed from the `sbc-authentication-options`

client/src/components/modals/affiliation-error.vue
cameron-freshworks: There is for sure. As mentioned above, building a more robust means of handling the various error codes from the various requests was some work to still be done. The scope of this ticket grew considerably, so that was something to come under another ticket.

client/src/components/modals/incorporate-login.vue
cameron-freshworks: Sorry, to come back to this, the card was to follow the same design pattern of the other dialogs in this UI. Also, the contents was strictly the common component pulled in, which was just a container.

client/src/components/modals/incorporate-login.vue
severinbeauvais: They're declared in App.vue but they are mounted near the top of the DOM, so testing them in the context of a page/view can be difficult since you can't mount them (Vue test utils only mounts the page/view component, not the entire app).
FWIW and apparently not a problem atm.

client/src/components/modals/incorporate-login.vue
severinbeauvais: Do you still need the padding?

client/src/components/modals/incorporate-login.vue
cameron-freshworks: Absolutely agree, to all the above.

client/src/router/index.ts
severinbeauvais: It is dynamic importing, but I think it's called lazy loading. For real.

client/src/router/index.ts
severinbeauvais: Hmm, the Signin component doesn't appear to use this route data at all.

client/src/components/modals/incorporate-login.vue
cameron-freshworks: I feel yes, the common component doesn't have anything and it looks a little 'tight', you could say.

client/src/router/index.ts
cameron-freshworks: haha awesome :) We should try this! or at least research it a bit more.

client/src/components/existing-request/existing-request-display.vue
cameron-freshworks: Will do 👍

client/src/components/modals/incorporate-login.vue
severinbeauvais: Can you add `content-class` to the v-dialog to get your padding?
I think using a v-card is overkill :>

cameron-freshworks: I could, but every other modal follows this pattern, so for the sake of congruency i would prefer to follow the pattern already established. If you feel like this is a good exception though, i could for sure.

client/src/router/index.ts
cameron-freshworks: I'm not sure it works internally in the SBC Common Components. I believe it passes that hint as a prop to the signin component. Which is defined on line 47 of the SbcSignin common component. I'd have to dig a little more into their code to fully understand.

client/src/components/modals/incorporate-login.vue
severinbeauvais: I don't think it's a big deal either way, but I did notice you lost your "pa-4" class.

client/src/router/index.ts
severinbeauvais: If I understand correctly, `path` specifies the expected (optional?) route param, and then `props: true` automatically converts params to props.
You don't use the props nor pass them to the SbcSignin component so I think this code is not needed... but I could be wrong.
Don't worry too much about this now if this needs to get merged asap.

client/src/components/modals/incorporate-login.vue
cameron-freshworks: 😅 Thanks!

client/src/router/index.ts
severinbeauvais: I just scratched the surface in /bcgov/entity#1890 but now you know something about it too :)

client/src/plugins/getConfig.ts
```diff
@@ -15,6 +15,27 @@ export function getConfig (): Promise<void> {
...
+
+    const authApiUrl = response.data[0]['AUTH_API_URL']
+    sessionStorage.setItem('AUTH_API_URL', authApiUrl)
+    // eslint-disable-next-line no-console
+    console.log('Set Auth URL to: ' + authApiUrl)
+
+    const legalApiUrl = response.data[0]['LEGAL_API_URL']
+    sessionStorage.setItem('LEGAL_API_URL', legalApiUrl)
+    // eslint-disable-next-line no-console
+    console.log('Set Legal URL to: ' + legalApiUrl)
```
LJTrent: remove

client/src/plugins/getConfig.ts
```diff
@@ -15,6 +15,27 @@ export function getConfig (): Promise<void> {
...
+
+    const keycloakConfigPath = response.data[0]['KEYCLOAK_CONFIG_PATH']
+    sessionStorage.setItem('KEYCLOAK_CONFIG_PATH', keycloakConfigPath)
+    // eslint-disable-next-line no-console
+    console.info('Set KeyCloak Config Path to: ' + keycloakConfigPath)
+
+    const businessUrl = response.data[0]['BUSINESSES_URL']
+    sessionStorage.setItem('BUSINESSES_URL', businessUrl)
+    // eslint-disable-next-line no-console
+    console.log('Set Business URL to: ' + businessUrl)
```
LJTrent: remove

client/src/plugins/getConfig.ts
```diff
@@ -15,6 +15,27 @@ export function getConfig (): Promise<void> {
     })
     const baseURL = response.data[0]['URL'] + '/api/v1'
     const paymentURL = response.data[0]['PAYMENT_PORTAL_URL']
+
+    const keycloakConfigPath = response.data[0]['KEYCLOAK_CONFIG_PATH']
+    sessionStorage.setItem('KEYCLOAK_CONFIG_PATH', keycloakConfigPath)
+    // eslint-disable-next-line no-console
+    console.info('Set KeyCloak Config Path to: ' + keycloakConfigPath)
```
LJTrent: remove

client/src/plugins/getConfig.ts
```diff
@@ -15,6 +15,27 @@ export function getConfig (): Promise<void> {
...
+
+    const businessUrl = response.data[0]['BUSINESSES_URL']
+    sessionStorage.setItem('BUSINESSES_URL', businessUrl)
+    // eslint-disable-next-line no-console
+    console.log('Set Business URL to: ' + businessUrl)
+
+    const authApiUrl = response.data[0]['AUTH_API_URL']
+    sessionStorage.setItem('AUTH_API_URL', authApiUrl)
+    // eslint-disable-next-line no-console
+    console.log('Set Auth URL to: ' + authApiUrl)
```
LJTrent: remove

client/vue.config.js
```diff
@@ -28,5 +28,16 @@ module.exports = {
     'vuetify',
     'vuex-module-decorators'
   ],
-  publicPath: '/namerequest'
+  publicPath: '/namerequest',
```
LJTrent, line 31: Hmm.  Not sure about this one.  Seems a little strange.  I will ask Scott Rumsby to review.

client/src/plugins/getConfig.ts
cameron-freshworks: removed in next commit

client/src/plugins/getConfig.ts
cameron-freshworks: removed in next commit

client/src/plugins/getConfig.ts
cameron-freshworks: removed in next commit

client/src/plugins/getConfig.ts
cameron-freshworks: removed in next commit

client/vue.config.js
cameron-freshworks: Sounds good. The proxy is to avoid CORS error when running locally.

PR #92, opened 2020-09-11
feature/4294-Use-Payment-Model

bcdevlucas:
*Issue #:* /bcgov/entity###

*Description of changes:*

### DIFF
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -722,13 +722,17 @@ export class NewRequestModule extends VuexModule {
   get nrNum () {
     const { nr } = this
     let nrNum
-    if (nr) nrNum = nr.nrNum
+    if (nr) {
+      nrNum = nr.nrNum
+    }
     return nrNum
   }
   get nrState () {
     const { nr } = this
     let state
-    if (nr) state = nr.state
+    if (nr) {
+      state = nr.state
+    }
     return state
   }
   get pickEntityTableBC () {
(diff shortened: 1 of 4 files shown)


LJTrent (changes requested):
remove this from the pr

LJTrent (changes requested):
remove all of the chnage sto new-request-module.ts except the required payment stuff

client/src/store/new-request-module.ts
```diff
@@ -1356,10 +1348,37 @@ export class NewRequestModule extends VuexModule {
...
+      } else if (applicants) {
+        this.setApplicantDetails(applicants)
       }
-      response = await axios.put(`/namerequests/${nrNum}`, data, {
+    } catch (error) {
+      // eslint-disable-next-line
+      console.log(error)
+    }
+  }
+  @Action
```
LJTrent, line 1378: Only add these chnages

client/src/store/new-request-module.ts
```diff
@@ -937,9 +933,7 @@ export class NewRequestModule extends VuexModule {
   get nrNames () {
```
LJTrent, line 933: revert these changes please. I think this is what is causing the check mark to display on the wrong line.

client/src/store/new-request-module.ts
bcdevlucas: Bad merge, and a couple differences with my auto-format code style cause i'm using webstorm not vscode. Resolved.

PR #93, opened 2020-09-14
entity 3963

scottrumsby:
- fixed bug which was causing the request_action_cd to be set to 'ASSUMED' innapropriately.  now only sets for entity types XCR, XUL, RLC as it did previously.


scottrumsby:
closing out i/f/o new PR 3911 which has more tickets in it

PR #94, opened 2020-09-14
Entity 4911, 4963

scottrumsby:
##### entity 4911
- AML request_action_cd now allows for CA and IN locations

##### entity 4963
- fixed assignment of request_action_cd = 'ASSUMED' to only apply to entity types XCR, XUL, RLC 

### DIFF
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -697,7 +697,7 @@ export class NewRequestModule extends VuexModule {
       { text: 'Foreign', value: 'IN' },
       { text: 'Help', value: 'INFO' }
     ]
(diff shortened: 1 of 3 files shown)


client/src/store/new-request-module.ts
```diff
@@ -697,7 +697,7 @@ export class NewRequestModule extends VuexModule {
       { text: 'Foreign', value: 'IN' },
       { text: 'Help', value: 'INFO' }
     ]
-    if (['CNV', 'AML', 'MVE'].includes(this.request_action_cd)) {
+    if (['CNV', 'MVE'].includes(this.request_action_cd)) {
```
LJTrent, line 700: what is for ?  why did you take out AMAL?

PR #101, opened 2020-09-22
entity 4529, 5047, 5055

scottrumsby:
#### 4529
- now have a conditions button to show name's decision_text when applicable.  launches conditions modal.

#### 5055
- removed RLC from the possible entity type options when Foreign -> AML is selected

#### 5047
- now submit multiple consent words if there are more than one.
- fixed a bug in earlier commit

### DIFF
--- a/client/src/store/list-data/request-action-mapping.ts
+++ b/client/src/store/list-data/request-action-mapping.ts
@@ -15,7 +15,8 @@ export const bcMapping: RequestActionMappingI = {
 export const xproMapping: RequestActionMappingI = {
   ASSUMED: ['XCR', 'RLC', 'XUL'],
   REN: ['XCR', 'XCP', 'RLC', 'XUL'],
-  REH: ['XCR', 'XCP', 'RLC', 'XUL']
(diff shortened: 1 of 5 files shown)


client/src/store/list-data/request-action-mapping.ts
```diff
@@ -15,7 +15,8 @@ export const bcMapping: RequestActionMappingI = {
 export const xproMapping: RequestActionMappingI = {
   ASSUMED: ['XCR', 'RLC', 'XUL'],
   REN: ['XCR', 'XCP', 'RLC', 'XUL'],
-  REH: ['XCR', 'XCP', 'RLC', 'XUL']
+  REH: ['XCR', 'XCP', 'RLC', 'XUL'],
+  AML: ['XCR', 'XCP', 'XUL', 'XLP', 'XLL', 'XSO']
```
LJTrent, line 19: Sorry. You will also need to update the entity selection screen for CANADA and FOREIGN for these entity types and only allow only for XCR, XCP, XUL and AML request_action.

PR #116, opened 2020-10-05
feature/4986-Back-Out-Of-Payment

bcdevlucas:
*Issue #:* /bcgov/entity###

*Description of changes:*


LJTrent (changes requested):
When the payment gets cancelled on the sbc pay screen.  It backs out nucely but leaves the NR in a reserved state.  The NR must be cancelled so that it is cancelled in postgres and oracle and appear as cancelled when the user returns to the NR Summary.  Leaving it as RESERVED is misleading for the client.

PR #122, opened 2020-10-08
locked package.json to specific versions

scottrumsby:
locked package.json to specific versions known to succesfully build in combination


scottrumsby:
cancelled as cause of build issue was identified and was not a dependency issue

PR #130, opened 2020-10-23
Entity 5319 & 5337

scottrumsby:
##### bcgov/entity#5319
- removed the [Add Word Here] instruction text from the nameChoice input when user opts to Send to Examination

##### bcgov/entity#5337
- Now handling the designation field of the names object (in edit mode) correctly

##

### DIFF
--- a/client/src/components/new-request/submit-request/reserve-submit.vue
+++ b/client/src/components/new-request/submit-request/reserve-submit.vue
@@ -77,8 +77,6 @@ export default class ReserveSubmitButton extends Vue {
     }
     if (['add_descriptive', 'add_distinctive'].includes(newReqModule.currentIssue.issue_type)) {
       if (!newReqModule.showActualInput) {
-        // eslint-disable-next-line
-        console.log('emitted')
         this.$root.$emit('show-original-name')
       }
     }
--- a/client/src/components/new-request/submit-request/reserve-submit.vue
+++ b/client/src/components/new-request/submit-request/reserve-submit.vue
@@ -18,6 +18,15 @@ export default class ReserveSubmitButton extends Vue {
   get entity_type_cd () {
     return newReqModule.entity_type_cd
   }
+  messages () {
+    if (Array.isArray(newReqModule.analysisJSON.issues)) {
+      newReqModule.analysisJSON.issues.map(issue => {
+        if (Array.isArray(issue.name_actions)) {
+          return issue.name_actions.map(action => (action || {}).message)
+        }
+      })
+    }
+  }
   get isAssumedName () {
     return newReqModule.isAssumedName
   }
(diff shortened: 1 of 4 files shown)


severinbeauvais:
Please use this syntax for the references tickets: ` /bcgov/entity#xxxx`

This provides a clickable link above and also adds a ref in the ticket back to this PR.

client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -63,7 +75,13 @@ export default class ReserveSubmitButton extends Vue {
       this.sendToExamination()
       return
     }
-
+    if (['add_descriptive', 'add_distinctive'].includes(newReqModule.currentIssue.issue_type)) {
+      if (!newReqModule.showActualInput) {
+        // eslint-disable-next-line
+        console.log('emitted')
```
severinbeauvais, line 81: This should not be in production code.

client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -18,6 +18,15 @@ export default class ReserveSubmitButton extends Vue {
   get entity_type_cd () {
     return newReqModule.entity_type_cd
   }
+  messages () {
+    if (Array.isArray(newReqModule.analysisJSON.issues)) {
```
cameron-freshworks, line 22: Just curious about this check here. Is this specifically to check if the data type of `newReqModule.analysisJSON.issues` is an Array?
Or is the aim to see if `.issues` is not empty? 

Curious if `Array.isArray()` would still return true if the value provided was an empty array.

severinbeauvais:
Are there any unit tests that need to be written or updated to cover these changes?

severinbeauvais:
Can you please add a reviewer who knows something about how this UI works (architecture, functionality, flow)?

client/src/components/new-request/submit-request/reserve-submit.vue
scottrumsby: My intention was simply to verify it was an array.  `Array.isArray([])` does return true but I think this is the extent of the data checking needed here because I just wanted to be sure it was an Array of any sort before calling .map

cameron-freshworks: Fair enough!

client/src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: Please remove this console log message.

severinbeauvais (approved):
I approve this PR on a technical level.
I cannot speak to the business correctness of it.
I also have some concerns that the code doesn't appear to use typings, and I don't see any unit tests.

After the debugging console log message is removed, this PR can be merged.

cameron-freshworks (approved):

PR #132, opened 2020-10-29
feature/5353-NR-Timeout-Changes

bcdevlucas:
*Issue #:* /bcgov/entity#5353

*Description of changes:*
See commit messages.

### DIFF
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -86,13 +86,13 @@ export default class App extends Mixins(SessionTimerMixin) {
     const { nrId } = newRequestModule
     const componentName = newRequestModule.displayedComponent
     if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
-      newRequestModule.checkinNameRequest().then(() => {})
-    } else if (nrId && ['ExistingRequestEdit'].indexOf(componentName) > -1) {
       // Cancel the NR using the rollback endpoint if we were processing a NEW NR
       // Don't await this request, that way there's no lag, fire it off async and don't block I/O
       // The empty then clause just prevents a linting issue that warns when you don't
       // await an async function which is not an issue, since we don't want to block I/O
       newRequestModule.rollbackNameRequest({ nrId, action: rollbackActions.CANCEL }).then(() => {})
+    } else if (nrId && ['ExistingRequestEdit'].indexOf(componentName) > -1) {
+      newRequestModule.checkinNameRequest().then(() => {})
     }
 
     // Redirect to the start
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -11,7 +11,10 @@
     <NrNotRequired />
     <PickEntityOrConversion />
     <PickRequestType />
-    <PaymentModal :onActivated="onPaymentModalActivated" />
+    <PaymentModal
+      :onActivate="onPaymentModalActivated"
+      :onCancel="onPaymentCancelled"
+    />
     <UpgradeModal />
     <ReapplyModal />
     <PaymentCompleteModal />
@@ -54,9 +57,13 @@ import Header from '@/components/header.vue'
 import TimeoutModal from '@/components/session-timer/timeout-modal.vue'
 import SessionTimerMixin from '@/components/session-timer/session-timer-mixin'
 
-import newRequestModule, { ROLLBACK_ACTIONS as rollbackActions } from '@/store/new-request-module'
-import timerModule from "@/modules/vx-timer"
-import paymentModule from "@/modules/payment"
+import newRequestModule, {
+  NR_COMPLETION_TIMER_NAME,
+  ROLLBACK_ACTIONS as rollbackActions
+} from '@/store/new-request-module'
+
+import timerModule from '@/modules/vx-timer'
+import paymentModule from '@/modules/payment'
 
 @Component({
   components: {
@@ -87,6 +94,18 @@ export default class App extends Mixins(SessionTimerMixin) {
   rollbackOnExpire: boolean
   checkInOnExpire: boolean
 
+  async resetAppState () {
+    // Redirect to the start
+    // Catch any errors, so we don't get errors like:
+    // Avoided redundant navigation to current location: "/"
+    this.$router.replace('/').catch(() => {})
+
+    // Display the tabs
+    await newRequestModule.resetAnalyzeName()
+    await newRequestModule.mutateName('')
+    await newRequestModule.mutateDisplayedComponent('Tabs')
+  }
+
   async onTimerModalExpired () {
     const { nrId } = newRequestModule
     const componentName = newRequestModule.displayedComponent
@@ -100,15 +119,7 @@ export default class App extends Mixins(SessionTimerMixin) {
       newRequestModule.checkinNameRequest().then(() => {})
     }
 
-    // Redirect to the start
-    // Catch any errors, so we don't get errors like:
-    // Avoided redundant navigation to current location: "/"
-    this.$router.replace('/').catch(() => {})
-
-    // Display the tabs
-    await newRequestModule.resetAnalyzeName()
-    await newRequestModule.mutateName('')
-    await newRequestModule.mutateDisplayedComponent('Tabs')
+    await this.resetAppState()
   }
 
   async onTimerModalSessionExtended () {
@@ -122,13 +133,46 @@ export default class App extends Mixins(SessionTimerMixin) {
   }
 
   async onPaymentModalActivated () {
-    timerModule.createAndStartTimer({
-      id: PAYMENT_COMPLETION_TIMER_NAME,
-      expirationFn: () => {
-        paymentModule.togglePaymentModal(false)
-      },
-      timeoutMs: PAYMENT_COMPLETION_TIMEOUT_MS
-    })
+    const { nrId } = newRequestModule
+    const componentName = newRequestModule.displayedComponent
+    // Only do this for New NRs!!!
+    if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
+      // First, clear the NR session timer
+      timerModule.stopTimer(NR_COMPLETION_TIMER_NAME)
+
+      // Start a new timer for the payment
+      timerModule.createAndStartTimer({
+        id: PAYMENT_COMPLETION_TIMER_NAME,
+        expirationFn: () => {
+          const { nrId } = newRequestModule
+          // Cancel the NR using the rollback endpoint if we were processing a NEW NR
+          // Don't await this request, that way there's no lag, fire it off async and don't block I/O
+          // The empty then clause just prevents a linting issue that warns when you don't
+          // await an async function which is not an issue, since we don't want to block I/O
+          newRequestModule.rollbackNameRequest({ nrId, action: rollbackActions.CANCEL }).then(() => {})
+          paymentModule.togglePaymentModal(false)
+        },
+        timeoutMs: PAYMENT_COMPLETION_TIMEOUT_MS
+      })
+    }
+  }
+
+  async onPaymentCancelled () {
+    const { nrId } = newRequestModule
+    const componentName = newRequestModule.displayedComponent
+    // Only do this for New NRs!!!
+    if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
+      // Cancel the NR using the rollback endpoint if we were processing a NEW NR
+      // Don't await this request, that way there's no lag, fire it off async and don't block I/O
+      // The empty then clause just prevents a linting issue that warns when you don't
+      // await an async function which is not an issue, since we don't want to block I/O
+      newRequestModule.rollbackNameRequest({ nrId, action: rollbackActions.CANCEL }).then(() => {})
+
+      // If the payment was cancelled direct the user back to the start
+      await this.resetAppState()
+    }
+
+    paymentModule.togglePaymentModal(false)
   }
 }
 
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -11,7 +11,7 @@
     <NrNotRequired />
     <PickEntityOrConversion />
     <PickRequestType />
-    <PaymentModal />
+    <PaymentModal :onActivated="onPaymentModalActivated" />
     <UpgradeModal />
     <ReapplyModal />
     <PaymentCompleteModal />
@@ -26,7 +26,7 @@
       :displayExpireNowButton="true"
     />
   </v-app>
-</template>df
+</template>
 
 <script lang="ts">
 import Conditions from '@/components/modals/conditions.vue'
@@ -35,7 +35,10 @@ import LocationInfoModal from '@/components/modals/location-info.vue'
 import NrNotRequired from '@/components/modals/nr-not-required.vue'
 import PickEntityOrConversion from '@/components/modals/pick-entity-or-conversion.vue'
 import PickRequestType from '@/components/modals/pick-request-type.vue'
-import PaymentModal from '@/components/payment/payment-modal.vue'
+import PaymentModal, {
+  PAYMENT_COMPLETION_TIMEOUT_MS,
+  PAYMENT_COMPLETION_TIMER_NAME
+} from '@/components/payment/payment-modal.vue'
 import PaymentHistoryModal from '@/components/payment/payment-history-modal.vue'
 import UpgradeModal from '@/components/payment/upgrade-modal.vue'
 import ReapplyModal from '@/components/payment/reapply-modal.vue'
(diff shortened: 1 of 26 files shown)


severinbeauvais:
Could you please update the issue number and add a few details in the description block above?

severinbeauvais:
Also please assign yourself to this PR.

severinbeauvais:
Also could you please rebase your branch to the latest master so that we only see YOUR changes (not 26 commits! haha).

client/src/App.vue
```diff
@@ -87,56 +90,92 @@ import * as types from '@/store/types'
...
+  async resetAppState () {
+    // Redirect to the start
+    // Catch any errors, so we don't get errors like:
+    // Avoided redundant navigation to current location: "/"
+    this.$router.replace('/').catch(() => {})
+
+    // Display the tabs
+    await newRequestModule.resetAnalyzeName()
+    await newRequestModule.mutateName('')
+    await newRequestModule.mutateDisplayedComponent('Tabs')
```
severinbeauvais, line 106: These are not all async methods. (This doesn't hurt but it's misleading.)

Async methods are those that can be blocked on something, like a network call.

Also on line 122 below.

severinbeauvais: (FWIW, we often need async/await in the unit tests to let Vuetify do some things before we verify stuff. In my experience we haven't needed this in the actual code.)

client/src/components/session-timer/session-timer-mixin.ts
```diff
@@ -0,0 +1,42 @@
...
+import * as types from '@/store/types'
+
+@Component
+export default class SessionTimerMixin extends Vue {
+  startNewNrTimer (): Promise<any> {
+    return this.startTimer({
+      id: NR_COMPLETION_TIMER_NAME,
+      expirationFn: () => {
+        this.$store.dispatch(types.SHOW_NR_SESSION_EXPIRY_MODAL)
+          .then(() => {})
```
severinbeauvais, line 21: This is a bit odd. Should this code be waiting for the dispatch call to finish? Or we don't care what happens (ie, empty then function)?

I would expect to see something like this (from onTimerModelSessionExtended):
```
expirationFn: async () => {
   await this.$store.dispatch(types.SHOW_NR_SESSION_EXPIRY_MODAL)
}```

Note that the arrow function is async, and it awaits the dispatch call.

File: client/src/components/session-timer/session-timer-mixin.ts [hunk 3]
```diff
@@ -0,0 +1,42 @@
...
+      },
+      timeoutMs: EXISTING_NR_TIMEOUT_MS
+    })
+  }
+
+  async startTimer (timerConfig): Promise<any> {
+    // Start the user session timer
+    return timerModule.createAndStartTimer(timerConfig)
```
severinbeauvais, line 40: As far as I can tell, createAndStartTimer is not declared as async, so returning it isn't async, and making THIS method async doesn't do anything.

client/src/components/session-timer/countdown-timer-display.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+  <div class="countdown" :class="countdownClass">
+    <div class="countdown-number" :style="countdownSecondsStyle"><small class="countdown-number-text">{{formattedCountdownNumber}}</small></div>
+    <svg>
+      <circle :style="countdownCircleStyle" style="stroke: #003366" r="18" cx="20" cy="20" />
```
severinbeauvais, line 5: This is cool :) but why didn't you use the Vuetify progress component? Eg,
https://vuetifyjs.com/en/components/progress-circular/#rotate

client/src/components/payment/upgrade-modal.vue
```diff
@@ -50,7 +57,24 @@ import { getBaseUrl } from './payment-utils'
...
+  PaymentMixin,
+  PaymentSessionMixin,
+  DisplayedComponentMixin
+) {
+  /**
+   * Optionally display the countdown timer.
+   * This could be turned into a prop for easier configuration.
+   */
+  get displayTimer () {
+    return false
```
severinbeauvais, line 71: Is this just debugging code? If not, why does this exist?

severinbeauvais: Ditto timerName below...?

client/src/components/new-request/main-container.vue
```diff
@@ -5,49 +5,62 @@
...
         <button id="back-to-search-btn"
                 class="modal-activator pa-0 link-std-sans-ul"
-                @click="cancelAnalyzeName">
+                @click="cancelAndResetState">
           <span class="link-std-sans-ul">
             <v-icon class="ma-0 pa-0 mr-n1 mini-back-arrow">arrow_back_ios</v-icon>
             {{ editMode ? 'Return' : 'Start Search Over' }}
           </span>
         </button>
+        <countdown-timer :timerName="timerName" colorString="#1669bb" bgColorString="#efefef" />
```
severinbeauvais, line 14: Our UX/UI designers have become careful about our colour usage, so if possible use named colours (or Material named colours) (or variables from theme.scss) instead of hex colours here. (Same reason we use constants in code.)

client/src/components/new-request/main-container.vue
```diff
@@ -5,49 +5,62 @@
...
-      // Catch any errors, so we don't get errors like:
-      // Avoided redundant navigation to current location: "/"
-      await this.$router.replace('/').catch(() => {})
+
+  @Watch('componentName')
+  async onDisplayedComponentChanged () {
+    const componentName = newReqModule.displayedComponent
+    if (['SubmissionTabs'].indexOf(componentName) > -1) {
+      await this.startNewNrTimer()
+      this.$set(this, 'displayTimer', true)
```
severinbeauvais, line 54: As far as I know, `this.$set` is to add properties to a reactive object.
If I understand correctly, this is not the case here... `displayTimer` is a class property and simply updating it should update the prop in the countdown-timer element.
If I misunderstand, please explain, thanks!

client/src/components/payment/payment-modal.vue
```diff
@@ -39,33 +44,60 @@ import * as paymentActions from './payment-actions'
...
   components: {
     RequestDetails,
-    FeeSummary
+    FeeSummary,
+    CountdownTimer
+  },
+  props: {
+    onActivate: {
+      type: Function,
+      default: async () => undefined
```
severinbeauvais, line 63: Why implement props this way instead of with `@Prop` decorator+method?

client/src/components/payment/upgrade-modal.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <v-dialog max-width="40%" :value="isVisible" persistent>
```
severinbeauvais, line 2: Keep this in mind for future dialogs: you should pass a prop to them to "attach" to. In the parent you would pass a selector to something. This allows a dialog to be attached to the parent, so you can unit test it! (Otherwise it's attached to the DOM root which you can't see in the unit test.)

Also it would be great if components had a id or class, for testing purposes!

client/src/store/new-request-module.ts
```diff
@@ -609,6 +608,12 @@ export class NewRequestModule extends VuexModule {
       homeJurisNum: this.corpNum
     }
   }
+  get currentIssue () {
+    if (this.analysisJSON && this.analysisJSON.issues && Array.isArray(this.analysisJSON.issues)) {
```
severinbeauvais, line 612: Looks OK but a couple of comments:
- you can use TS conditional chaining here, eg. `if (this.analysisJSON?.issues ...`
- let's start adding return types to methods/getters, to make it easier to review and implement new code

severinbeauvais (changes requested):
See comments above.

client/src/App.vue
bcdevlucas: Actions are async in Vue and can be awaited eg. resetAnalyzeName. Probably shouldn't have awaited those mutations.

client/src/components/session-timer/session-timer-mixin.ts
bcdevlucas: We don't care what happens. I am aware of how to use async and ES fat arrow functions as well...
Awaiting the display of that NR session expiry modal blocks I/O and slows things down and I explicitly don't want that to occur. When the session expiry modal displays, a new timer starts for 45 seconds (or whatever that window is set to).

client/src/components/session-timer/session-timer-mixin.ts
bcdevlucas: Again, in Vue, actions are async. createAndStartTimer is an action.

client/src/components/session-timer/countdown-timer-display.vue
bcdevlucas: I wanted something to look specifically the way I wanted to, and to keep it lightweight, CSS only. Stroke is an easy property to work with.

client/src/components/payment/upgrade-modal.vue
bcdevlucas: That just allows you to optionally display timers on the update, receipt, and re-apply pages should you want to. Lorna and I were discussing whether or not we wanted timers on those, but then she went on holiday.

client/src/components/new-request/main-container.vue
bcdevlucas: Busy cranking out fixes, I can update that, just trying to get things working here. I need to do a css clean up in a couple areas related to payments. Noted regarding the colors. The colors I have used are the same ones you use for fonts and buttons.

client/src/components/payment/payment-modal.vue
bcdevlucas: @Prop decorator doesn't appear to cover all use cases that you can do with standard Vue syntax. Also that library used to add those @Prop are a 3rd party library not core Vue functionality IIRC.

client/src/store/new-request-module.ts
bcdevlucas: Merge fragment. Not sure why git detected that as a change.

client/src/components/new-request/main-container.vue
bcdevlucas: this.$set from within a component allows you set data from within a 'method' function of a Vue component. This would be more clear using standard Vue syntax as opposed to the component decorators.
https://stackoverflow.com/questions/51296654/vuejs-set-data-in-methods-function

client/src/App.vue
severinbeauvais: Are you sure? I see some async actions but not "resetAnalyzeName"...?

client/src/components/session-timer/session-timer-mixin.ts
severinbeauvais: I know you are aware of async and arrow functions. I have seen lots of your code use them! 😃 Sorry, I thought this might be an oversight or incomplete code.

Why bother having a "then" block? I think you can just leave it off. Either way, a comment would explain this and prevent a future developer from scratching their head.

client/src/components/session-timer/session-timer-mixin.ts
severinbeauvais: I'm genuinely trying to understand this.

Ref: https://vuex.vuejs.org/guide/actions.html#dispatching-actions-in-components

It looks like actions _can_ be async, but it looks like this one isn't.
```
export const createAndStartTimer = ({ commit }, opts: TimerOptions) => {
  commit(types.CREATE_AND_START_TIMER, opts)
}
```

Let me know if I misunderstand.

client/src/components/session-timer/countdown-timer-display.vue
severinbeauvais: Sounds good.

client/src/components/payment/upgrade-modal.vue
severinbeauvais: Where is the conditional? As far as I can tell, this always returns false.

How about adding a comment indicating this is for possible future use?

client/src/components/new-request/main-container.vue
severinbeauvais: It's OK for now.

Note that we have tickets asking to update some of those text/font/button/icon styles.

So it will be a game of catch-up. Later!

client/src/components/new-request/main-container.vue
severinbeauvais: Thanks for the explanation. That SO post is specifically for adding/changing an object's properties, as explained here:
https://www.telerik.com/blogs/so-what-actually-is-vue-set

As for an object watcher, it only fires when the object itself is reassigned. To watch for object property changes you have to use this:
```
@Watch('myObject', { deep: true })
```

Leave it for now but I think it's tech debt related to mixed use of object style vs component style. AFAIK the Entity UIs have not needed to use `this.$set` for any reason.

client/src/components/payment/payment-modal.vue
severinbeauvais: I'm curious about that but leave it for now.

BTW, you already import that library to provide `@Prop`... it's "vue-property-decorator'".

client/src/store/new-request-module.ts
severinbeauvais: OK. This is another reason why a rebase would have been great before review :)

client/src/App.vue
bcdevlucas: Whether you label them as async or not, an action is dispatched, and as result is inherently async.
Typescript / linting doesn't pick up on it...
https://medium.com/@bogna.ka/everyday-life-in-vue-world-vuex-actions-asynchronous-nature-ad154d246e4b

client/src/components/payment/upgrade-modal.vue
severinbeauvais: I created a new tech debt ticket for this:
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5492

client/src/components/session-timer/session-timer-mixin.ts
bcdevlucas: The then block prevents a linting issue with the current settings that does not allow an unawaited or then'ed call on an async function. I don't want to await those calls, so I just added the empty then block to get around the warning... I left notes in the code in most of the places I did that: 
```
// Don't await this request, that way there's no lag, fire it off async and don't block I/O	      // Don't await this request, that way there's no lag, fire it off async and don't block I/O
      // The empty then clause just prevents a linting issue that warns when you don't	      // The empty then clause just prevents a linting issue that warns when you don't
      // await an async function which is not an issue, since we don't want to block I/O
```

client/src/components/session-timer/session-timer-mixin.ts
bcdevlucas: Dispatching an action always returns a promise. I guess I didn't declare the actions async when i did the VX timer store module, but yes, dispatching those will also be async and returns a promise. Vuex does it under the hood when the action is registered.
```
export const createAndStartTimer = ({ commit }, opts: TimerOptions)
```

client/src/App.vue
severinbeauvais: Oh wow. Thanks for the link. It would be nice if Vue fixed their docs then!!!

OK, as a **future** exercise, action method declarations should be updated to return an explicit promise type, so a developer doesn't have to guess whether the return type is void, any, or async!

client/src/components/session-timer/session-timer-mixin.ts
severinbeauvais: OK, cool.

client/src/components/session-timer/session-timer-mixin.ts
severinbeauvais: It's weird Vue-ism and new to me.

This will all be a lot easier to sort through when types are implement globally (in another ticket).

OK for now.

severinbeauvais (approved):

severinbeauvais:
Please wait for Cam's review before merging.

client/src/components/payment/payment-modal.vue
```diff
@@ -1,34 +1,39 @@
...
-        <v-btn @click="hideModal" id="payment-close-btn" class="normal" text>Cancel</v-btn>
+        <v-btn @click='confirmPayment' id='payment-pay-btn' class='primary' text>Accept</v-btn>
+        <v-btn @click='cancelPayment' id='payment-close-btn' class='normal' text>Cancel</v-btn>
       </v-card-actions>
     </v-card>
   </v-dialog>
 </template>
 
-<script lang="ts">
+<script lang='ts'>
```
cameron-freshworks, line 28: Nothing major but i think within the Vue template itself "" is the standard and elsewhere in this project.

bcdevlucas: True.

cameron-freshworks (approved):
Quite a bit i don't understand but my comments were minor. LGTM, especially if it unblocks you. 

@bcdevlucas Just let me know when your g2g and we'll get er in 👍

bcdevlucas:
@cameron-freshworks Thanks.

PR #134, opened 2020-10-29
5386

scottrumsby:
*Issue #:* /bcgov/entity#5386

*Description of changes:*

- all workflows through name-analysis now end with Send to Examination.  Auto and Conditional Approval disabled.
- updated tests

### DIFF
--- a/client/src/components/common/applicant-info-nav.vue
+++ b/client/src/components/common/applicant-info-nav.vue
@@ -2,7 +2,7 @@
   <v-col cols="5"
          class="text-right">
     <v-btn x-large
-           id="submit-back-btn"
+           :id="`submit-back-btn-${isValid}`"
            class="mr-3"
            v-if="showBack"
            @click="back">
@@ -11,7 +11,7 @@
     <v-btn x-large
            @click="next"
            :disabled="!isValid"
-           id="submit-continue-btn">
+           :id="`submit-continue-btn-${isValid}`">
       {{ nextText }}
     </v-btn>
   </v-col>
--- a/client/src/components/new-request/submit-request/reserve-submit.vue
+++ b/client/src/components/new-request/submit-request/reserve-submit.vue
@@ -46,12 +46,20 @@ export default class ReserveSubmitButton extends Vue {
     if (this.setup === 'cancel') {
       return 'Stop & Send To Examination'
     }
+
+    // delete the next 3 lines when re-enabling auto and conditional approvals
+    if (this.setup !== 'assumed') {
+      return 'Send to Examination'
+    }
+
     if (this.location !== 'BC' && this.setup !== 'assumed') {
       return 'Send for Examination'
     }
     switch (this.setup) {
+      // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
       case 'consent':
         return 'Conditionally Reserve'
+      // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
       case 'examine':
         return 'Send for Examination'
       case 'assumed':
@@ -75,11 +83,14 @@ export default class ReserveSubmitButton extends Vue {
       this.sendToExamination()
       return
     }
-    if (['add_descriptive', 'add_distinctive'].includes(newReqModule.currentIssue.issue_type)) {
-      if (!newReqModule.showActualInput) {
-        this.$root.$emit('show-original-name')
+    if (newReqModule.currentIssue && newReqModule.currentIssue.issue_type) {
+      if (['add_descriptive', 'add_distinctive'].includes(newReqModule.currentIssue.issue_type)) {
+        if (!newReqModule.showActualInput) {
+          this.$root.$emit('show-original-name')
+        }
       }
     }
+
     let goToNames = () => {
       newReqModule.mutateSubmissionType('examination')
       newReqModule.mutateSubmissionTabComponent('NamesCapture')
@@ -90,6 +101,16 @@ export default class ReserveSubmitButton extends Vue {
       goToNames()
       return
     }
+
+    /* the next 4 lines disable auto and condiotional approvals.  all setup types except
+    'assumed' are short-circuited to call goToNames.  The assumed logic can remain in effect
+    since they already go to examination per the existing logic.  to re-enable approvals
+    delete the 4 lines that immediately follow this comment */
+    if (setup !== 'assumed') {
+      goToNames()
+      return
+    }
+
     switch (setup) {
       case 'assumed':
         if (this.$xproMapping['ASSUMED'].includes(entity_type_cd)) {
@@ -99,9 +120,11 @@ export default class ReserveSubmitButton extends Vue {
         newReqModule.mutateAssumedNameOriginal()
         goToNames()
         return
+      // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
       case 'examine':
         goToNames()
         return
+      // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
       case 'consent':
         newReqModule.postNameRequests('conditional')
         newReqModule.mutateSubmissionType('consent')
--- a/client/src/components/new-request/submit-request/reserve-submit.vue
+++ b/client/src/components/new-request/submit-request/reserve-submit.vue
@@ -18,15 +18,6 @@ export default class ReserveSubmitButton extends Vue {
   get entity_type_cd () {
     return newReqModule.entity_type_cd
   }
-  messages () {
-    if (Array.isArray(newReqModule.analysisJSON.issues)) {
-      newReqModule.analysisJSON.issues.map(issue => {
-        if (Array.isArray(issue.name_actions)) {
-          return issue.name_actions.map(action => (action || {}).message)
-        }
-      })
-    }
-  }
   get isAssumedName () {
     return newReqModule.isAssumedName
   }
@@ -83,7 +74,7 @@ export default class ReserveSubmitButton extends Vue {
       this.sendToExamination()
       return
     }
-    if (newReqModule.currentIssue && newReqModule.currentIssue.issue_type) {
+    if (newReqModule.currentIssue?.issue_type) {
       if (['add_descriptive', 'add_distinctive'].includes(newReqModule.currentIssue.issue_type)) {
         if (!newReqModule.showActualInput) {
           this.$root.$emit('show-original-name')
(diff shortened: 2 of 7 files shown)


client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -46,12 +46,20 @@ export default class ReserveSubmitButton extends Vue {
     if (this.setup === 'cancel') {
       return 'Stop & Send To Examination'
     }
+
+    // delete the next 3 lines when re-enabling auto and conditional approvals
+    if (this.setup !== 'assumed') {
```
cameron-freshworks, line 51: With regards to `assumed`, `consent`, `examine` etc, it might be better to store these string values in an enum. Especially if these values are used elsewhere. 
I believe if you reference an enum value in the switch Case, ie `case ApprovalTypes.CONSENT: ...`
it might also sort out the TS errors.

client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -46,12 +46,20 @@ export default class ReserveSubmitButton extends Vue {
...
+      return 'Send to Examination'
+    }
+
     if (this.location !== 'BC' && this.setup !== 'assumed') {
       return 'Send for Examination'
     }
     switch (this.setup) {
+      // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
```
cameron-freshworks, line 59: As mentioned above, there might be a solution to avoid suppressing the errors.

client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -75,11 +83,14 @@ export default class ReserveSubmitButton extends Vue {
       this.sendToExamination()
       return
     }
-    if (['add_descriptive', 'add_distinctive'].includes(newReqModule.currentIssue.issue_type)) {
-      if (!newReqModule.showActualInput) {
-        this.$root.$emit('show-original-name')
+    if (newReqModule.currentIssue && newReqModule.currentIssue.issue_type) {
```
cameron-freshworks, line 86: If the TS version of NameRequest is at least 3.7, you can probably get away with optional chaining here. 
ie `newReqModule.currentIssue?.issue_type`. This would prevent the error `issue_type` doesn't exist on null/undefined etc.

client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -99,9 +120,11 @@ export default class ReserveSubmitButton extends Vue {
         newReqModule.mutateAssumedNameOriginal()
         goToNames()
         return
+      // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
       case 'examine':
```
cameron-freshworks, line 124: Same as above, i think enums could help you out here. Would definitely be a good idea, since these string values are used in a few places. 
If these conditions are use app-wide i would definitely recommend starting an enums directory at @src.

client/tests/unit/applicant-info-1.spec.ts
```diff
@@ -30,6 +30,6 @@ describe('applicant-info-1.vue', () => {
     expect(wrapper.find('#applicant-info-1-form').element).toBeTruthy()
   })
   it('Initially renders the disabled continue button', () => {
-    expect(wrapper.find('#submit-continue-btn').classes().includes('v-btn--disabled')).toBeTruthy()
+    expect(wrapper.find('#submit-continue-btn-false').classes().includes('v-btn--disabled')).toBeTruthy()
```
cameron-freshworks, line 33: Nice! Unit tests :) 
I'm sure this test is fine, but in the past we found `toBeTruthy()` and `toBeFalsey()` could sometimes be unreliable, and moved to `.toBe(true)` and `.toBe(false)`

cameron-freshworks (commented):
Looks good so far! Just had a few questions / comments to resolve 👍

client/src/components/common/applicant-info-nav.vue
```diff
@@ -2,7 +2,7 @@
   <v-col cols="5"
          class="text-right">
     <v-btn x-large
-           id="submit-back-btn"
+           :id="`submit-back-btn-${isValid}`"
```
severinbeauvais, line 5: This breaks the selector in "main.sass" - `#submit-back-btn`.

Also line 14 below

client/src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: The package.json says ^3.5.3 ... which got auto-updated to 3.9.3 👍

client/src/components/new-request/submit-request/reserve-submit.vue
scottrumsby: the string values here are fed into the component via props and only really apply within it.
wouldn't an enum used as you suggest return a numeric index, like `ApprovalTypes.consent === 2` or whatever depending on the order.  I'm probably failing to see something obvious here but I don't get how this would help except possibly the type error.

client/tests/unit/applicant-info-1.spec.ts
severinbeauvais: BTW you can also do
`wrapper.find('#submit-continue-btn-false').classes('v-btn--disabled')...`

client/tests/unit/names-capture.spec.ts
```diff
@@ -32,23 +32,23 @@ describe('send-for-examination.vue', () => {
...
     it('calls showNextComponent() when continue button is clicked and form is valid', async () => {
       store.mutateNameChoicesToInitialState()
       store.setActiveComponent('NamesCapture')
       store.mutateNameChoices({ key: 'name1', value: 'LALA NAME' })
       store.mutateNameChoices({ key: 'designation1', value: 'INC.' })
       await wrapper.vm.$nextTick()
-      let btn = wrapper.find('#submit-continue-btn')
+      let btn = wrapper.find('#submit-continue-btn-true')
       btn.trigger('click')
       await wrapper.vm.$nextTick()
```
severinbeauvais, line 53: You can also do `await Vue.nextTick()`.

client/tests/unit/reserve-submit.spec.ts
```diff
@@ -10,6 +10,9 @@ const vuetify = new Vuetify()
 
 localVue.use(Vuetify)
 
+/* commented out lines and blocks of tests were correct and passing before it was decided not to allow auto
```
severinbeauvais, line 13: +1 for the comment!

But once this is merged, it's going to be hard to know that this comment applies to some changes below. For this reason I recommend moving the comment to the subject code blocks.

severinbeauvais: Or maybe define a global prop to toggle between old/new code?

client/src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: No, you can have an enum map to strings. It might seem pointless but it allows for strict type checking and is easier to find globally in code. Eg,
https://github.com/bcgov/business-filings-ui/blob/master/src/enums/nameRequestTypes.ts

client/src/components/new-request/submit-request/reserve-submit.vue
scottrumsby: if that works the way i think it should (ie. preventing pyramid of doom scenario), then i cannot believe i didn't know about this.

severinbeauvais (approved):
I found one id breakage but the rest are just comments/ suggestions.

client/src/components/new-request/submit-request/reserve-submit.vue
cameron-freshworks: oh ya, pretty new to me as well! I think it was added roughly this time last year.

client/tests/unit/names-capture.spec.ts
scottrumsby: the official documentation for vue-test-utils consistently uses wrapper.vm.$nextTick(). is there any competing reason to switch aside from maybe slightly less typing?

client/src/components/common/applicant-info-nav.vue
scottrumsby: good catch.  forgot this completely.

client/tests/unit/applicant-info-1.spec.ts
scottrumsby: my IDE has always auto-suggested toBeTruthy and toBeFalsy so I've always used them I can't say I was specifically even aware of alternative.  But given the issues you say you've experienced I guess I should probably change all uses throughout all the tests.

cameron-freshworks: It's just a suggestion, i don't think you need to update all the tests 👍

client/src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: Haha, totally Scott!
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

client/tests/unit/names-capture.spec.ts
severinbeauvais: I guess they both work, but I was following this:
https://vue-test-utils.vuejs.org/guides/testing-async-components.html

client/src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: TypeScript is now closer than ever to C# (not surprising since they're both from MS). But I like it!

cameron-freshworks (approved):
Looks good. Just let us know when your last commit goes in, and we'll get this in.

client/src/components/new-request/submit-request/reserve-submit.vue
scottrumsby: i got this partially implemented over the weekend and it makes no difference to the type errors.  Those strings are not used anywhere else in the app either so I think we should not bother with the enum at the moment because it doesn't seem to provide any benefit

severinbeauvais: OK.

We should look into this in the future, and enable some of the TS strict flags. Search for "--strict" on this page: https://www.typescriptlang.org/docs/handbook/compiler-options.html

client/src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -18,15 +18,6 @@ export default class ReserveSubmitButton extends Vue {
   get entity_type_cd () {
     return newReqModule.entity_type_cd
   }
-  messages () {
```
severinbeauvais, line 21: Is this obsolete?

scottrumsby: totally.  removed.

PR #135, opened 2020-10-29
removed @bizpal/open-services-bst package

severinbeauvais:
*Issue #:* /bcgov/entity#4906

*Description of changes:*
- deleted package
- removed Business Structures Tool from Search component
- removed obsolete CSS override

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -10,7 +10,6 @@
     "lint:nofix": "vue-cli-service lint --no-fix"
   },
   "dependencies": {
-    "@bizpal/open-services-bst": "^1.0.3",
     "axios": "^0.19.0",
     "babel-plugin-transform-decorators": "^6.24.1",
     "core-js": "^3.3.2",
--- a/client/src/components/new-request/search.vue
+++ b/client/src/components/new-request/search.vue
@@ -1,17 +1,7 @@
 <template>
   <v-container fluid id="new-request-container" class="copy-normal">
     <v-row justify="end">
-      <v-col cols="6" class="copy-normal">I need a name to:</v-col>
-      <v-col cols="6" style="display: flex; justify-content: flex-end">
-        <business-structures-tool>
-          <template #start="{ homeModal }">
-             <div id="#help-me-choose-activator"
-                  class="link-std"
-                  style="background-color: white"
-                  @click="homeModal()">Help Me Choose</div>
-          </template>
-        </business-structures-tool>
-      </v-col>
+      <v-col cols="12" class="copy-normal">I need a name to:</v-col>
       <v-col cols="5">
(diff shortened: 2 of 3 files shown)


client/src/components/new-request/search.vue
```diff
@@ -1,17 +1,7 @@
...
-        <business-structures-tool>
-          <template #start="{ homeModal }">
-             <div id="#help-me-choose-activator"
-                  class="link-std"
-                  style="background-color: white"
-                  @click="homeModal()">Help Me Choose</div>
-          </template>
-        </business-structures-tool>
-      </v-col>
+      <v-col cols="12" class="copy-normal">I need a name to:</v-col>
```
severinbeauvais, line 4: I changed this col to 12 units wide so it's always on one line (otherwise the responsive layout is messed up).

client/src/sass/main.sass
```diff
@@ -288,11 +288,6 @@
 .hidden
   display: none !important
 
-// @bizpal/open-services-bst imports Buefy which uses the .container class name which conflicts with vuetify
-// the following CSS fixes the issues from including the open-services-bst
-#landing-container, .container
-  max-width: unset !important
```
severinbeauvais, line 294: I tested locally (in Chrome) with and without this and it's now better without it (ie, responsive).

client/package.json
```diff
@@ -10,7 +10,6 @@
     "lint:nofix": "vue-cli-service lint --no-fix"
   },
   "dependencies": {
-    "@bizpal/open-services-bst": "^1.0.3",
```
severinbeauvais, line 13: I think we should check in the package-lock file so this UI is always built with the same versions of packages (ie, server vs local builds), but its diff doesn't look right. Ideas?

client/src/sass/main.sass
scottrumsby: yeah... for what it was, which was basically a questionnaire, it was massive.  it added 15 mb to package by itself.

scottrumsby (approved):

client/package.json
scottrumsby: we have been checking in package-lock.json so far

scottrumsby: you would have had it when you ran `npm i`.  are there a bunch of changes beyond the dependencies of the bizpal tool? it was a huge tool with many dependencies...

severinbeauvais: Yeah, it looks like the structure is messed up. I may be using a different version of npm... Let's look at it tomorrow.

cameron-freshworks (approved):

cameron-freshworks (approved):
Looks good! Held off on merge because you mentioned you might have another look into the packagae-lock but you are g2g.

PR #136, opened 2020-10-30
Name Request Login

cameron-freshworks:
*Issue #:* /bcgov/entity#5328

*Description of changes:*

* Imported and implemented Login mast 

* Updated Vue router

* Implemented Signin and Signout components

* Added .env

* Refactored Vuetify and MDI icon implementation to bring icons in sync between apps. 

* Updated MDI to latest.

* Implemented psuedo auth config to satisfy the common components auth redirection. 

*Note:*

* All login redirects ( success/fail/sign out ) are currently set back to name request. Each path is a variable prop we can update in the future if the flow is too change with regards to success/failure etc. 


client/package.json
```diff
@@ -14,10 +14,11 @@
     "babel-plugin-transform-decorators": "^6.24.1",
     "core-js": "^3.3.2",
     "lodash": "^4.17.20",
+    "@mdi/font": "^4.5.95",
     "material-design-icons-iconfont": "^5.0.1",
     "moment": "^2.27.0",
     "remove-accents": "^0.4.2",
-    "sbc-common-components": "2.3.3",
+    "sbc-common-components": "^2.4.3",
```
severinbeauvais, line 21: Please remove the caret. Since we generally don't know what changes in sbc-common-components, it's better to control what version we are using and then update manually.

client/.env
```diff
@@ -0,0 +1 @@
+VUE_APP_PATH = /nameRequest
```
severinbeauvais, line 1: As a result of this, please update the `publicPath` property in vue.config.js (same as in Filings UI).

client/src/sass/overrides.sass
```diff
@@ -43,15 +43,15 @@
...
+  min-width: 0
+  text-transform: none
+  font-weight: 400
+  letter-spacing: normal
+  font-size: .875rem
 
-.v-btn.normal
-  color: white !important
-  background-color: $grey-5 !important
+.v-icon + span,
```
severinbeauvais, line 52: This is quite a change. Do you have a list of buttons affected (for our review and also for QA to verify)?

client/src/main.ts
```diff
@@ -1,15 +1,16 @@
...
+import { getVueRouter } from '@/router'
 import store from './store'
 import { getConfig } from '@/plugins/getConfig'
-import vuetify from './plugins/vuetify'
+import Vuetify from "vuetify/lib"
 import KeyCloakService from 'sbc-common-components/src/services/keycloak.services'
 
-import 'quill/dist/quill.core.css'
+import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader
 import 'material-design-icons-iconfont/dist/material-design-icons.css'
```
severinbeauvais, line 10: Is this obsolete?

client/public/index.html
```diff
@@ -5,9 +5,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+    <link rel="preload" href="<%= BASE_URL %>fonts/materialdesignicons-webfont.927457ed.woff2" as="font" type="font/woff2" crossorigin>
```
severinbeauvais, line 8: Did you do anything to remove the MaterialIcons (not MDI) that are copied to the dist folder?

client/package.json
cameron-freshworks: ah right, will do, thanks!

client/.env
cameron-freshworks: I was curious about this. I thought about it but hesitated. 
If we do that, do we need to set this variable in each OS env?

client/src/main.ts
```diff
@@ -1,15 +1,16 @@
 import Vue from 'vue'
 import App from './App.vue'
-import router from './router'
+import { getVueRouter } from '@/router'
 import store from './store'
 import { getConfig } from '@/plugins/getConfig'
-import vuetify from './plugins/vuetify'
+import Vuetify from "vuetify/lib"
```
severinbeauvais, line 6: Please change " to '.

How much damage would be caused by removing the `'quotes': 'off'` rule in .eslintrc.js ?

client/src/sass/overrides.sass
cameron-freshworks: Yea, i'm pretty it affects almost ALL the buttons but we have a few styling ticket changes to come. This was just the first bit to at least bring the btn's in sync with Business/Filings etc, as it significantly affected the appearance of the common components.

Imo i think its ok that a few btn's look a bit different until we dig in to those styling tickets but totally your call.

client/src/main.ts
cameron-freshworks: No, not quite. A lot of the dropdown arrows and form icons are utilizing this. I didn't want to comb through everything icon related in this ticket because it extended a tad bit outta this tickets scope.

client/src/main.ts
severinbeauvais, line 6: This is replacing a plugin with a direct Vuetify import...
1. you should delete that obsolete plugin
2. you should copy over the other Vuetify override properties

... Or change it back and just update the iconfont property.

client/public/index.html
cameron-freshworks: No, though i'm not entirely sure what you mean.

client/src/main.ts
cameron-freshworks: ah good catch, thanks! 

hmm i dunno, living on the edge 😆 
Whats the value in removing it?

client/package.json
```diff
@@ -14,10 +14,11 @@
     "babel-plugin-transform-decorators": "^6.24.1",
     "core-js": "^3.3.2",
     "lodash": "^4.17.20",
+    "@mdi/font": "^4.5.95",
     "material-design-icons-iconfont": "^5.0.1",
```
severinbeauvais, line 18: Uninstall this :)

client/src/main.ts
cameron-freshworks: Absolutely, will do! I'll look into removing it.

I saw it imported directly into some components so i hesitated in doing so. I don't know the whole flow and how to test it... so i didn't want to do too much that could potentially be destructive to unknown parts.

client/src/main.ts
```diff
@@ -1,15 +1,16 @@
...
 import KeyCloakService from 'sbc-common-components/src/services/keycloak.services'
 
-import 'quill/dist/quill.core.css'
+import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader
 import 'material-design-icons-iconfont/dist/material-design-icons.css'
 import '@/sass/main.sass'
 import '@/sass/overrides.sass'
+import 'quill/dist/quill.core.css'
```
severinbeauvais, line 13: Are you sure about moving this CSS after the others? It means that it will override any styles declared in the previous files.

Typically, it's a good idea to have OUR overrides last here.

Maybe add a comment that order matters :)

client/package.json
cameron-freshworks: It's still being utilized in some places i believe. I would prefer to leave this Style related refactoring to the style tickets if possible 😁

client/src/main.ts
cameron-freshworks: haha thats actually news to me! I was curious about this, will re-order these 👍 Thanks!

client/src/views/auth/Signin.vue
```diff
@@ -1,82 +1,40 @@
...
-            },
-            incorporationApplication: {
-              nameRequest: {
-                legalType: this.nr?.entity_type_cd,
-                nrNumber: this.nr?.nrNum
-              }
-            }
-          }
-        }
-        const filingResponse = await BusinessServices.createBusiness(filingBody)
```
severinbeauvais, line 67: Did this code move somewhere else?

cameron-freshworks: No, it became obsolete. In ticket #5332, this will be redone, in a fashion.

client/.env
severinbeauvais: The Vue config file should pick it up from the environment.

This project doesn't have a `.env.production` file. It might be needed. Dunno.

client/src/sass/overrides.sass
severinbeauvais: I'm OK with this change, but it would be nice to verify WHAT changed so we can look at it closely (for example, what if the new style conflicts with a local style and the button is all messed up now?).

severinbeauvais: Aw nuts. I see there are 38 instances of a v-btn.

Do you think we should just risk this change and fix things later?

client/src/main.ts
severinbeauvais: OK. If you haven't already, please create a tech debt ticket to remove this properly.

client/src/sass/overrides.sass
cameron-freshworks: hmm, on second thought i think i'll try and isolate the header. 
We will still fix this later but in the meantime, best to leave whats existing untouched considering the scheduling etc.

cameron-freshworks: You're 100% right on the local conflicts. I imagine that would be extensive.

client/.env
cameron-freshworks: Thought about that as well. Prod would essentially be the same, thought not entirely sure if it's required. Will look into it.

client/public/index.html
severinbeauvais: Hmm, looks like importing the Material Icons CSS is enough to load them. It also looks like the Material Design Icons CSS does the same thing. So maybe this line isn't needed here???

There is something about preloading fonts so the app has them available right away, but this doesn't seem to be needed for icons.

Do the MDI still load without this line?

client/src/main.ts
severinbeauvais: It would enforce consistent quote usage (as we have with the Entity UIs).

cameron-freshworks: oh sorry, i misunderstood. I thought you meant turn it to `off` 😁

severinbeauvais: We can look at tightening the lint rules later.

client/src/main.ts
severinbeauvais: https://github.com/bcgov/bcrs-business-create-ui/blob/716868610ebe2a07c0c55fe34b52652e28a92950/src/main.ts#L18

client/public/index.html
severinbeauvais: I tested locally in Filings UI and it appears that this line is NOT required.

client/src/main.ts
severinbeauvais: Weird. It should not be needed in other components if it's already loaded when Vue is started.

severinbeauvais: Is there a way to test the other 4 components that import the vuetify plugin and confirm that they don't need to?

client/package.json
cameron-freshworks: Updated in next commit 👍

client/.env
cameron-freshworks: So after looking into this, the contents of the `.env` file will be accessible in all environments. Since the Path is consistent between local and deployments, we should be good with just a `.env`.

client/public/index.html
cameron-freshworks: Copy that, removed in next commit!

client/src/main.ts
cameron-freshworks: I'm not too familiar with those components tbh. I think i'd prefer to defer the refactoring ( ie removal of the plugin etc ) to the style tickets. Are you ok with that?

client/src/sass/overrides.sass
cameron-freshworks: Restored original btn styles and isolated the Header btns, in next commit 👍

client/src/main.ts
severinbeauvais: Yep.

severinbeauvais: So what are you doing for the icons? Going MDI or MD?

cameron-freshworks: Right now, using them both?
We should probably convert them all over to MDI as that is what the other UI's/designs so on, just don't want go out of the scope of this ticket. ie refactoring the style/icons/btns etc in this ui.

severinbeauvais: So it works if you import both but start Vuetify with the old ones?

https://github.com/bcgov/namerequest/blob/9bf8cc015d51b94be520d524c7aeaed09576a659/client/src/plugins/vuetify.ts#L24

cameron-freshworks: Yea, both icon types are displaying at the moment. I've tried to remove MD but we will need to go through and find the equivalent in MDI and refactor/rename them. (In the style update tickets ;) )

severinbeauvais: I'm still nervous about this. I see you do instantiate Vuetify with "mdi" iconfont (main.ts line 40), and then some components also instantiate Vuetify with "md" iconfont (via plugs/vuetify.ts). This suggests some part of the app have one font and other parts have another. I'm not sure how that plays together 😨

cameron-freshworks: I agree completely, this needs some attention, any proposals as to what to do?

This ticket was to bring in the SbcHeader and allow for authentication. Obviously after importing it, the header inherited style from the UI that skewed it's design badly and the current Icon library didn't support the Common Components requirements.

We can ..
A: Take this to the team and ask to extend this tickets scope and requirements to include refactoring the base Sass to bring it in line with the other UI's.
B: Tackle this work in a future style update ticket, under this story..
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5467
C: any other suggestions i'm open too!

severinbeauvais: Can you JUST change to using MDI icons (and revert the Vuetify code change you made)?

cameron-freshworks: Yep

cameron-freshworks: Just going to add in a little extra override style to the header as well 👍 

I viewed the style changes as a whole a bit larger scope than just the icons, my thoughts is they were creating the epic i linked above to capture all this work, which is i pushed back on just trying to add onto this ticket.

client/package.json
severinbeauvais, line 18: You can uninstall this now I think.

client/src/main.ts
severinbeauvais: Do you need to change any icon declarations? (Eg, v-icon from "close" to "mdi-close" ?)

client/package.json
cameron-freshworks: Done!

client/src/main.ts
cameron-freshworks: Not if we keep the plugin, it seems. Just had to change the `iconfont` in the plugin.

severinbeauvais (approved):

scottrumsby (approved):
Looks good.  Sorry about the "issue" i thought i found.

cameron-freshworks:
> Looks good. Sorry about the "issue" i thought i found.

Awesome, thanks!

PR #142, opened 2020-11-04
5332 Login Prompt / Affiliation / Draft Business Creation

cameron-freshworks:
*Issue #:* /bcgov/entity#5332

*Description of changes:*

* Imports and utilizes SbcAuthenticationOptionsDialog to prompt login at the point of Affiliating their NR, if NOT authenticated.

* Implements affiliation Mixin.

* Incorporates functionality for FUTURE `Incorporate` action/btn, in existing request display component.

* Implemented Global Getters and refactored a couple getters into store getters.

* Updated jest test related package versions.

* Updated jest config.


client/package-lock.json
```diff
@@ -1908,12 +1908,127 @@
...
+      "version": "25.2.3",
+      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-25.2.3.tgz",
+      "integrity": "sha512-JXc1nK/tXHiDhV55dvfzqtmP4S3sy3T3ouV2tkViZgxY/zeUkcpQcQPGRlgF4KmWzWW5oiWYSZwtCB+2RsE4Fw==",
       "dev": true,
       "requires": {
-        "jest-diff": "^24.3.0"
+        "jest-diff": "^25.2.1",
+        "pretty-format": "^25.2.1"
+      },
+      "dependencies": {
```
severinbeauvais, line 1919: This isn't right.

I know you know this -- just adding a comment to remind myself to check this before merging.

client/package.json
```diff
@@ -30,13 +30,13 @@
     "vuex-class": "^0.3.2"
   },
   "devDependencies": {
-    "@types/jest": "^24.9.0",
+    "@types/jest": "^25.1.4",
```
severinbeauvais, line 33: Any impact going from v24 to v25?

client/package.json
```diff
@@ -30,13 +30,13 @@
...
     "@types/sinon": "^7.5.1",
-    "@vue/cli-plugin-babel": "^4.0.0",
+    "@vue/cli-plugin-babel": "^4.2.3",
```
severinbeauvais, line 35: It's actually already using v4.4.1 (according to package lock file).

severinbeauvais: Also "@vue/cli-plugin-unit-jest" below.

client/package.json
```diff
@@ -52,7 +52,7 @@
     "sass-loader": "^8.0.0",
     "sinon": "^8.1.1",
     "style-resources-loader": "^1.3.2",
-    "typescript": "^3.5.3",
+    "typescript": "^3.8.3",
```
severinbeauvais, line 55: It's actually already on v3.9.3.

client/src/components/mixins/nr-affiliation-mixin.ts
```diff
@@ -0,0 +1,63 @@
...
+      const requestBody: CreateNRAffiliationRequestBody = {
+        businessIdentifier: nr.nrNum,
+        phone: nr.applicants?.phoneNumber || '',
+        email: nr.applicants?.emailAddress || ''
+      }
+
+      // Request to affiliate NR to current account
+      const nrResponse = await AuthServices.createNRAffiliation(currentOrganizationId, requestBody)
+      if (nrResponse?.status === 201) {
+      // update the legal api if the status is success
```
severinbeauvais, line 27: Possible lint error (indentation). Run `npm run lint` before your final commit.

client/src/components/existing-request/existing-request-display.vue
```diff
@@ -279,11 +288,15 @@ export default class ExistingRequestDisplay extends Vue {
...
 
-  /** Open Incorporate Now Login Modal and apply NR Data to Session */
-  activateILModal () {
-    newReqModule.mutateIncorporateLoginModalVisible(true)
-    // nr persisted in the session to be used for affiliation/creation upon authentication in Signin.vue.
-    sessionStorage.setItem('NR_DATA', JSON.stringify(this.nr))
+  /** Affiliate the current Nr if authenticated or prompt login if unauthenticated */
+  affiliateOrLogin (): void {
+    if (this.isAuthenticated) {
+      this.createAffiliation(this.nr)
```
severinbeauvais, line 294: Would be good to await this call (since it's async).

Is there a spinner or something while waiting for this to finish, since I imagine it takes several seconds to create the NR affiliation and to create the business.

client/src/components/mixins/nr-affiliation-mixin.ts
```diff
@@ -0,0 +1,63 @@
...
+              nameRequest: {
+                legalType: nr.entity_type_cd,
+                nrNumber: nr.nrNum
+              }
+            }
+          }
+        }
+        const filingResponse = await BusinessServices.createBusiness(filingBody)
+        // navigate to manage business dashboard
+        if (filingResponse?.status === 201) {
```
severinbeauvais, line 47: Do you need an else block (with a corresponding `throw` so execution goes to the catch block below)?

Same for line 26 above.

client/src/components/modals/incorporate-login.vue
```diff
@@ -1,26 +1,32 @@
...
   }
 })
 export default class IncorporateLoginModal extends Vue {
+  readonly getNameRequestUrl!: string
+
   get showModal () {
     return newReqModule.incorporateLoginModalVisible
   }
-  set showModal (value: boolean) {
+  toggleModal (value: boolean) {
```
severinbeauvais, line 29: It's not toggling if you set the value 😛

client/src/components/modals/incorporate-login.vue
```diff
@@ -1,26 +1,32 @@
 <template>
-  <v-dialog v-model="showModal" max-width="60%" >
+  <v-dialog v-model="showModal" max-width="850px" persistent>
```
severinbeauvais, line 2: Would it be easy to "attach" this v-dialog to its parent? If no, nvm.

client/package-lock.json
severinbeauvais: OK, now it's right.

client/src/store/index.ts
```diff
@@ -21,5 +22,6 @@ export default new Vuex.Store<RootStateI>({
     checkInOnExpire: false
   },
   actions: rootActions,
-  mutations: rootMutations
+  mutations: rootMutations,
+  getters: Getters
```
severinbeauvais, line 26: If you imported these things "as" their property name, this would be even cleaner. Like this:
https://github.com/bcgov/business-filings-ui/blob/2fbb3c2973c7d6454305034312efa5b1960c9c7e/src/store/index.ts#L14

client/src/views/auth/Signin.vue
```diff
@@ -8,27 +8,34 @@
...
   @Prop({ default: '' }) redirectUrl: string
   @Prop({ default: '' }) redirectUrlLoginFail: string
 
   /** called when keycloak session is ready. */
   private async onReady () {
     if (this.redirectUrl) {
-      // navigate to the route we originally came from
-      await this.$router.push(this.redirectUrl)
+      // navigate to the route we originally came from and process an affiliation if an Nr is pending.
+      await this.$router.push(this.redirectUrl).then(() => {
```
severinbeauvais, line 34: What does `then` do on a router push?

severinbeauvais (commented):
Looks good so far.

client/package.json
cameron-freshworks: Besides avoiding some type errors in Jest testing, not that i noticed!

client/package.json
cameron-freshworks: Ah! Confusing. Why is this the case?

client/package.json
cameron-freshworks: Ya, i don't understand this one, how can the package and lock be so out of sync.

client/src/components/mixins/nr-affiliation-mixin.ts
cameron-freshworks: Linter didn't pick it up, but i will fix/indent here, thanks!

client/package.json
severinbeauvais: The caret at the front of the semantic version indicates that it can auto-update to anything up to "4.x.y".

Note that we check in the package lock file to specify exactly which versions are used for the build (otherwise it would auto-update).

This gets updated when you do a `npm i` I think.

You can check the package lock file to see the exact versions that are used.

severinbeauvais: ^^ We don't want the build server to auto-update. We want it to use exactly the same package versions that we use locally.

client/package.json
severinbeauvais: See my response above. It's due to the caret.

(And that's the reason we don't use a caret for the sbc-common-components.)

client/src/components/existing-request/existing-request-display.vue
cameron-freshworks: That's probably a good idea! Just out of curiosity though, with no subsequent actions and also since the requests within are awaited, does it matter to await this?

No spinner at the moment and there is no global state to disable the app. I guess using a spinner you could disable any further action until the requests are resolved, i'll look into this!

client/src/components/mixins/nr-affiliation-mixin.ts
cameron-freshworks: Good catch, yes! I can't really rely on a 'Created' response can i..

client/src/components/modals/incorporate-login.vue
cameron-freshworks: Valid! fixed in next commit 😆

client/src/components/modals/incorporate-login.vue
cameron-freshworks: This is refactored a bit, i made a new common component that is a dialog itself, so this will change a bit.

client/src/store/index.ts
cameron-freshworks: I don't disagree! 
I'll look into it 👍

client/src/components/existing-request/existing-request-display.vue
severinbeauvais: In this case you do not have to await, since there is nothing after it that depends on it. (So you're throwing it out there to complete whenever.)

Yeah, you:
1. want the user to know something is happening while you create the affiliation and business
2. don't want the user to click something else if they get impatient

client/src/components/mixins/nr-affiliation-mixin.ts
severinbeauvais: Because this code is a try-catch block, if `createBusiness()` fails (and since you have no `.catch()` it will throw an exception), it will be caught.

I'm just thinking there might be other valid response statuses. 2xx 3xx?

client/src/views/auth/Signin.vue
cameron-freshworks: A promise to resolve something after the redirect.

I believe it could work just the same as 

`await this.$router.push(this.redirectUrl)`
` // If there is stored NR data to process, create the affiliation`
`const nr = JSON.parse(sessionStorage.getItem('NR_DATA'))`
` if (nr) this.createAffiliation(nr)`

severinbeauvais: OK, cool.

Shouldn't the affiliation be created _before_ the route navigation? What if it fails?

client/src/components/mixins/nr-affiliation-mixin.ts
cameron-freshworks: I'm not sure, i'll have to look in auth. But when they create affiliations locally, the only (successful) response that is watch for was a 'Created' (201) response. I will check again though!

client/src/views/auth/Signin.vue
cameron-freshworks: Another good point. There really is no value in the redirect if we are starting the affiliation process. In this instance, the redirect to root should only happen in the event of a failed affiliation/temp business creation, will refactor this!

cameron-freshworks: Since the affiliation process will either result in the user at manage business dashboard or root. This redirect here should only occur when a client is simply logging in through regular means.

Thanks for this!

client/src/components/mixins/nr-affiliation-mixin.ts
severinbeauvais: Looks like checking for 201 is a "safety check" then. I'm in favour! But what would you do if it wasn't 201?

client/package.json
cameron-freshworks: How should we approach this. Update all the versions to the ones actually used according to the package-lock?

severinbeauvais: Only if you are as AR as me. And it will gradually get out of sync again anyway (which should have no impact except to legibility).
So,  maybe update it this time then fuggedaboutit.

cameron-freshworks: 😆 😆  Alrighty!

client/src/components/existing-request/existing-request-display.vue
cameron-freshworks: Updated in next commit 👍

client/src/store/index.ts
cameron-freshworks: Updated in next commit!

client/src/views/auth/Signin.vue
cameron-freshworks: Updated in next commit!

cameron-freshworks:
Build failed because the SBC common component i made, hasn't been merged in yet 😁

client/src/views/auth/Signin.vue
severinbeauvais: Looks good!

client/src/components/mixins/nr-affiliation-mixin.ts
cameron-freshworks: Ah, i knew i missed something. Looking into this quickly :)

client/src/components/mixins/nr-affiliation-mixin.ts
```diff
@@ -0,0 +1,68 @@
...
+      const currentOrganizationId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT')).id
+      const requestBody: CreateNRAffiliationRequestBody = {
+        businessIdentifier: nr.nrNum,
+        phone: nr.applicants?.phoneNumber || '',
+        email: nr.applicants?.emailAddress || ''
+      }
+
+      // Request to affiliate NR to current account
+      const nrResponse = await AuthServices.createNRAffiliation(currentOrganizationId, requestBody)
+      if (nrResponse?.status === 201) {
```
severinbeauvais, line 29: I'm looking at createBusiness() and I see 3 possible return cases:
1. the POST succeeded and the status is 2xx, so return the response
2. the POST succeeded and the status is not 2xx, so return nothing
3. the POST failed, so return `{ errorMsg }`

Based on this, I agree that it's wise to check here that you have case 1. Two questions:
- must you receive a 201, or is any 2xx valid?
- what are you doing to catch errors, since createBusiness() only returns an exception if the second call (removeAffiliation) fails?

severinbeauvais: (Sorry, move this comment to line 48.)

cameron-freshworks: I definitely agree, i think i need to expand this condition to include a 200 (in case) as well as throw an error to the catch

severinbeauvais: I see that createNRAffiliation() will either return a resolved or rejected promise (ie, then and catch) (ie, success or exception).

If createNRAffiliation() fails then the catch for the block here will be executed, which is fine.

If createNRAffiliation() succeeds then you check for a 201, which is OK, but what if you receive another "succeess" status? Is that possible?

cameron-freshworks: K, pushed a change. 

That being said, we will still need to do work on handling the errors more explicitly, but that will be covered when designs are updated for those dialogs etc.

client/src/components/mixins/nr-affiliation-mixin.ts
severinbeauvais: Your change looks good. If the "if" fails then you throw an error, which is caught by the catch block etc. Great!

severinbeauvais (approved):

client/src/views/auth/Signin.vue
```diff
@@ -30,9 +30,11 @@ export default class Signin extends Mixins(NrAffiliationMixin) {
   /** called when keycloak session is ready. */
   private async onReady () {
     if (this.redirectUrl) {
+      await this.$router.push(this.redirectUrl)
+
       // If there is stored NR data to process, create the affiliation OR navigate to redirectUrl
       const nr = JSON.parse(sessionStorage.getItem('NR_DATA'))
-      nr ? await this.createAffiliation(nr) : await this.$router.push(this.redirectUrl)
+      if (nr) await this.createAffiliation(nr)
```
cameron-freshworks, line 37: @severinbeauvais I had to change this back. Signin wouldn't resolve back to root in the event that the business process failed. So it's safer to process the redirect first, and then handle the business processing afterwards :)

severinbeauvais (approved):
I think your latest changes look fine.
It would be nice for Lucas or Scott to review this as well, but if you want to merge ASAP then go for it.

cameron-freshworks:
> I think your latest changes look fine.
> It would be nice for Lucas or Scott to review this as well, but if you want to merge ASAP then go for it.

Absolutely, i went through a little Demo with Scott since this functionality is still behind a missing NR action.

scottrumsby (approved):
Seems okay to me.

PR #143, opened 2020-11-05
hotfix/5353+5553-Fix-Draft-Timer

bcdevlucas:
*Issue #:* /bcgov/entity###

*Description of changes:*

- Remove NR ID requirement to extend session when dealing with a new NR.
- Improve timer display, refactor a couple things to prevent widget from flickering into display temporarily when it's not supposed to.
- Fix redirect to start when clicking 'Start Search Over'.

### DIFF
--- a/client/src/components/mixins/displayed-component-mixin.ts
+++ b/client/src/components/mixins/displayed-component-mixin.ts
@@ -39,6 +39,9 @@ export default class DisplayedComponentMixin extends Vue {
       await newRequestModule.checkinNameRequest()
       timerModule.stopTimer({ id: EXISTING_NR_TIMER_NAME })
       this.redirectToStart()
+    } else {
+      await newRequestModule.cancelAnalyzeName('Tabs')
+      this.redirectToStart()
     }
   }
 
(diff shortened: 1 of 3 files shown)


severinbeauvais:
Please update the issue numbers in the description. This "links" this PR with the ticket in Zenhub,

client/src/components/mixins/displayed-component-mixin.ts
```diff
@@ -39,6 +39,9 @@ export default class DisplayedComponentMixin extends Vue {
       await newRequestModule.checkinNameRequest()
       timerModule.stopTimer({ id: EXISTING_NR_TIMER_NAME })
       this.redirectToStart()
+    } else {
```
severinbeauvais, line 42: 👍 
A lot of code problems can be avoided (or easier to find) if all "if" blocks have an "else".
```
if () {
  ...
} else if () {
  ...
} else {
  // even if you think this should never happen, do something here
}
```

severinbeauvais (approved):
Looks good.

PR #146, opened 2020-11-10
Fix FeeSummary component timing issue /bcgov/entity#5246

bcdevlucas:
*Issue #:* /bcgov/entity#5246

*Description of changes:*

### DIFF
--- a/client/src/components/payment/fee-summary.vue
+++ b/client/src/components/payment/fee-summary.vue
@@ -68,31 +68,16 @@
 </template>
 
 <script lang="ts">
-import { Vue, Component, Prop, Watch, Emit } from 'vue-property-decorator'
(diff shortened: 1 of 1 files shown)


client/src/components/payment/fee-summary.vue
```diff
@@ -94,22 +94,22 @@ export default class FeeSummary extends Vue {
 
   /* getter */
   protected get totalFees (): number {
-    return this.fees.reduce((feeTotal: number, item: any) => {
+    return this.fees instanceof Array ? this.fees.reduce((feeTotal: number, item: any) => {
```
severinbeauvais, line 97: What is this actually doing? It looks like (lines 85-86) this.fees is initialized to an array, so when is it NOT an array?

severinbeauvais (approved):

PR #147, opened 2020-11-13
Fix PDF download, updated for new Payment API / SBC Pay API

bcdevlucas:
*Issue #:* /bcgov/entity###
/bcgov/entity#5318

*Description of changes:*


client/src/components/payment/payment-mixin.ts
```diff
@@ -17,6 +17,10 @@ export default class PaymentMixin extends Vue {
     return this.$store.getters[paymentTypes.GET_SBC_PAYMENT]
   }
 
+  get sbcPaymentStatus () {
+    return this.$store.getters[paymentTypes.GET_SBC_PAYMENT_STATUS]
```
severinbeauvais, line 21: At some point, "paymentTypes" should be changed to PascalCase so it's not confused with a property (which should be camelCase).

client/src/modules/payment/models.ts
```diff
@@ -1,5 +1,10 @@
 import { ErrorI } from '@/modules/error/store/actions'
 
+export const PaymentStatus = {
```
severinbeauvais, line 3: If this type is the same as SbcPaymentStatus, why have both?

client/src/modules/payment/store/getters.ts
```diff
@@ -10,6 +10,9 @@ export default {
   [types.PAYMENT_COMPLETE_MODAL_IS_VISIBLE]: state => state[STATE_KEY].isPaymentCompleteVisible,
   [types.PAYMENT_IS_IN_PROGRESS]: state => state[STATE_KEY].isProcessing,
   [types.GET_SBC_PAYMENT]: state => state[STATE_KEY].sbcPayment,
+  [types.GET_SBC_PAYMENT_STATUS]: state => (state[STATE_KEY].sbcPayment)
```
severinbeauvais, line 13: "types" should be capitalized to distinguish it from a property/variable.

client/src/modules/payment/models.ts
bcdevlucas: The codes are different on both sides for various reasons including that a payment record on their end is not 100% analagous to a payment on ours. Our payment only references thee SbcPayment. If we need the SbcPayment we do not get it from our side, we always make the API call as we cannot control what happens on the SbcPayment side of things.

client/src/modules/payment/store/getters.ts
bcdevlucas: Not enforced by eslint, and those are not actually 'types' referenced in those files, they're exported string constants so technically they are just exported variables. I could have written 'action_dictionary.<type>' or 'all_actions.<type> I guess but yeah...

client/src/components/payment/payment-mixin.ts
bcdevlucas: Not enforced by eslint, and those are not actually 'types' referenced in those files, they're exported string constants so technically they are just exported variables. I could have written 'action_dictionary.<type>' or 'all_actions.<type> I guess but yeah...

client/src/modules/payment/models.ts
severinbeauvais: I agree with that approach. The 2 types just look very similar to me :)

client/src/components/payment/payment-mixin.ts
severinbeauvais: You're right, but they're used as types/enum.

Please fix if you're in this area again.

PR #149, opened 2020-11-18
Set longer NR timeout when in dev mode

severinbeauvais:
*Issue #:* None

*Description of changes:*
- set NR timeouts to 60 min (instead of 5 min) when in dev mode
- deleted unused debounce()
- renamed a const for consistency

### DIFF
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -41,7 +41,8 @@ import { ErrorI } from '@/modules/error/store/actions'
 import * as types from '@/store/types'
 
 const qs: any = querystring
-const ANALYSIS_TIMEOUT_MS = 180 * 1000 // 3 minuteslet source: any
+const ANALYSIS_TIMEOUT_MS = 180 * 1000 // 3 minutes
+let source: any
 
 const MINUTES_60_IN_MS = 60 * (60 * 1000)
 const MINUTES_5_IN_MS = 5 * (60 * 1000)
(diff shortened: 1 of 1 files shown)


client/src/store/new-request-module.ts
```diff
@@ -41,14 +41,16 @@ import { ErrorI } from '@/modules/error/store/actions'
...
-const analysisTimeout: number = 180000
+const ANALYSIS_TIMEOUT_MS = 180 * 1000 // 3 minutes
 let source: any
 
+const MINUTES_60_IN_MS = 60 * (60 * 1000)
+const MINUTES_5_IN_MS = 5 * (60 * 1000)
+
 export const NR_COMPLETION_TIMER_NAME = 'nrCompletionTimer'
-export const NR_COMPLETION_TIMEOUT_MS = 5 * (60 * 1000) // Set to 5 minutes
+export const NR_COMPLETION_TIMEOUT_MS = window['webpackHotUpdate'] ? MINUTES_60_IN_MS : MINUTES_5_IN_MS
```
severinbeauvais, line 51: `window['webpackHotUpdate']` is a function when running in dev mode (ie, locally) and undefined otherwise.

cameron-freshworks: ah interesting! I'll look it up.

Is this something we will leave in permanently?
Either way maybe just a small comment in the code similar to what you mentioned above would be awesome :)

cameron-freshworks (approved):
Looks good! Just the one comment.

client/src/store/new-request-module.ts
severinbeauvais: Done.

scottrumsby (approved):
looks fine

PR #150, opened 2020-11-18
Remove Canadian and Foreign Unlimited Liability Company request types

lekshmimallika-aot:
*Issue #:* /bcgov/entity#5586

*Description of changes:*
Removed XUL from xpro entity types
Removed XUL from the entity type mapping for specified request action codes

Note: There is a reference to 'XUL' in line 949 and 566 in new-request-module. I am not sure whether this should be removed.


scottrumsby (commented):
The other references in new-request-module should probably be removed too.  It shouldn't cause any errors leaving them in but anywhere they are used is basically unreachable code now.

scottrumsby (approved):
Looks good.

PR #153, opened 2020-11-19
Updated package versions + some unit test fixes

severinbeauvais:
*Issue #:* None

*Description of changes:*
Commit 1:
- updated packages to versions actually being used (per lock file)
- moved some packages to dev
- fixed "data-app" warnings in unit tests
- bumped project version

Commit 2:
- updated vue-test-utils
- bumped project version


client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "0.1.0",
+  "version": "0.1.2",
```
severinbeauvais, line 3: I jumped 0.0.1 in each commmit (in case we want to revert anything).

client/package.json
```diff
@@ -10,53 +10,53 @@
     "lint:nofix": "vue-cli-service lint --no-fix"
   },
   "dependencies": {
-    "axios": "^0.19.0",
+    "@mdi/font": "^4.5.95",
```
severinbeauvais, line 13: To clarify, these are the version this project was already using (per the package-lock.json file). This just makes it easier to compare against other projects.

client/package.json
```diff
@@ -10,53 +10,53 @@
...
+    "@vue/cli-plugin-router": "^4.4.1",
+    "@vue/cli-plugin-typescript": "^4.4.1",
+    "@vue/cli-plugin-unit-jest": "^4.4.1",
+    "@vue/cli-plugin-vuex": "^4.4.1",
+    "@vue/cli-service": "^4.4.1",
     "@vue/eslint-config-standard": "^4.0.0",
     "@vue/eslint-config-typescript": "^4.0.0",
-    "@vue/test-utils": "^1.0.0-beta.30",
-    "cross-env": "^7.0.0",
+    "@vue/test-utils": "^1.1.1",
```
severinbeauvais, line 42: ... Except for this one, which is a "significant" jump.

In this project, all the unit tests worked fine with the new version. (I was hoping it would resolve some warnings but it didn't. It appears those warnings are from the component is built or mounted or .../)

In other projects, the unit tests are sensitive to changes between beta versions so this package can't simply (sic) be upgrade simply (sic).

client/src/components/modals/nr-not-required.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-  <v-dialog v-model="showModal" max-width="40%">
+  <!-- NB: dummy modal attach point so unit tests can see it -->
+  <v-dialog v-model="showModal" max-width="40%" attach="">
```
severinbeauvais, line 3: This seems to render the dialog in-line. Cool!

(We might not need to attach dialogs to their parent any more in our other UI projects.)

(Yes, I tested that this dialog still displays correctly in the running app.)

cameron-freshworks: Is there a default `attach` value within the vue component itself? Just curious how this would differ from not passing the property at all? Unless this is overriding a default..?

client/tests/unit/applicant-info-1.spec.ts
```diff
@@ -5,7 +5,9 @@ import Vuetify from 'vuetify'
 import $intJurisdictions from '@/store/list-data/intl-jurisdictions'
 import $canJurisdictions from '@/store/list-data/canada-jurisdictions'
 
-const store = newReqModule
+// Prevent the warning "[Vuetify] Unable to locate target [data-app]"
+document.body.setAttribute('data-app', 'true')
```
cameron-freshworks, line 9: bueno!

cameron-freshworks (approved):
Looks good! Just had one question

client/src/components/modals/nr-not-required.vue
severinbeauvais: The default is that it attaches to v-app.

With no attach prop, in the unit tests, you don't see the dialog contents (because it's rendered outside the component).

With my code change, it basically attaches to itself, so you can see the dialog's HTML in the unit test.

https://vuetifyjs.com/en/api/v-dialog/

client/tests/unit/applicant-info-1.spec.ts
severinbeauvais: Yes but needs to be used sparingly as it seems to break some tests (and, in some cases, it's just not needed).

PR #155, opened 2020-11-23
5664

scottrumsby:
/bcgov/entity#5664

Updated the grey-box.vue components to match the new design documents


client/src/components/new-request/analyze-results.vue
```diff
@@ -270,13 +301,115 @@ export default class AnalyzeResults extends Vue {
     }
     return []
   }
+  get designations () {
+    if (this.issue && this.issue.designations) {
+      return this.issue.designations
+    }
+    return null
+  }
   get designationIsFixed () {
```
scottrumsby, line 310: This getter was previously living in grey-boxes.vue and had a watcher to update the store when its value changed because both this component (parent comp) and grey-boxes.vue (child component) relied on it.  Obviously makes more sense to have it exist in parent and pass to child as a Prop.

client/src/components/new-request/analyze-results.vue
```diff
@@ -244,8 +247,36 @@ export default class AnalyzeResults extends Vue {
...
+    let { allDesignationsStripped } = this
+    if (this.nameActionWords.length > 0) {
+      for (let word of this.nameActionWords) {
+        nameTest = replaceWord(nameTest, word)
+        allDesignationsStripped = replaceWord(allDesignationsStripped, word)
+      }
+    }
+    return (allDesignationsStripped === nameTest)
+  }
   get changesInBaseName () {
```
scottrumsby, line 275: This getter was previously living in grey-boxes.vue and had a watcher to update the store when its value changed because both this component (parent comp) and grey-boxes.vue (child component) relied on it. Obviously makes more sense to have it exist in parent and pass to child as a Prop.

client/src/components/new-request/grey-box.vue
```diff
@@ -140,25 +134,23 @@ import { matchWord, removeExcessSpaces, replaceWord } from '@/plugins/utilities'
...
 })
 export default class GreyBox extends Vue {
+  @Prop(Boolean) changesInBaseName: boolean
+  @Prop(Boolean) designationIsFixed: boolean
   @Prop(Number) i: number
   @Prop(Number) issueIndex: number
   @Prop(Object) option: OptionI
   @Prop(String) originalName: string
 
-  @Watch('changesInBaseName', { immediate: true })
```
scottrumsby, line 148: These were no longer needed now that related getters exist in the parent component.

client/src/components/new-request/grey-box.vue
```diff
@@ -450,6 +294,28 @@ export default class GreyBox extends Vue {
     }
     return []
   }
+  get optionClasses () {
```
scottrumsby, line 297: This was previously implemented as a method.  as the i parameter is actually the same `this.i`, have moved it to be a getter.

client/src/components/new-request/submit-request/entity-cannot-be-auto-analyzed.vue
```diff
@@ -10,29 +10,6 @@
         <v-col cols="12"
                class="text-center mt-n4"
                v-html="title" />
-        <v-col v-for="(box, i) in boxes" :key="'box-'+i">
```
scottrumsby, line 13: per the desings this is no longer to be communicated inside of a grey box

client/server.js
```diff
@@ -1,1122 +0,0 @@
-const jsonServer = require('json-server')
```
scottrumsby, line 1: I deleted this file because at one time it was a useful mock backend server you could run in node to test the front end but its very outdated now and crashes the front-end because it returns invalid data.

client/src/components/new-request/analyze-results.vue
```diff
@@ -79,16 +79,19 @@
...
-              <v-col :key="issue.issue_type + '-' + option.header + '-' + optionIndex"
-                     cols="auto"
-                     v-for="(option, optionIndex) of issue.setup">
-                <GreyBox :i="optionIndex"
-                         :issueIndex="issueIndex"
-                         :option="option"
-                         :originalName="originalName" />
-              </v-col>
-            </v-row>
+
```
cameron-freshworks, line 82: Super minor but the comment line and spacing look a tad off.

client/src/components/new-request/analyze-results.vue
```diff
@@ -244,8 +247,36 @@ export default class AnalyzeResults extends Vue {
...
+  }
+  get allowProceed () {
+    /* newReqModule.designationIssueTypes is a list of all the designation-related issue types.  designationIsFixed
+     will not be true unless one of these issues was solved.  some sets of issues will not include a designation-related
+     issue and this will need to return true in these cases even tho designationIsFixed will be false */
+    if (!this.hasDesignationIssue) {
+      if (this.isLastIndex) {
+        return true
+      }
+    }
```
cameron-freshworks, line 261: Might be able to simplify this condition into one statement. 
ie `(!this.hasDesignationIssue && this.isLastIndex)`

client/src/components/new-request/analyze-results.vue
```diff
@@ -270,13 +301,115 @@ export default class AnalyzeResults extends Vue {
...
+      if (!this.changesInBaseName) {
+        // creating a new version of allDesignationsList because we are going to modify it but still need the original
+        let AllDesignationsList = allDesignationsList
+        // allDesignations is an array of objects containing {words, end} keys, set designationEntityTypes to the
+        // relevant object
+        const designationEntityTypes = allDesignations[this.entity_type_cd]
+        /* designationEntityTypes.words.length === 0 is true for types 'PAR', 'FI', 'PA', and the proprietorships so
+         the designation issue is fixed as long as there are no designations present and no nameAction words which in
+         this situation would only be designation-like words to be removed */
+        if (designationEntityTypes && designationEntityTypes.words.length === 0) {
```
severinbeauvais, line 322: FYI, you can now do `if (designationEntityTypes?.words.length === 0)...`.

If "designationEntityTypes" is falsey they the statement is undefined (which is not === 0).

client/src/components/new-request/grey-box.vue
```diff
@@ -1,130 +1,124 @@
...
+            <!-- NAME/DESIGNATION-MANIPULATING OPTION BOXES -->
+            <template v-if="isDesignationIssueType">
+              <transition name="fade"
+                          mode="out-in">
+                <v-col v-if="!designationIsFixed && !changesInBaseName && i === 0"
+                       class="text-center designation-error-col"
+                       key="designation-error-col">
+                  <template v-if="Array.isArray(designations) && designations.length > 0">
+                    <div class="designation-buttons pa-0">
+                      <button :id="'designation-'+d"
```
severinbeauvais, line 52: It would be nice to have "button" or "btn" as part of the id.

client/src/sass/main.sass
```diff
@@ -196,6 +196,10 @@
   color: $link !important
   border-color: $link !important
 
+#reserve-submit-button
```
severinbeauvais, line 199: Wouldn't this make more sense to be in the file that uses it (possibly as scoped sass)?

client/src/vue.d.ts
```diff
@@ -1,4 +1,5 @@
 import Vue from 'vue-property-decorator'
+import quill from 'quill'
```
severinbeauvais, line 2: I don't see where this is used in here...?

client/src/components/new-request/grey-box.vue
scottrumsby: is this something we would actually like to change? Because the vuetify tag was called 'v-btn' i'm pretty sure anywhere/everywhere in the app where I had to name a class, or ID, or make a method or getter refer to a button, I wrote it "btn" unfortunately.

severinbeauvais (commented):
1. Are there arrows in the gray boxes that need to be changed to "to"?

2. Where is the text (option.header) that goes beside the "Option x:" title? Does it match the InVision design?

3. Where is the change from "Helpful Tip" to "Required Action"?

4. Where is the change to the text in the Obtain Consent box? (With checkbox and with button.)

5. Where is the change to the text in the Add Words box?

6. Where is the change to the content in the Send for Review box?

7. Where is the change to the Change Business Type box?

8. Where is the "Move the designation to the end of the name" text change?

9. Where is the "Review your name as it appears..." text/content change?

client/src/sass/main.sass
severinbeauvais: I think, if possible, that "main.sass" should contain global CSS, whereas specific styles should be localized to components.

client/src/components/new-request/grey-box.vue
severinbeauvais: Whether it's a Vuetify button or a plain HTML button, it's still ultimately rendered as a button element.

I'm thinking it would make it easier, in a SASS file or a test file, to know that this element is a button, if the name contained "btn", eg. designation-foo-btn. And it would be consistent with other button IDs.

client/src/sass/main.sass
scottrumsby: would that actually work?  the id property is set in the parent component's template and the scoped sass would be in the child component so I don't think anything scoped would referencing the #id would have any effect.

and i don't think I put any non-scoped CSS inside a SFC because it just gets confusing if globally-scoped CSS could be hiding in any component anywhere in the app.  I figured best to keep that in one place.

client/src/components/new-request/grey-box.vue
scottrumsby: sure

client/src/vue.d.ts
scottrumsby: its not... will remove

scottrumsby: > 1. Are there arrows in the gray boxes that need to be changed to "to"?
needs to be changed in backend
> 2. Where is the text (option.header) that goes beside the "Option x:" title? Does it match the InVision design?
It will when the text in the backend gets updated
> 3. Where is the change from "Helpful Tip" to "Required Action"?
backend change
> 4. Where is the change to the text in the Obtain Consent box? (With checkbox and with button.)
backend change
> 5. Where is the change to the text in the Add Words box?
> 6. Where is the change to the content in the Send for Review box?
ditto
> 7. Where is the change to the Change Business Type box?
> 8. Where is the "Move the designation to the end of the name" text change?
same
> 9. Where is the "Review your name as it appears..." text/content change?
likewise

scottrumsby: its not.  removed.

scottrumsby:
> 1. Are there arrows in the gray boxes that need to be changed to "to"?
> 2. Where is the text (option.header) that goes beside the "Option x:" title? Does it match the InVision design?
> 3. Where is the change from "Helpful Tip" to "Required Action"?
> 4. Where is the change to the text in the Obtain Consent box? (With checkbox and with button.)
> 5. Where is the change to the text in the Add Words box?
> 6. Where is the change to the content in the Send for Review box?
> 7. Where is the change to the Change Business Type box?
> 8. Where is the "Move the designation to the end of the name" text change?
> 9. Where is the "Review your name as it appears..." text/content change?

> 1. Are there arrows in the gray boxes that need to be changed to "to"?
needs to be changed in backend
> 2. Where is the text (option.header) that goes beside the "Option x:" title? Does it match the InVision design?
It will when the text in the backend gets updated
> 3. Where is the change from "Helpful Tip" to "Required Action"?
backend change
> 4. Where is the change to the text in the Obtain Consent box? (With checkbox and with button.)
backend change
> 5. Where is the change to the text in the Add Words box?
backend
> 6. Where is the change to the content in the Send for Review box?
ditto
> 7. Where is the change to the Change Business Type box?
> 8. Where is the "Move the designation to the end of the name" text change?
same
> 9. Where is the "Review your name as it appears..." text/content change?
likewise

client/src/components/new-request/analyze-results.vue
scottrumsby: yup

client/src/components/new-request/analyze-results.vue
scottrumsby: fixed

client/src/sass/main.sass
severinbeauvais: No, I don't think it would work for a child to specify the CSS for an element in the parent.

I agree that non-scoped CSS should not be in a SFC.

I did a global search for "reserve-submit-button" and found 4 matches... since duplicate IDs are a no-no, how about changing it to a class? Then I'll feel comfortable with the class in main.sass.

client/src/components/new-request/analyze-results.vue
```diff
@@ -78,17 +79,19 @@
...
+              <v-row class="colour-p-blue-text justify-center" dense v-if="showGreyBoxes">
+                <v-col :key="issue.issue_type + '-' + option.header + '-' + optionIndex"
+                       v-for="(option, optionIndex) of issue.setup">
+                  <GreyBox :changesInBaseName="changesInBaseName"
+                           :designationIsFixed="designationIsFixed"
+                           :i="optionIndex"
+                           :issueIndex="issueIndex"
+                           :option="option"
+                           :originalName="originalName"
+                           class="issue.setup.length === 3 && i === 1 ? 'mx-2' : ''" />
```
severinbeauvais, line 92: This won't work until you change it to `:class`.
(I've made this mistake before :P )

client/src/components/new-request/submit-request/entity-cannot-be-auto-analyzed.vue
```diff
@@ -13,7 +13,7 @@
         <v-col v-for="(box, i) in boxes" :key="'box-'+i">
           <v-container class="copy-small text-left" :class="box.class">
             <v-row align-content="space-between" style="height: 100%">
-              <v-col class="h5 py-0"><v-icon class="pr-2 colour-p-blue-text">mdi-information</v-icon>
+              <v-col class="h5 py-0"><v-icon class="pr-2 colour-p-blue-text">info</v-icon>
```
severinbeauvais, line 16: Please change this back to a MDI icon. We will transition everything to MDI icons and then stop using MD icons (for consistency between apps).

scottrumsby: ooops... that was me careless restoring from local history.  I meant to remove the icons altogether

cameron-freshworks (approved):
All good by me, once you've resolved your comments with Severin 👍

client/public/config/configuration.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "URL":"https://namex-dev.pathfinder.gov.bc.ca",
+    "URL":"https://namex-test.pathfinder.gov.bc.ca",
```
severinbeauvais, line 3: Please don't commit this. When running locally, we should be using the DEV services.

client/src/App.vue
```diff
@@ -25,6 +25,9 @@
         :redirect-on-logout="getNameRequestUrl"
       />
       <router-view />
+      <transition-group name="fade">
+        <v-btn v-for="(letter, i) of word" @click="word = 'yap'" :key="'letter'+i">{{ word }}</v-btn>
```
severinbeauvais, line 29: Is this (and the declaration for `word` debugging code?

client/src/components/new-request/analyze-characters.vue
```diff
@@ -12,21 +12,23 @@
...
-                  Your name must not begin with a special character.   The first character must be alphanumeric.
-                  Only the following special characters may be used in your name::</p>
-                <p class="my-3 text-center copy-large copy-bold">/ [ ] ^ * + - = & ( ) . , " ' # @ ! ? ; :</p>
+                <p>
+                  Review your name as it appears above and, if needed, make any necessary changes and try your ur
+                  search again.</p>
+                <p class="copy-bold colour-p-blue-text">
+                  Only the following special characters may be used in your name:</p>
+                <p class="mt-n3 copy-bold colour-p-blue-text">/ [ ] ^ * + - = & ( ) . , " ' # @ ! ? ; :</p>
+                <p class="mb-n1">Your name must not begin with a special character.</p>
```
severinbeauvais, line 26: No need to change this, but I find that it's preferable for an element to specify its top margin/padding and let elements below it worry about their whitespace. (Ie, I try to stay away from using mb/pb.)

client/src/components/new-request/analyze-characters.vue
```diff
@@ -81,3 +83,8 @@ export default class AnalyzeCharacters extends Vue {
 }
 
 </script>
+
+<style scoped lang="sass">
+.padding-50-px
+  padding: 25px 50px 25px 50px !important
```
severinbeauvais, line 89: This classname no longer makes sense.

How about just setting padding on the v-container, eg, `py-6 px-12` ? (I know these resolve to 24 48 24 48 but I think we should move to rems wherever possible.)

client/src/components/new-request/analyze-results.vue
```diff
@@ -179,16 +191,16 @@ export default class AnalyzeResults extends Vue {
     scrollingContainer: false
   }
   contents: string = ''
+  finalName: string = ''
   highlightCheckboxes: boolean = false
   issueIndex: number = 0
-  originalName: string | null = null
+  originalName: string | null = ''
```
severinbeauvais, line 197: No need to change this but FYI...

TS already knows this is a string (since you initialize it), and I've seen some linter rules complain about this redundancy, so I wouldn't declare this as a string.

Also, string can be null, so you don't need the `| null`. (In fact, any JS object can be null, and since everything is an object, I think you never need `| null`.)
https://www.w3schools.com/js/js_datatypes.asp#:~:text=In%20JavaScript%20null%20is%20%22nothing,typeof%20null%20is%20an%20object.

severinbeauvais (changes requested):
There are still some outstanding issues that should be fixed.

client/src/components/new-request/analyze-characters.vue
scottrumsby: that works for me, I was just trying to be precise per the mockups but if the built in classes are good enough then I'm all for it

client/src/components/new-request/analyze-characters.vue
scottrumsby: ooops, i was testing a transition group to make sure i was using it correctly.  Im going to rebase and force push this out of the PR

client/src/components/new-request/analyze-characters.vue
severinbeauvais: When the numbers are large enough, a couple of px here or there isn't noticeable.

I've also noticed that Tracey is a bit flexible with these things :)

client/src/sass/main.sass
scottrumsby: the components were all mutually exclusive... they could never be rendered at the same time.  but I went ahead and did as you suggested and changed it to a class

client/src/App.vue
scottrumsby: ooops, i was testing a transition group to make sure i was using it correctly. Im going to rebase and force push this out of the PR

client/src/components/new-request/analyze-results.vue
severinbeauvais: @scottrumsby Does this class property actually work?

client/public/config/configuration.json
severinbeauvais: @scottrumsby Please revert this change. Thx!

client/src/components/new-request/analyze-results.vue
scottrumsby: its works in that it doesn't do anything.  will delete it.

severinbeauvais (approved):
Looks good. Merge when you're ready.

PR #157, opened 2020-11-25
Implemented Refund modal

severinbeauvais:
*Issue #:* /bcgov/entity#4488

*Description of changes:*
- implemented refund modal component
- implemented refund summary component
- cleaned up payment history modal and summary component for consistency
- dynamically moved refund action to end of list
- added whitespace above refund button
- mapped some actions to updated text
- created and loaded SCSS assets
- implemented misc SCSS Vuetify overrides (per designs/Tracey)
- updated project version
- removed obsolete package

### DIFF
--- a/client/src/assets/scss/base.scss
+++ b/client/src/assets/scss/base.scss
@@ -1,0 +1,11 @@
+/*
+ * Base Styles
+ */
+
+@import "theme.scss";
+
+/* @font-face should go here */
+
+/* HTML element styles should go here */
+
+/* global styles should go here */
--- a/client/src/assets/scss/overrides.scss
+++ b/client/src/assets/scss/overrides.scss
@@ -1,0 +1,52 @@
+/*
+ * Vuetify Overrides
+ */
+
+@import "theme.scss";
+
+.v-application {
+  font-size: 1rem; /* 16px */
+  line-height: 1.5rem; /* 24px */
+  font-family: 'BCSans', Verdana, Arial, sans-serif !important;
+  padding: 0;
+}
+
+.v-application .text-body-1,
+.v-application .text-body-2 {
+  font-family: 'BCSans', Verdana, Arial, sans-serif !important;
+}
+
+.v-btn {
+  letter-spacing: normal;
+  border-bottom-width: 1px !important;
+}
+
+/* generic layout for dialogs */
+.v-dialog > .v-card {
+  padding: 2.5rem;
+
+  .v-card__title {
+    padding: 0;
+    font-size: 1.5rem;
+    font-weight: bold;
+    color: $gray9;
+    letter-spacing: normal;
+    margin-top: -0.5rem;
+    margin-bottom: -0.5rem;
+  }
+
+  .v-card__text {
+    padding: 2rem 0 0 0;
+    font-size: 1rem;
+    color: $gray7;
+    line-height: 1.5rem;
+  }
+
+  .v-card__actions {
+    padding: 2rem 0 0 0;
+  }
+
+  .v-btn.dialog-close {
+    background-color: unset !important;
+  }
+}
--- a/client/src/components/payment/payment-complete-modal.vue
+++ b/client/src/components/payment/payment-complete-modal.vue
@@ -16,7 +16,7 @@
         <!--<span>Time Remaining - 10:00</span>-->
         <v-spacer></v-spacer>
         <v-btn @click="hideModal" id="receipt-close-btn" class="normal" text>Done</v-btn>
-        <v-btn @click="downloadReceipt" id="download-receipt-btn" class="primary" text>Download Receipt</v-btn>
+        <v-btn @click="downloadReceipt" class="primary download-receipt-btn" text>Download Receipt</v-btn>
       </v-card-actions>
     </v-card>
   </v-dialog>
--- a/client/src/components/payment/payment-history-modal.vue
+++ b/client/src/components/payment/payment-history-modal.vue
@@ -1,19 +1,22 @@
 <template>
-  <v-dialog max-width="40%" :value="isVisible" persistent v-if="payments">
-    <v-card class="pa-9">
-      <v-card-text class="h3">Payment History</v-card-text>
-      <v-card-text class="copy-normal">
+  <v-dialog width="30rem" :value="isVisible" persistent v-if="payments">
+    <v-card>
+
+      <v-card-title>Payment History</v-card-title>
+
+      <v-card-text>
         <payment-summary v-for="summary in paymentSummaries"
-          v-bind:key="summary.id"
-          v-bind:id="summary.id"
-          v-bind:summary="summary.payment"
-          v-bind:receipt="summary.receipt"
+          :key="summary.id"
+          :id="summary.id"
+          :summary="summary.payment"
+          :receipt="summary.receipt"
         />
       </v-card-text>
-      <v-card-actions>
-        <v-spacer></v-spacer>
-        <v-btn @click="hideModal" id="history-close-btn" class="normal" text>Close</v-btn>
+
+      <v-card-actions class="justify-end">
+        <v-btn outlined id="history-close-btn" class="outlined" @click="hideModal()">Close</v-btn>
       </v-card-actions>
+
     </v-card>
   </v-dialog>
 </template>
@@ -21,7 +24,6 @@
 <script lang="ts">
 import { Component, Mixins, Watch } from 'vue-property-decorator'
 
-import RequestDetails from '@/components/common/request-details.vue'
 import PaymentSummary from '@/components/payment/payment-summary.vue'
 
 import paymentModule from '@/modules/payment'
@@ -34,7 +36,6 @@ import NameRequestMixin from '@/components/mixins/name-request-mixin'
 
 @Component({
   components: {
-    RequestDetails,
     PaymentSummary
   },
   data: () => ({
@@ -71,3 +72,9 @@ export default class PaymentHistoryModal extends Mixins(NameRequestMixin, Paymen
   }
 }
 </script>
+
+<style lang="scss" scoped>
+.payment-summary:not(:first-of-type) {
+  margin-top: 1.5rem;
+}
+</style>
--- a/client/src/components/payment/payment-summary.vue
+++ b/client/src/components/payment/payment-summary.vue
@@ -1,33 +1,33 @@
 <template>
-  <v-card class='mb-5' v-if="summary">
-    <header class='font-weight-bold px-3 py-3'>
-      <slot name='header'>
-        <span>
-          <small>{{receipt.receiptDate}}</small>
-        </span>
-        <span style="float: right" v-if="receipt">
-          <small><small>Receipt No.</small></small> {{receipt.receiptNumber}}
-        </span>
-      </slot>
+  <v-card v-if="summary" class="payment-summary">
+
+    <header class="font-weight-bold pa-3">
+      <span>
+        <small>{{receipt.receiptDate}}</small>
+      </span>
+      <span style="float: right" v-if="receipt">
+        <small><small>Receipt No.</small></small> {{receipt.receiptNumber}}
+      </span>
     </header>
 
-    <div v-show='fetchError'>
-      <v-alert color='error' icon='warning' outlined>{{fetchError}}</v-alert>
+    <div v-show="fetchError">
+      <v-alert color="error" icon="mdi-alert" outlined>{{fetchError}}</v-alert>
     </div>
 
-    <ul class='fee-list' v-show='!fetchError'>
-      <li class='container fee-list__item' v-if="receipt">
-        <div class='fee-list__item-name'>Amount</div>
-        <div class='fee-list__item-value'>${{receipt.receiptAmount.toFixed(2)}} CAD</div>
+    <ul class="fee-list" v-show="!fetchError">
+      <li class="container fee-list__item" v-if="receipt">
+        <div class="fee-list__item-name">Amount</div>
+        <div class="fee-list__item-value">${{receipt.receiptAmount.toFixed(2)}} CAD</div>
       </li>
-      <li class='container fee-list__item' v-if="summary">
-        <div class='fee-list__item-name'>Status</div>
-        <div class='fee-list__item-value'>{{summary.statusCode}}</div>
+      <li class="container fee-list__item" v-if="summary">
+        <div class="fee-list__item-name">Status</div>
+        <div class="fee-list__item-value">{{summary.statusCode}}</div>
       </li>
-      <li class='container fee-list__item'>
-        <div class='fee-list__item-name'>Receipt</div>
-        <div class='fee-list__item-value'>
-          <v-btn @click='downloadReceipt' id='download-receipt-btn' class='primary' text small>Download PDF</v-btn>
+      <li class="container fee-list__item">
+        <div class="fee-list__item-name">Receipt</div>
+        <div class="fee-list__item-value">
+          <v-btn text small class="primary download-receipt-btn"
+            :loading="loading" @click="downloadReceipt()">Download PDF</v-btn>
         </div>
       </li>
     </ul>
@@ -35,7 +35,7 @@
   </v-card>
 </template>
 
-<script lang='ts'>
+<script lang="ts">
 import { Component, Mixins, Prop } from 'vue-property-decorator'
 import '../../plugins/vuetify'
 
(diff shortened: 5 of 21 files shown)


client/src/components/payment/payment-complete-modal.vue
```diff
@@ -16,7 +16,7 @@
         <!--<span>Time Remaining - 10:00</span>-->
         <v-spacer></v-spacer>
         <v-btn @click="hideModal" id="receipt-close-btn" class="normal" text>Done</v-btn>
-        <v-btn @click="downloadReceipt" id="download-receipt-btn" class="primary" text>Download Receipt</v-btn>
```
severinbeauvais, line 19: I found duplicate "download-receipt-btn" ids so I simply changed this to a classname.

client/src/components/payment/payment-history-modal.vue
```diff
@@ -1,27 +1,29 @@
...
       </v-card-actions>
+
     </v-card>
   </v-dialog>
 </template>
 
 <script lang="ts">
 import { Component, Mixins, Watch } from 'vue-property-decorator'
 
-import RequestDetails from '@/components/common/request-details.vue'
```
severinbeauvais, line 24: Wasn't used.

client/src/components/payment/payment-history-modal.vue
```diff
@@ -1,27 +1,29 @@
 <template>
-  <v-dialog max-width="40%" :value="isVisible" persistent v-if="payments">
-    <v-card class="pa-9">
-      <v-card-text class="h3">Payment History</v-card-text>
-      <v-card-text class="copy-normal">
+  <v-dialog width="30rem" :value="isVisible" persistent v-if="payments">
```
severinbeauvais, line 2: Note: normal width is 30rem (unless you shrink the window a lot).

client/src/components/payment/payment-history-modal.vue
```diff
@@ -1,27 +1,29 @@
 <template>
-  <v-dialog max-width="40%" :value="isVisible" persistent v-if="payments">
-    <v-card class="pa-9">
-      <v-card-text class="h3">Payment History</v-card-text>
```
severinbeauvais, line 4: Note: removed classes here to use global styles instead.

client/src/components/payment/payment-summary.vue
```diff
@@ -1,41 +1,41 @@
 <template>
-  <v-card class='mb-5' v-if="summary">
-    <header class='font-weight-bold px-3 py-3'>
-      <slot name='header'>
```
severinbeauvais, line 4: Slot was not needed for HTML header element.

client/src/components/payment/payment-summary.vue
```diff
@@ -1,41 +1,41 @@
...
+      </span>
+      <span style="float: right" v-if="receipt">
+        <small><small>Receipt No.</small></small> {{receipt.receiptNumber}}
+      </span>
     </header>
 
-    <div v-show='fetchError'>
-      <v-alert color='error' icon='warning' outlined>{{fetchError}}</v-alert>
+    <div v-show="fetchError">
+      <v-alert color="error" icon="mdi-alert" outlined>{{fetchError}}</v-alert>
```
severinbeauvais, line 14: Now it displays an icon.

client/src/components/payment/payment-summary.vue
```diff
@@ -1,41 +1,41 @@
...
       </li>
-      <li class='container fee-list__item'>
-        <div class='fee-list__item-name'>Receipt</div>
-        <div class='fee-list__item-value'>
-          <v-btn @click='downloadReceipt' id='download-receipt-btn' class='primary' text small>Download PDF</v-btn>
+      <li class="container fee-list__item">
+        <div class="fee-list__item-name">Receipt</div>
+        <div class="fee-list__item-value">
+          <v-btn text small class="primary download-receipt-btn"
+            :loading="loading" @click="downloadReceipt()">Download PDF</v-btn>
```
severinbeauvais, line 30: I added "loading" state here so it looks like it's doing something while the receipt is downloading.

client/src/components/payment/payment-summary.vue
```diff
@@ -50,107 +50,46 @@ import PaymentMixin from '@/components/payment/payment-mixin'
   }
 })
 export default class PaymentSummary extends Mixins(PaymentMixin) {
-  @Prop(Number) id: number
-  @Prop(Object) summary: any
-  @Prop(Object) receipt: any
-  @Prop({ default: {
```
severinbeauvais, line 56: Incorrect object default, and wasn't used anyway, so I deleted it.

client/src/components/payment/payment-summary.vue
```diff
@@ -50,107 +50,46 @@ import PaymentMixin from '@/components/payment/payment-mixin'
...
 
-.container {
+.container.fee-list__item {
   display: flex;
   flex-flow: row nowrap;
   line-height: 1.2rem;
   font-size: 0.875rem;
-}
-
-.filing_receipt-list {
```
severinbeauvais, line 94: A bunch of classnames here weren't used, so I deleted them.

client/src/components/payment/refund-modal.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+  <v-dialog width="40rem" :value="isVisible" persistent v-if="payments">
+    <v-card>
+
+      <v-card-title class="d-flex justify-space-between">
+        <div>Cancel and Refund</div>
+        <v-btn icon large color="primary" class="dialog-close" @click="hideModal()">
```
severinbeauvais, line 7: Note the 'X' to close button. Copy this if new designs call for this.

client/src/components/payment/refund-modal.vue
```diff
@@ -0,0 +1,117 @@
...
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
+      </v-card-title>
+
+      <v-card-text class="text-body-1">
```
severinbeauvais, line 12: Note Vuetify class "text-body-1" (and text-body-2 below).

client/src/components/payment/refund-modal.vue
```diff
@@ -0,0 +1,117 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import "@/assets/scss/theme.scss";
+
+#refund-confirm-btn,
+#refund-cancel-btn {
+  text-transform: none;
+  padding-left: 1.5rem;
+  padding-right: 1.5rem;
```
severinbeauvais, line 115: Per Tracey, there is no common style for the buttons. (Many buttons are "special".)

client/src/components/payment/payment-history-modal.vue
```diff
@@ -1,27 +1,29 @@
...
+          :summary="summary.payment"
+          :receipt="summary.receipt"
         />
       </v-card-text>
-      <v-card-actions>
-        <v-spacer></v-spacer>
-        <v-btn @click="hideModal" id="history-close-btn" class="normal" text>Close</v-btn>
+
+      <v-card-actions class="justify-end">
+        <v-btn outlined id="history-close-btn" class="outlined" @click="hideModal()">Close</v-btn>
```
severinbeauvais, line 17: Note the new style for the cancel/close button.

client/src/components/payment/refund-summary.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <v-expand-transition>
+    <!-- do not display until payments are fetched -->
+    <section id="refund-summary" v-if="payments.length > 0">
```
severinbeauvais, line 4: This allows this component to "expand" once the data is loaded.

client/src/components/payment/refund-summary.vue
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /** The total of all the fees. */
+  private get total (): number {
+    const fees = _flatten(
+      // for now we only care about Filing, Priority and Service fees
+      this.lineItems.map(li => [li.filingFees, li.priorityFees, li.serviceFees])
+    )
+    const total = fees.reduce((t, n) => (t + n), 0)
+    return total
```
severinbeauvais, line 61: This returns 0 if payments is empty ([ ]).

client/src/main.ts
```diff
@@ -8,10 +8,14 @@ import KeyCloakService from 'sbc-common-components/src/services/keycloak.service
...
 // NB: order matters - do not change
 import 'quill/dist/quill.core.css'
-import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader
+import '@mdi/font/css/materialdesignicons.min.css'
 import '@/sass/main.sass'
 import '@/sass/overrides.sass'
 
+import '@/assets/scss/base.scss'
+import '@/assets/scss/layout.scss'
+import '@/assets/scss/overrides.scss'
```
severinbeauvais, line 17: When I got here, some components were using SASS and others were using SCSS.

For consistency with the other Entities UIs, I went with SCSS.

We should migrate SASS &rarr; SCSS at some point in the future.

client/src/sass/overrides.sass
```diff
@@ -72,6 +72,10 @@
   color: white !important
   background-color: $other-bg !important
 
+.v-btn.outlined
+  color: $link !important
+  background-color: white !important
```
severinbeauvais, line 77: The inverse of "primary".

client/src/components/payment/refund-summary.vue
bcdevlucas: Why use lodash for a single function instead of map + reduce, just another dep to keep up to date?

bcdevlucas (approved):
1) Left a comment regarding lodash...
2) Instead of using the string 'REFUND' there are constants in src/components/payment/payment-actions.ts?

Otherwise looks good...

client/src/components/payment/refund-summary.vue
severinbeauvais: That's a great question and one that we thought about in the other Entities projects.

As you pointed out, Lodash is already used by the UI. That can be worked around by importing just the Lodash functions used (ie, https://www.npmjs.com/package/lodash.debounce and https://www.npmjs.com/package/lodash.flatten).

But then, if you look deeper, you'll find that a lot of our dependencies use Lodash anyway (eg, Babel, TS/eslint, webpack, sbc-common-component, etc, etc).

However, I agree with you so will look at using map() instead of flatten() !

severinbeauvais:
> 1. Left a comment regarding lodash...
> 2. Instead of using the string 'REFUND' there are constants in src/components/payment/payment-actions.ts?
> 
> Otherwise looks good...

@bcdevlucas Thank you for your review. Did you notice the API call I'm using? It's not exactly what we previously discussed but I think it's the right call (pun intended). What do you think?

client/src/components/payment/refund-summary.vue
severinbeauvais: Updated in latest commit!

client/src/components/payment/refund-summary.vue
```diff
@@ -36,26 +36,22 @@
...
   @Prop({ default: () => [] })
   readonly payments: any[]
 
   /** The array of line items in all completed SBC payments. */
   private get lineItems (): any[] {
-    const lineItems = _flatten(
-      this.payments.map(p => (p.sbcPayment.statusCode === 'COMPLETED') ? p.sbcPayment.lineItems : [])
-    )
+    const arrays = this.payments.map(p => (p.sbcPayment.statusCode === 'COMPLETED') ? p.sbcPayment.lineItems : [])
+    const lineItems = [].concat(...arrays)
```
severinbeauvais, line 47: ES6 has a new array method `flat` but, according to http://caniuse.com, not all browser suport it, while `concat` is OK.
While I was there, I also confirmed that `reduce` is also well supported.

client/src/assets/scss/base.scss
```diff
@@ -0,0 +1,11 @@
...
+ * Base Styles
+ */
+
+@import "theme.scss";
+
+/* @font-face should go here */
+
+/* HTML element styles should go here */
+
+/* global styles should go here */
```
cameron-freshworks, line 11: Groundwork for the eventual migration to scss?

client/src/assets/scss/overrides.scss
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.v-application .text-body-1,
+.v-application .text-body-2 {
+  font-family: 'BCSans', Verdana, Arial, sans-serif !important;
+}
+
+.v-btn {
+  letter-spacing: normal;
+  border-bottom-width: 1px !important;
```
cameron-freshworks, line 21: How will this conflict with the existing sass overrides for `v-btn`?

client/src/assets/scss/base.scss
severinbeauvais: Yes.

severinbeauvais: What else can I say here? This project has some components using SASS and others using SCSS. It would be nice to be consistent between the Entities UIs.

client/src/components/payment/payment-history-modal.vue
```diff
@@ -1,27 +1,29 @@
 <template>
-  <v-dialog max-width="40%" :value="isVisible" persistent v-if="payments">
-    <v-card class="pa-9">
-      <v-card-text class="h3">Payment History</v-card-text>
-      <v-card-text class="copy-normal">
+  <v-dialog width="30rem" :value="isVisible" persistent v-if="payments">
+    <v-card>
+
+      <v-card-title>Payment History</v-card-title>
+
```
cameron-freshworks, line 6: Super minor, is there extra spacing here?

client/src/components/payment/refund-modal.vue
cameron-freshworks: i like it!

client/src/assets/scss/overrides.scss
severinbeauvais: Empirically, I can say that both SCSS and SASS are in play, so yes, this affects existing buttons (and this is a style update from Tracey).

Since the SCSS files are imported after the SASS files, this code overrides SASS styles.

client/src/components/payment/payment-history-modal.vue
severinbeauvais: No, I wanted each block to be separate.

cameron-freshworks: K i take that back, i prefer the spacing in the template as well. Just not the case in many of the templates but would be cool to progress towards that. So ignore the first comment :)

client/src/components/payment/refund-modal.vue
severinbeauvais: Yes, instead of using a local/custom style for the larger first para text and the smaller bottom para text, I leveraged the Vuetify styles. (I only had to override the Vuetify font face.)

client/src/components/payment/payment-summary.vue
```diff
@@ -50,107 +50,46 @@ import PaymentMixin from '@/components/payment/payment-mixin'
...
-        status_code: ''
-      }]
-    },
-    summary: {
-      completionDate: '',
-      statusCode: ''
-    }
-  } })
-  protected fetchError: string = ''
+  @Prop(Number)
```
cameron-freshworks, line 53: Curious about this. What is the significance of the type definition here if we are already doing that on the property itself?
Also, would these props be better off with a default?

severinbeauvais: 🤷

client/src/components/payment/payment-summary.vue
```diff
@@ -50,107 +50,46 @@ import PaymentMixin from '@/components/payment/payment-mixin'
...
     const { id } = this
+    this.loading = true
     await this.downloadReceiptPdf(id)
+    this.loading = false
   }
 }
 </script>
 
-<style lang='scss' scoped>
-@import '../../assets/scss/theme';
```
cameron-freshworks, line 80: I never saw the scss in the app until now. Certainly could be confusing when moving between components 😅 I know time is tight right now, but we should definitely track that we need to refactor the styling to scss app wide.

client/src/components/payment/payment-summary.vue
severinbeauvais: TBH I don't know. I tried to not change this file beyond cleanup (since this was partly the file I used as a template for the new components).

cameron-freshworks (approved):
Looks good, minor question/comments!

client/src/components/payment/payment-summary.vue
cameron-freshworks: Fair enough. We can research but definitely no action required here!

client/src/assets/scss/overrides.scss
cameron-freshworks: I wonder if it retains the properties defined earlier that are not re-defined here. For example `font-size` and `text-transform` etc etc

cameron-freshworks: My assumption, is those properties will still be retained. Either way, nothing major!

client/src/assets/scss/base.scss
cameron-freshworks: Absolutely!

client/src/components/payment/payment-summary.vue
severinbeauvais: https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5730

client/src/components/payment/payment-summary.vue
severinbeauvais: https://www.npmjs.com/package/vue-property-decorator#Prop

client/src/assets/scss/overrides.scss
severinbeauvais: Yes, this code only overrides some properties. The other properties are retained.

PR #162, opened 2020-11-26
Replace "Top 10 words to avoid" with new component

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5570

*Description of changes:*
 - refer https://projects.invisionapp.com/share/8PZBB5CDQSB#/screens/436810233_4820_NR_Homepage_Updates


severinbeauvais:
I think you need to rebase this PR... it contains commits from your previous PR!

client/src/assets/scss/base.scss
```diff
@@ -9,3 +9,11 @@
 /* HTML element styles should go here */
 
 /* global styles should go here */
+
+
+.pr-75 {
+  padding-right: 75px;
+}
+.pl-75 {
```
severinbeauvais, line 17: I see what you're trying to do and I think it's fine, but I think you should add `!important` to both of these so ensure they override any other padding specs (just as the Vuetify spacing helper classes do).

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -0,0 +1,53 @@
...
+    </v-row>
+  </v-container>
+</template>
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({
+  components: {}
+})
+export default class NewAndBetterWay extends Vue {
```
severinbeauvais, line 33: I laughed when I saw this component's name, but it really is appropriate :)

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <v-container id="new-and-better-way" class="pt-16">
+    <v-row no-gutters class="pa-9">
+      <v-col cols="6" class="pl-3 pr-75">
+        <v-img src="../../assets/images/bcnamerequest-laptop_x2.jpg" />
+      </v-col>
+      <v-col cols="6" class="pr-3 pl-75">
```
severinbeauvais, line 7: What do these columns look like when the window is resized?

With your right and left padding in here, I wonder if a different layout or alignment system would work better?

Aligning content in a col: https://stackoverflow.com/questions/57769340/vuetify-right-align-group-of-buttons

Or flexbox: https://vuetifyjs.com/en/styles/flex/

PS If it's OK currently then don't change it, but the large padding makes me wonder.

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -0,0 +1,53 @@
...
+        <v-row>
+          <v-col cols="12" class="h3 pb-5 mt-n5">
+            A new and better way to get a name <br/>for your business in B.C.
+          </v-col>
+          <v-col cols="12">
+            <ul>
+              <li><span>Find out if you can have your name right away (<i>Coming soon</i>)</span></li>
```
severinbeauvais, line 14: Why do you need the span element?

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -0,0 +1,53 @@
...
+        </v-row>
+      </v-col>
+    </v-row>
+  </v-container>
+</template>
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({
+  components: {}
```
severinbeauvais, line 31: You can omit this if you're not using any components.

severinbeauvais (approved):
A few questions about possibly improving the layout but either way it's OK with me.

client/src/components/lower-info-area/new-and-better-way.vue
vysakh-menon-aot: bullet point has to be in blue and text in black. to do that we need a separate element

client/src/components/lower-info-area/new-and-better-way.vue
vysakh-menon-aot: tracey mentioned about spacing required from the middle, min of 75px. It should be fine in big screen since we have a max-width of 1300px for the entire page (yet to implement)

severinbeauvais:
Thanks for the updates. LGTM!

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -0,0 +1,51 @@
...
+  ul
+
+    li
+      list-style-type: circle
+      color: $BCgovABlue2
+      padding: 5px 0
+
+      span
+        color: black
+</style>
```
cameron-freshworks, line 51: Small change but perhaps convert this to SCSS entirely. We've slowly started doing this, and it will ease the burden if we do it 1by1 as we go :)

vysakh-menon-aot: sure

cameron-freshworks (approved):
Just one comment! Might be good to do that update now, but otherwise g2g :)

PR #173, opened 2020-12-02
fixes according to ux verification

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5569

*Description of changes:*

### DIFF
--- a/client/src/assets/scss/theme.scss
+++ b/client/src/assets/scss/theme.scss
@@ -27,7 +27,6 @@ $BCgovABlue1: #6e93ff;
 $BCgovABlue2: #3b6cff;
 $BCgovABlue3: #0846ff;
 $BCgovABlue4: #003bee;
-$BCgovABlue5: #1976D2;
 
 // BC GOV GOLD PALETTE
 $BCgovGold0: #fff7e3;
@@ -53,6 +52,8 @@ $text: $gray7;
 
 // Some popular colors
 $app-blue: #1669bb;
+$app-md-blue: #1976D2;
+$app-light-blue: #e3f0f7;
 $app-red: #d3272c;
 $app-green: #2e8540;
 $border: #899eb2;
--- a/client/src/assets/scss/theme.scss
+++ b/client/src/assets/scss/theme.scss
@@ -27,6 +27,7 @@ $BCgovABlue1: #6e93ff;
 $BCgovABlue2: #3b6cff;
 $BCgovABlue3: #0846ff;
 $BCgovABlue4: #003bee;
+$BCgovABlue5: #1976D2;
 
 // BC GOV GOLD PALETTE
 $BCgovGold0: #fff7e3;
--- a/client/src/components/lower-info-area/name-build-sub-components/example-name.vue
+++ b/client/src/components/lower-info-area/name-build-sub-components/example-name.vue
@@ -2,7 +2,7 @@
   <v-container id="example-name">
     <v-row>
       <v-col cols="5" class="mx-3">
-        <v-row>
+        <v-row justify="center">
           <img src="@/assets/images/Sarahsx1.png" class="name-img" />
         </v-row>
         <v-row align="center" justify="center">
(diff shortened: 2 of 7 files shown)


client/src/assets/scss/theme.scss
```diff
@@ -27,6 +27,7 @@ $BCgovABlue1: #6e93ff;
 $BCgovABlue2: #3b6cff;
 $BCgovABlue3: #0846ff;
 $BCgovABlue4: #003bee;
+$BCgovABlue5: #1976D2;
```
severinbeauvais, line 30: The accent colours are ordered. Could you update please?

client/src/components/lower-info-area/name-build-sub-components/example-name.vue
```diff
@@ -62,9 +62,10 @@ export default class ExampleName extends Vue {}
     margin-bottom: 0;
   }
 
-  .name-img {
-    width: 100%;
-    height: 50px;
+  @media (max-width: 1299px) {
+    .name-img {
+      height: 50px;
```
severinbeauvais, line 67: Is the height OK if screen > 1299px ?

client/src/assets/scss/theme.scss
vysakh-menon-aot: How to do that? by renaming $BCgovABlue4 -> $BCgovABlue5? I did not understand that

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div id="new-and-better-way" class="pt-16">
+  <div id="new-and-better-way" class="pt-16 pb-4">
```
severinbeauvais, line 2: I have found that a best practise is to add top padding (or margin) to the element below this one -- this prevents THIS element from "assuming" how much whitespace is required below it. 

(Whereas the element below knows how much whitespace to have above it.)

Can this be fixed here?

client/src/components/lower-info-area/name-build-sub-components/example-name.vue
vysakh-menon-aot: Yes. Confirmed with Tracey

client/src/components/lower-info-area/new-and-better-way.vue
```diff
@@ -42,11 +42,11 @@ export default class NewAndBetterWay extends Vue {
   ul {
     li {
       list-style-type: circle;
-      color: $BCgovABlue2;
+      color: $BCgovABlue5;
       padding: 5px 0;
 
       span {
-        color: black;
+        color: $gray7;
```
severinbeauvais, line 49: Please change to `$text`.

severinbeauvais (approved):
A few comment but once resolved I approve this.

client/src/components/lower-info-area/new-and-better-way.vue
vysakh-menon-aot: It wont work here since the background color of this div and the div below is different

client/src/assets/scss/theme.scss
severinbeauvais: Yeah, sorry, I was unclear.

Actually, those "BCGov" colours are from the official BC Gov theme (outside this project). So we should not change it.

I suggest two options:
1. Get Tracey to agree to use one of the existing BC Gov accent colours, or
2. Create a new variable `$app-lt-blue` that we can all reuse.

severinbeauvais: While you're there, could you please use a colour variable for `.bg-light-blue` ? (This is cleanup.)

severinbeauvais: (That way we keep all our colours in one place!)

severinbeauvais: (And it will be easier to find a colour for reuse, and also reduce the number of colours we use!)

client/src/components/lower-info-area/new-and-better-way.vue
severinbeauvais: OK, you found the exception to my rule :)

severinbeauvais: (Although to be technical, it's not whitespace if it's a different colour ;) )

client/src/assets/scss/theme.scss
vysakh-menon-aot: will do

cameron-freshworks (approved):
Looks good!

PR #176, opened 2020-12-03
UX-Beta tooltip

cameron-freshworks:
*Issue #:* /bcgov/entity#5542

*Description of changes:*

* Added a Beta tag tool tip

* Decreased some spacing in the stats counters

### DIFF
--- a/client/src/views/landing.vue
+++ b/client/src/views/landing.vue
@@ -6,7 +6,17 @@
           <Stats class="pr-10"/>
         </v-col>
         <v-col cols="12" class="mb-n2">
-          <span class="h2 colour-white">Name Request <sup class="beta-tag">Beta</sup></span>
+          <span class="h2 colour-white">Name Request
+            <v-tooltip bottom nudge-right="10" content-class="bottom-tooltip" transition="fade-transition">
+              <template v-slot:activator="{ on }">
+                <sup class="beta-tag" v-on="on">Beta</sup>
+              </template>
+              <p>The Name Request website is available as a Beta version. Name Requests obtained through the Beta are
+                official name requests and can be used in the province of British Columbia.</p>
+              <p>As part of a new <u>agile software development process</u>, the Name Request website will be
+                continually updated and improved based on feedback from citizens and businesses in BC.</p>
+            </v-tooltip>
+          </span>
         </v-col>
         <div class="main-container-style mt-3">
           <transition name="fade" mode="out-in" :duration="{ enter: 100, leave: 100 }">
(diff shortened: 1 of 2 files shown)


client/src/views/landing.vue
```diff
@@ -6,7 +6,17 @@
...
         <v-col cols="12" class="mb-n2">
-          <span class="h2 colour-white">Name Request <sup class="beta-tag">Beta</sup></span>
+          <span class="h2 colour-white">Name Request
+            <v-tooltip bottom nudge-right="10" content-class="bottom-tooltip" transition="fade-transition">
+              <template v-slot:activator="{ on }">
+                <sup class="beta-tag" v-on="on">Beta</sup>
+              </template>
+              <p>The Name Request website is available as a Beta version. Name Requests obtained through the Beta are
+                official name requests and can be used in the province of British Columbia.</p>
+              <p>As part of a new <u>agile software development process</u>, the Name Request website will be
```
severinbeauvais, line 16: 👍

vysakh-menon-aot (approved):

client/src/views/landing.vue
```diff
@@ -76,25 +86,28 @@ export default class Landing extends Vue {
...
-
-.beta-tag
-  top: -1rem
-  color: $gold
-  font-size: 19px
-  font-weight: bold
-  text-transform: uppercase
+#landing-container {
+  margin: 0;
+  padding: 0;
```
severinbeauvais, line 94: This will be easier to clean up in the future if you use `ma-0 pa-0` in its class instead.

client/src/views/landing.vue
```diff
@@ -76,25 +86,28 @@ export default class Landing extends Vue {
 }
 </script>
 
-<style lang="sass" scoped>
-#landing-container
-  margin: 0
-  padding: 0
+<style lang="scss" scoped>
```
severinbeauvais, line 89: 👍 👍

severinbeauvais:
I saw the beta tag tooltip and a spacing update. What about the other check items in the ticket?

client/src/views/landing.vue
cameron-freshworks: Yea, this is actually pretty helpful info!

client/src/views/landing.vue
cameron-freshworks: ah good call!

severinbeauvais: Yeah, I can see where we might want to redefine, say, containers, to not have negative margins or whatever, and clean up a bunch of the HTML.
(Probably not as an explicit ticket but as cleanup in other tickets.)

cameron-freshworks:
> I saw the beta tag tooltip and a spacing update. What about the other check items in the ticket?

The other was width on the tooltip, to keep it one line which is in there. 
Last item i think was the opacity, which is the same as all the other tooltips. Perhaps it is because the background transitions behind these tooltips, it seems more transparent. not sure 🤔 

Either way, it is the same opacity, currently handled by the vuetify transition itself. I'm honestly not sure if there is a prop to change the opacity, we might have to write a custom vuetify transition if wanted to adjust that.

client/src/views/landing.vue
cameron-freshworks: Absolutely, agree 100%. Updated!

cameron-freshworks: Those class helpers have changed my life 😆

severinbeauvais (approved):

PR #190, opened 2020-12-08
Stopped handling invalid states + updated "Consumed" and "Expired"

severinbeauvais:
*Issue #:* /bcgov/entity#5816

*Description of changes:*
- updated logic for Conditions link
- stopped handling COMPLETED state
- stopped handling COND_RESERVED state
- stopped handling RESERVED state
- inferred Consumed differently
- inferred Expired differently
- removed fallback for unhandled states
- fixed minor UX issues


client/src/components/existing-request/existing-request-display.vue
```diff
@@ -323,48 +330,66 @@ export default class ExistingRequestDisplay extends Mixins(NrAffiliationMixin, C
...
+      this.hasExpirationDatePassed
+    )
   }
 
-  /** True if NR is in Completed state. */
-  private get isNrCompleted (): boolean {
-    return (this.nr.state === NrState.COMPLETED)
+  /** True if NR is in Approved or Conditional state. */
+  private get isNrApprovedOrConditional (): boolean {
+    return [NrState.APPROVED, NrState.CONDITIONAL].includes(this.nr.state)
```
severinbeauvais, line 380: Note that this is a getter so it's cached -- so it's only calculated once!

client/src/enums/nr-state.ts
```diff
@@ -12,5 +12,5 @@ export enum NrState {
   REFUND_REQUESTED = 'REFUND_REQUESTED',
   REFUNDED = 'REFUNDED', // NOT YET IMPLEMENTED
   REJECTED = 'REJECTED',
-  RESERVED = 'RESERVED',
+  RESERVED = 'RESERVED', // INTERIM STATE
```
severinbeauvais, line 15: At a meeting today with Kaine, Janis and Lorna, it was said that COND_RESERVED and RESERVED are interim states that the UI will never see, however I noticed some UI pages that check for these states.

Perhaps it was meant that the "Existing Request Display" page will never see these (since the user will not be able to query the NR using its temp number) (the NR has only a temp number at this point).

client/src/enums/nr-state.ts
```diff
@@ -2,8 +2,8 @@
 export enum NrState {
   APPROVED = 'APPROVED',
   CANCELLED = 'CANCELLED',
-  COMPLETED = 'COMPLETED',
-  COND_RESERVED = 'COND_RESERVE',
+  COMPLETED = 'COMPLETED', // NOT A VALID STATE
```
severinbeauvais, line 5: We should remove this as soon as the BE removes it first :)

cameron-freshworks (approved):
Looks good 👍

vysakh-menon-aot (approved):

lekshmimallika-aot (approved):

PR #193, opened 2020-12-08
5747 - Existing Request Search Screen

scottrumsby:
*Issue #:* /bcgov/entity#5747

*Description of changes:*

- Made the changes from the first slide of existing request search design document


severinbeauvais:
According to the InVision, the Retrieve Name Request button should stay blue and should activate validation on the required fields.
However, when I land on this page, this button is disabled.

severinbeauvais:
1. The Invision specs out a gutter of ~35px~ 40px around everything. However, there is extra left/right whitespace around the content. Please remove it.

2. The top text needs to be updated. (You can leave the : at the end.)

3. See comment above regarding the Retrieve button. I confirmed with Tracey that this button should always be enabled (as it is not AA compliant when disabled). Clicking it should show any invalid/incomplete fields on the page (and obviously not submit if invalid).

severinbeauvais (changes requested):
See comments above (which Tracey and I went through together).

client/src/components/existing-request/existing-request-search.vue
```diff
@@ -10,50 +10,53 @@
...
-                       @input="setExistingRequestSearch('nrNum', $event)"
-                       filled
-                       validate-on-blur
-                       :rules="nrRules"
-                       placeholder="Business Name or NR Number"
-                       id="nr-num-text-field" />
+                      @input="setExistingRequestSearch('nrNum', $event)"
+                      class="copy-normal"
+                      filled
+                      validate-on-blur
```
cameron-freshworks, line 23: What does this prop do? Sounds fancy!

cameron-freshworks (commented):
From a code standpoint it looks good, tho I think Severin and Tracey had some UX/design tweaks to be made 👍

client/src/components/existing-request/existing-request-search.vue
scottrumsby: validate-on-blur?  it prevents the field validation from happening until after the field is removed from the focus so the red error messages don't show up from the first character entered (useful if the field has a length rule for instance)

cameron-freshworks: Ah, very cool! Thanks sir

cameron-freshworks (approved):

severinbeauvais:
I have verified the changes - looks good, thanks!

severinbeauvais (approved):

PR #202, opened 2020-12-09
Auto-analyze fixes - Timeout + names longer than 3 words

lekshmimallika-aot:
*Issue #:* /bcgov/entity#5818, /bcgov/entity#5861

*Description of changes:*
Update in search pending text
Showing generic send to review screen for names longer than 4 words
Handle gateway timeout error

### DIFF
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -1230,6 +1230,17 @@ export class NewRequestModule extends VuexModule {
     }
     return false
   }
+
+  get hasNameMoreThanThreeWords () {
+    if (this.name) {
+      let { name } = this
+      if (name.split(' ').length > 3) {
+        return true
+      }
+    }
+    return false
+  }
+
   get nrId () {
     const { nr } = this
     let nrId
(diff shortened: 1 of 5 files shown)


client/src/store/new-request-module.ts
```diff
@@ -1812,12 +1823,10 @@ export class NewRequestModule extends VuexModule {
...
     } catch (error) {
-      // eslint-disable-next-line
-      console.log(error)
-      if (error.code === 'ECONNABORTED') {
+      if (error.code === 'ECONNABORTED' || error.message === 'Network Error') {
         this.mutateNameAnalysisTimedOut(true)
-        this.mutateSubmissionTabComponent('EntityNotAutoAnalyzed')
-        this.mutateDisplayedComponent('SubmissionTabs')
+        this.mutateName(this.name)
+        this.mutateDisplayedComponent('SendToExamination')
```
severinbeauvais, line 1829: So, if the request times out then we display the Send To Examination page?

severinbeauvais (approved):

kialj876 (approved):

client/src/store/new-request-module.ts
lekshmimallika-aot: Yes. It will search upto 2 minutes now. If we are not able to get the result within that time it shows the new generic screen @cameron-freshworks has built (SendToExamination) that says Entity cannot be auto-analyzed. From there the user can navigate to send for review by clicking Continue Button

PR #203, opened 2020-12-10
Updated Receipts Modal

severinbeauvais:
*Issue #:* /bcgov/entity#5755

*Description of changes:*
- renamed Payment History Modal -> Receipts Modal
- deleted obsolete Vuetify imports
- created Receipt Mixin
- updated Payment Summary component
- updated Receipts Modal
- misc cleanup


client/src/components/payment/refund-summary.vue
```diff
@@ -56,13 +53,9 @@ export default class RefundSummary extends Vue {
     return total
   }
 
-  /** Returns the line item description mapped to a user-friendly name. */
-  private mapItemName (item: any): string {
```
severinbeauvais, line 60: I moved this to a mixin.

client/src/components/mixins/receipt-mixin.ts
```diff
@@ -0,0 +1,13 @@
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component
+export default class ReceiptMixin extends Vue {
+  /** Returns the line item description mapped to a user-friendly name. */
+  rcptDescToName (description: string): string {
+    switch (description) {
+      case 'Reg. Submission Online': return 'Name Request'
+      case 'Upgrade to Priority': return 'Priority Request'
```
cameron-freshworks, line 9: Are these passed from the back end?

client/src/components/payment/payment-summary.vue
```diff
@@ -1,70 +1,53 @@
...
+    return `Receipt No. ${this.summary?.receipt.receiptNumber}`
+  }
 
-  private loading = false
+  private get receiptDate (): string {
+    return this.summary?.receipt.receiptDate
+  }
+
+  private get receiptDescription (): string {
+    const lineItem = this.summary?.payment.sbcPayment.lineItems[0] // just look at first one
```
cameron-freshworks, line 41: Should we optional chain `payment` and `sbcPayment` as well?

client/src/components/payment/payment-summary.vue
```diff
@@ -1,70 +1,53 @@
...
+    return this.rcptDescToName(lineItem?.description)
+  }
 
-  async downloadReceipt () {
-    const { id } = this
+  private get receiptAmount (): string {
+    return `${this.summary?.receipt.receiptAmount.toFixed(2)} CAD`
```
cameron-freshworks, line 46: Maybe same here, could `receipt` or `receiptAmount` have a chance of coming back null?

cameron-freshworks (approved):
Looks really good! Just had a few questions.

client/src/components/mixins/receipt-mixin.ts
severinbeauvais: Yes. It's part of the SBC payment line items that the BE (either name or sbc-pay, I'm not sure) has prepared.

client/src/components/payment/payment-summary.vue
severinbeauvais: "Summary" needs to be check for null/undefined because this modal is created when the app starts (not just when we ask to display it).

I don't think there's much to be gained in checking "payment" or sbcPayment. The previous code did not check it. Lots of other code will fail if those are null/undefined. I think it's OK to assume those objects are present.

client/src/components/payment/payment-summary.vue
severinbeauvais: Anything is possible... The old code didn't check this so I didn't add it. I think it's probably OK.

client/src/components/mixins/receipt-mixin.ts
cameron-freshworks: Copy that!

vysakh-menon-aot (approved):

PR #214, opened 2020-12-11
Updates to check if name request needed modal

lekshmimallika-aot:
*Issue #:* /bcgov/entity#5547

*Description of changes:*

### DIFF
--- a/client/src/components/modals/nr-not-required.vue
+++ b/client/src/components/modals/nr-not-required.vue
@@ -1,11 +1,9 @@
 <template>
   <!-- NB: dummy modal attach point so unit tests can see it -->
   <v-dialog v-model="showModal" width="50rem" attach="">
-    <v-card class="pa-9 nr-not-required-dialog">
-      <v-card-title>
-        <span>
-          <slot name="title">You don't need a Name Request if...</slot>
-        </span>
+    <v-card>
+      <v-card-title class="d-flex justify-space-between">
+        <div>You don't need a Name Request if...</div>
       </v-card-title>
       <v-card-text class="copy-normal">
         <ol>
@@ -29,7 +27,7 @@
           </li>
         </ol>
       </v-card-text>
-      <v-card-actions>
+      <v-card-actions class="justify-center">
         <v-btn text id="nr-required-close-btn" @click="showModal = false">Close</v-btn>
       </v-card-actions>
     </v-card>
@@ -56,9 +54,6 @@ export default class NrNotRequired extends Vue {
 </script>
 
 <style lang="scss" scoped>
-  .nr-not-required-dialog .v-card__actions {
-    justify-content: center;
-  }
   #nr-required-close-btn {
     width: 6.375rem;
   }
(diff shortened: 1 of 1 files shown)


client/src/components/modals/nr-not-required.vue
```diff
@@ -1,29 +1,35 @@
 <template>
   <!-- NB: dummy modal attach point so unit tests can see it -->
-  <v-dialog v-model="showModal" max-width="40%" attach="">
-    <v-card class="pa-9">
-      <v-card-text class="h3">You don't need a Name Request if...</v-card-text>
+  <v-dialog v-model="showModal" width="50rem" attach="">
+    <v-card class="pa-9 nr-not-required-dialog">
```
severinbeauvais, line 4: Invalid class. This modal should look like the Receipts modal, Cancel modal, etc. (40px gutter all around)

client/src/components/modals/nr-not-required.vue
```diff
@@ -1,29 +1,35 @@
...
+      <v-card-title>
+        <span>
+          <slot name="title">You don't need a Name Request if...</slot>
```
severinbeauvais, line 7: I don't get it. Why do you need to specify the slot? What's wrong with the old code?

client/src/components/modals/nr-not-required.vue
```diff
@@ -1,29 +1,35 @@
...
+        </span>
+      </v-card-title>
       <v-card-text class="copy-normal">
         <ol>
           <li>
             <b>You are already Federally Incorporated.</b><br>
-            You do not require a name request. Please proceed to COLIN to extra-provincially register your business.
+            You do not require a name request. Please go to <a :href="corpOnlineLink" target="_blank">
```
severinbeauvais, line 14: For all these links, was there a requirement to show a "mdi-launch" icon (to indicate this will open in a new page)?
If so, see names-capture.vue.

client/src/components/modals/nr-not-required.vue
lekshmimallika-aot: h3 style is not the one to be used. If you put the plain text under v-card-title it is not picking the style that is added in the overrides stylesheet.

client/src/components/modals/nr-not-required.vue
lekshmimallika-aot: No icon for these links

client/src/components/modals/nr-not-required.vue
severinbeauvais: Shouldn't this modal look like the other modals? Please compare with, say, the Cancel Name Request modal.

lekshmimallika-aot: Updated to be in sync with cancel and receipts modal

client/src/components/modals/nr-not-required.vue
```diff
@@ -45,3 +52,12 @@ export default class NrNotRequired extends Vue {
...
 
 </script>
+
+<style lang="scss" scoped>
+  #nr-required-close-btn {
+    width: 6.375rem;
+  }
+  a {
+    text-decoration: underline;
+  }
```
cameron-freshworks, line 62: Very minor, an extra indent here. I think in this project most of the scss start on the same indent as the style tag.

cameron-freshworks (approved):
Looks good! Just one small comment, but your g2g when you're ready

PR #220, opened 2020-12-13
Submission & Pay Modal Loading Indicators

cameron-freshworks:
*Issue #:* /bcgov/entity#5878

*Description of changes:*

* Updated placeholders to labels to be inline with other UI Form elements.

* Removed some redundant form messages.

* Applied loading indicator to the submission btn while processing network request.

* Applied loading indicator to the payment modal pay btn while processing network request.

### DIFF
--- a/client/src/components/common/applicant-info-nav.vue
+++ b/client/src/components/common/applicant-info-nav.vue
@@ -10,6 +10,7 @@
     <v-btn x-large
            @click="next"
            :disabled="!isValid"
+           :loading="isloadingSubmission"
            :id="`submit-continue-btn-${isValid}`">
       {{ nextText }}
     </v-btn>
@@ -25,6 +26,7 @@ import timerModule from '@/modules/vx-timer'
 @Component({})
 export default class ApplicantInfoNav extends Vue {
   @Prop(Boolean) isValid: boolean
+  private isloadingSubmission: boolean = false
 
   get backText () {
     if (this.editMode) {
@@ -76,6 +78,7 @@ export default class ApplicantInfoNav extends Vue {
   }
   next () {
     if (this.tab === 3) {
+      this.isloadingSubmission = true
       this.submit()
       return
     }
@@ -96,6 +99,7 @@ export default class ApplicantInfoNav extends Vue {
           let request = await newReqModule.getNameRequest(nrId)
           if (request.stateCd === 'CANCELLED') {
             newReqModule.setActiveComponent('Timeout')
+            this.isloadingSubmission = false
             return
           }
         }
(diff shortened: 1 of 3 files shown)


client/src/components/common/applicant-info-nav.vue
```diff
@@ -76,6 +78,7 @@ export default class ApplicantInfoNav extends Vue {
   }
   next () {
     if (this.tab === 3) {
+      this.isloadingSubmission = true
       this.submit()
```
severinbeauvais, line 82: Should you await submit()?

client/src/components/payment/payment-modal.vue
```diff
@@ -113,6 +118,7 @@ export default class PaymentModal extends Mixins(
   }
 
   async confirmPayment () {
+    this.isLoadingPayment = true
```
severinbeauvais, line 121: I think this needs to be set to False if the user cancels this modal.

severinbeauvais (approved):
A couple of comments but when you're satisfied, I approve this for merging.

vysakh-menon-aot (approved):

client/src/components/common/applicant-info-nav.vue
cameron-freshworks: I think it was 'ok' because the functionality within `submit()` was asynchronous, but i don't see an issue with awaiting it entirely before returning. Probably safer? Will update

client/src/components/payment/payment-modal.vue
cameron-freshworks: I think with vue modals, each time we close the modal, it's vue instance is destroyed and it's data along with it. Since this loading property is local, it should default back to `false` just by closing/cancelling it.

cameron-freshworks: Will test a few more times though, but so far it seems to be ok.

cameron-freshworks: But, actually now that you bring this up. Do you think we should disable cancel once they have initiated payment? 
If you click `Pay` but cancel immediately after, it still goes to pay.

client/src/components/common/applicant-info-nav.vue
severinbeauvais: If the functionality within submit() is async then THIS should await it. It isn't causing an error atm because this is the end of this method (ie, there is nothing after this statement), but fixing it might save a problem later :)

client/src/components/payment/payment-modal.vue
severinbeauvais: 1. A modal is just a component. If the component was destroyed between instances then, yes, its data is cleared, and when it is re-instantiated then created() and mounted() are called, etc.

However, I think in this project, all modals are created once at app start and are never really destroyed, only hidden.

2. Definitely agree with you! Make it so. :)

cameron-freshworks: Hmm interesting, yea it seems like it is resetting somehow in the lifecycle of the modal, but i can certainly put in a safety reset!

Done and done 👍

PR #240, opened 2020-12-17
Error Handling on `addRequestActionComment`

cameron-freshworks:
*Issue #:* /bcgov/entity#5981

*Description of changes:*

* Throw an error instead of returning the data if a failure occurs in the `addRequestActionComment` method.

* Minor adjustments to the Error modal.

* Short circuited the network requests, to not make the call if the data had any failures when structured.

### DIFF
--- a/client/src/components/common/error-modal.vue
+++ b/client/src/components/common/error-modal.vue
@@ -2,14 +2,11 @@
   <v-dialog v-model="showModal">
     <v-card>
       <v-card-title class="d-flex justify-space-between">
-        <div>Name Request Encountered an Error</div>
-        <v-btn icon large class="dialog-close" @click="hideModal()">
-          <v-icon>mdi-close</v-icon>
-        </v-btn>
+        <div>Name Request encountered an error</div>
       </v-card-title>
 
       <v-card-text class="copy-normal">
-        <ul v-if="hasErrors">
+        <ul v-if="hasErrors" class="ml-n1">
           <li v-for="error in errors" :key="error.id" v-html="error.error" />
         </ul>
       </v-card-text>
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -1931,8 +1931,6 @@ export class NewRequestModule extends VuexModule {
    */
   @Action
   async confirmAction (action: string): Promise<boolean> {
-    // eslint-disable-next-line no-console
-    console.log('NetWork Call 1 ')
     try {
       const nrData = await this.getNameRequest(this.nr.id)
       if (!nrData) throw new Error('Got error from getNameRequest()')
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -1992,7 +1992,8 @@ export class NewRequestModule extends VuexModule {
   }
 
   @Action
-  addRequestActionComment (data) {
+  async addRequestActionComment (data) {
+    data = undefined
     try {
       let requestAction = this.requestActionOriginal || this.request_action_cd
       let { shortDesc } = this.requestActions.find(request => request.value === requestAction)
@@ -2019,7 +2020,10 @@ export class NewRequestModule extends VuexModule {
       return data
     } catch (error) {
       console.error('addRequestActionComment() =', error) // eslint-disable-line no-console
-      throw new Error(`addRequestActionComment() = ${error}`)
+      await errorModule.setAppError(
+        { id: 'add-request-action-error', error: 'An error occurred when building the name request' }
+      )
+      return null
     }
   }
 
@@ -2180,9 +2184,9 @@ export class NewRequestModule extends VuexModule {
     try {
       const { nrId } = this
       const nr = this.editNameReservation
-      const requestData = await this.addRequestActionComment(nr)
+      const requestData = nr && await this.addRequestActionComment(nr)
 
-      const response = await axios.patch(`/namerequests/${nrId}/edit`, requestData, {
+      const response = requestData && await axios.patch(`/namerequests/${nrId}/edit`, requestData, {
         headers: {
           'Content-Type': 'application/json'
         }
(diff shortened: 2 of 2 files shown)


severinbeauvais:
Does this also solve /bcgov/entity#5978 ?

vysakh-menon-aot (approved):

client/src/components/common/error-modal.vue
```diff
@@ -2,14 +2,11 @@
   <v-dialog v-model="showModal">
     <v-card>
       <v-card-title class="d-flex justify-space-between">
-        <div>Name Request Encountered an Error</div>
-        <v-btn icon large class="dialog-close" @click="hideModal()">
```
severinbeauvais, line 6: Did Tracey ask for this?

cameron-freshworks: Sorry i should have mentioned, yes she did. 
As well as nudging the bullet point a little.

cameron-freshworks:
> Does this also solve /[bcgov/entity#5978](https://github.com/bcgov/entity/issues/5978) ?

Ah interesting. Let me run through all these scenarios!

client/src/store/new-request-module.ts
```diff
@@ -2019,7 +2019,7 @@ export class NewRequestModule extends VuexModule {
       return data
     } catch (error) {
       console.error('addRequestActionComment() =', error) // eslint-disable-line no-console
-      return data
+      throw new Error(`addRequestActionComment() = ${error}`)
```
severinbeauvais, line 2022: Don't do this. It will cause a chain of errors going back to the `@Action`.

Return null if you must, and handle that instead.

cameron-freshworks: ok, will try that.

severinbeauvais (changes requested):
Don't throw an error in `@Action`.

cameron-freshworks:
> Does this also solve /[bcgov/entity#5978](https://github.com/bcgov/entity/issues/5978) ?

As long as we catch and `console.error()` (line 2021), we will still get the various errors that could occur in the many different places prior to this method.  

For example if there is any javascript error or bad data etc.. in either `draftNameReservation`, `conditionalNameReservation` or `reservedNameReservation` it will become apparent and caught in `addRequestActionComment`. 

This doesn't solve the Sentry errors... as long as we console.error() they will show up.

severinbeauvais:
> > Does this also solve /[bcgov/entity#5978](https://github.com/bcgov/entity/issues/5978) ?
> 
> As long as we catch and `console.error()` (line 2021), we will still get the various errors that could occur in the many different places prior to this method.
> 
> For example if there is any javascript error or bad data etc.. in either `draftNameReservation`, `conditionalNameReservation` or `reservedNameReservation` it will become apparent and caught in `addRequestActionComment`.
> 
> This doesn't solve the Sentry errors... as long as we console.error() they will show up.

I was hoping you had a code fix that resolved the JS errors...? 5978 cover four such errors.

client/src/store/new-request-module.ts
```diff
@@ -2180,9 +2183,9 @@ export class NewRequestModule extends VuexModule {
     try {
       const { nrId } = this
       const nr = this.editNameReservation
-      const requestData = await this.addRequestActionComment(nr)
+      const requestData = nr && await this.addRequestActionComment(nr)
 
-      const response = await axios.patch(`/namerequests/${nrId}/edit`, requestData, {
+      const response = requestData && await axios.patch(`/namerequests/${nrId}/edit`, requestData, {
```
severinbeauvais, line 2188: What happens to the code below (eg, line 2194) if response is undefined (because requestData is falsy)?

client/src/store/new-request-module.ts
```diff
@@ -2253,9 +2256,9 @@ export class NewRequestModule extends VuexModule {
           break
       }
 
-      const requestData: any = await this.addRequestActionComment(data)
+      const requestData: any = data && await this.addRequestActionComment(data)
       try {
-        const response: AxiosResponse = await axios.post(`/namerequests`, requestData, {
+        const response: AxiosResponse = requestData && await axios.post(`/namerequests`, requestData, {
```
severinbeauvais, line 2261: Same comment here. See line 2266.

cameron-freshworks:
> > > Does this also solve /[bcgov/entity#5978](https://github.com/bcgov/entity/issues/5978) ?
> > 
> > 
> > As long as we catch and `console.error()` (line 2021), we will still get the various errors that could occur in the many different places prior to this method.
> > For example if there is any javascript error or bad data etc.. in either `draftNameReservation`, `conditionalNameReservation` or `reservedNameReservation` it will become apparent and caught in `addRequestActionComment`.
> > This doesn't solve the Sentry errors... as long as we console.error() they will show up.
> 
> I was hoping you had a code fix that resolved the JS errors...? 5978 cover four such errors.

Ok, we can work through that.. though i would prefer another commit at it seems like a top priority to at least prevent the fees from generating, in the event of some failures. 

Are you ok if i do that next, but in another commit?

client/src/store/new-request-module.ts
```diff
@@ -2307,10 +2311,10 @@ export class NewRequestModule extends VuexModule {
         data['corpNum'] = this.corpNum
       }
 
-      const requestData = await this.addRequestActionComment(data)
+      const requestData = data && await this.addRequestActionComment(data)
 
       try {
-        const response = await axios.put(`/namerequests/${nrId}`, requestData, {
+        const response = requestData && await axios.put(`/namerequests/${nrId}`, requestData, {
```
severinbeauvais, line 2317: One more.

I guess this will cause a JS "undefined" error that will be caught by the catch, but this isn's ideal. Would be better to handle this scenario with an appropriate error (or log) and return false, or something along those lines.

severinbeauvais:
> Are you ok if i do that next, but in another commit?

Yes. But see my other comments.

client/src/store/new-request-module.ts
cameron-freshworks: This is indeed the case, as with all of them. They are caught when the request doesn't occur and the error is boiled up from the point of failure. 
This still needs work for sure, to flesh out and isolate these (on top of fixing the individual errors if we can)

severinbeauvais (approved):
I agree to proceed with this changes ASAP.
And then to look at my comment re: JS exceptions.

cameron-freshworks:
> I agree to proceed with this changes ASAP.
> And then to look at my comment re: JS exceptions.

Copy that. I will cut a branch specific to that ticket and address those asap.

client/src/store/new-request-module.ts
severinbeauvais: OK, create a ticket or add to the description of an exsting ticket to track this work. Thanks!

PR #248, opened 2020-12-18
Form Validation Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#6011

*Description of changes:*

* Updated form validation fields

### DIFF
--- a/client/src/components/common/applicant-info-1.vue
+++ b/client/src/components/common/applicant-info-1.vue
@@ -539,9 +539,6 @@ export default class ApplicantInfo1 extends Vue {
     newReqModule.getAddressDetails(id)
   }
   updateApplicant (key, value) {
-    // eslint-disable-next-line no-console
-    console.log('inside updateApplicant')
-    // this.clearValidation()
     newReqModule.updateApplicantDetails({ key, value })
     if (key === 'addrLine1') {
       this.showAddressMenu = true
--- a/client/src/components/common/applicant-info-1.vue
+++ b/client/src/components/common/applicant-info-1.vue
@@ -27,7 +27,6 @@
                             :rules="firstLastNameRules"
                             :value="applicant.firstName"
                             @blur="messages = {}"
-                            @focus="handleFocus('firstName', 'First Name')"
                             @input="mutateApplicant('firstName', $event)"
                             dense
                             filled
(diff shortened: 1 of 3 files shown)


client/src/components/common/applicant-info-1.vue
```diff
@@ -440,12 +437,6 @@ export default class ApplicantInfo1 extends Vue {
   get provinceOptions () {
     return this.$canJurisdictions.map(jurisdiction => ({ value: jurisdiction.value, text: jurisdiction.text }))
   }
-  get provinceStateOptions () {
-    if (this.location === 'IN') {
-      return null
-    }
-    return this.$canJurisdictions
-  }
```
cameron-freshworks, line 448: Unused code

client/src/components/common/applicant-info-1.vue
```diff
@@ -539,18 +530,15 @@ export default class ApplicantInfo1 extends Vue {
...
   }
   queryAddress (id) {
     newReqModule.getAddressDetails(id)
   }
-  showNextTab () {
-    newReqModule.mutateSubmissionTabComponent('ApplicantInfo2')
-  }
-  showPreviousTab () {
-    newReqModule.mutateSubmissionTabComponent('NamesCapture')
-  }
```
cameron-freshworks, line 553: Unused code

severinbeauvais (approved):
Looks OK but you've really changed how each field is functioning, so a couple of us should test this thoroughly in Dev.

client/src/components/common/applicant-info-1.vue
severinbeauvais: If you were keen, you could also delete this from applicant-info-2 and applicant-info-3. :)

cameron-freshworks:
> Looks OK but you've really changed how each field is functioning, so a couple of us should test this thoroughly in Dev.

Definitely!
Essentially i removed the name fields from the blanket updating method where the validations get reset on any character input. This was causing it to bypass the rules validations until the field was clicked off.
Though i understand why that method does what it does in the first place, so it can allow the auto complete in the address fields to do its job.

client/src/components/common/applicant-info-1.vue
cameron-freshworks: oh right, will do!

PR #249, opened 2020-12-18
5964 quick search

kialj876:

*Issue #:* /bcgov/entity#5964

*Description of changes:*
- quick search that uses the same endpoint that examiners use in namex and shows list to user
- see ticket / invision link in it for details

### DIFF
--- a/client/public/config/configuration.json
+++ b/client/public/config/configuration.json
@@ -19,6 +19,9 @@
     "DEV_EXISTING_NR_TIMEOUT_MS": 0,
     "PAYMENT_COMPLETION_TIMER_NAME": "paymentCompletionTimer",
     "PAYMENT_COMPLETION_TIMEOUT_MS": 120000,
-    "DEV_PAYMENT_COMPLETION_TIMEOUT_MS": 900000
+    "DEV_PAYMENT_COMPLETION_TIMEOUT_MS": 900000,
+    "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
+    "QUICK_SEARCH_PUBLIC_ID": "",
+    "QUICK_SEARCH_PUBLIC_SECRET": ""
   }
 ]
--- a/client/src/components/new-request/analyze-pending.vue
+++ b/client/src/components/new-request/analyze-pending.vue
@@ -12,11 +12,11 @@
       </v-row>
       <v-row justify="center" class="mt-n7 mb-n3 py-0">
         <v-col cols="12" class="detailed-search-title">
-          <b>Detailed Search</b>
+          <b>Detailed Analysis</b>
         </v-col>
         <v-col cols="12" class="text-center text-body-2">
           We are attempting a detailed analysis of this name before you send it for review.
-          <p>This search will run for up to two minutes. Note: We might not obtain results for certain names.</p>
+          <p>This analysis will run for up to two minutes. Note: We might not obtain results for certain names.</p>
         </v-col>
       </v-row>
       <v-row justify="center">
--- a/client/src/components/new-request/analyze-pending.vue
+++ b/client/src/components/new-request/analyze-pending.vue
@@ -11,6 +11,9 @@
         <NameInput is-search-again="true"/>
       </v-row>
       <v-row justify="center" class="mt-n7 mb-n3 py-0">
+        <v-col cols="12" class="detailed-search-title">
+          <b>Detailed Search</b>
+        </v-col>
         <v-col cols="12" class="text-center text-body-2">
           We are attempting a detailed analysis of this name before you send it for review.
           <p>This search will run for up to two minutes. Note: We might not obtain results for certain names.</p>
@@ -78,4 +81,13 @@ export default class AnalyzePending extends Vue {
 
 </script>
 <style scoped lang="scss">
+@import '@/assets/scss/theme.scss';
+.detailed-search-title {
+  font-size: 1rem;
+  color: $gray9;
+  text-align: center;
+  padding: 0;
+  margin: 0;
+  padding-top: 1.5625rem;
+}
 </style>
--- a/client/src/components/new-request/quick-search-names.vue
+++ b/client/src/components/new-request/quick-search-names.vue
@@ -1,0 +1,42 @@
+<template>
+  <v-col id="quick-search-names">
+      <v-row class="quickSearchNames" v-for="name of namesList" :key="name.name">
+        <b v-if="name.type==='exact'">{{name.name}}</b>
+        <span v-else>{{name.name}}</span>
+      </v-row>
+  </v-col>
+</template>
+
+<script lang="ts">
+import MainContainer from '@/components/new-request/main-container.vue'
+import newReqModule from '@/store/new-request-module'
+import { Component, Vue, Watch } from 'vue-property-decorator'
+
+@Component({
+  components: { MainContainer }
+})
+export default class QuickSearchNames extends Vue {
+  get namesList () {
+    return newReqModule.quickSearchNames
+  }
+}
+
+</script>
+
+<style scoped lang="scss">
+@import '@/assets/scss/theme.scss';
+#quick-search-names {
+    height: 300px;
+    border-style: solid;
+    border-width: 1px;
+    border-radius: 5px;
+    display: block;
+    overflow: auto;
+}
+.quickSearchNames {
+    font-size: 16px;
+    color: $gray7;
+    padding-left: 5px;
+    padding-top: 15px;
+}
+</style>
--- a/client/src/components/new-request/quick-search-names.vue
+++ b/client/src/components/new-request/quick-search-names.vue
@@ -26,17 +26,17 @@ export default class QuickSearchNames extends Vue {
 <style scoped lang="scss">
 @import '@/assets/scss/theme.scss';
 #quick-search-names {
-    height: 300px;
+    height: 18.75rem;
     border-style: solid;
-    border-width: 1px;
-    border-radius: 5px;
+    border-width: 0.0625rem;
+    border-radius: 0.3125rem;
     display: block;
     overflow: auto;
 }
 .quickSearchNames {
-    font-size: 16px;
+    font-size: 1rem;
     color: $gray7;
-    padding-left: 5px;
-    padding-top: 15px;
+    padding-left: 0.3125rem;
+    padding-top: 0.9375rem;
 }
 </style>
--- a/client/src/components/new-request/quick-search-results.vue
+++ b/client/src/components/new-request/quick-search-results.vue
@@ -1,0 +1,135 @@
+<template>
+  <MainContainer id="quick-search-results-container">
+    <template v-slot:container-header>
+      <v-col cols="auto" class="h6 py-0 mt-1">
+        You are searching for a name for a
+        {{ entityText === ' BC Corporation' && location.text === ' BC' ? '' : ' ' + location.text }}
+        {{ entityText }}
+      </v-col>
+    </template>
+    <template v-slot:content>
+      <v-row no-gutters class="pt-3">
+        <NameInput/>
+      </v-row>
+      <v-row justify="center" class="ma-0 pa-0">
+        <v-col cols="12" class="quick-search-title">
+          <b>Quick Search Results</b>
+        </v-col>
+      </v-row>
+      <v-row v-if="namesList.length" justify="center" class="ma-0 pa-0">
+        <v-col cols="12" class="quick-search-info">
+          We found <b>{{namesList.length}}</b> names using these, or similar, words.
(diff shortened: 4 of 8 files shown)


client/public/config/configuration.json
```diff
@@ -19,6 +19,9 @@
     "DEV_EXISTING_NR_TIMEOUT_MS": 0,
     "PAYMENT_COMPLETION_TIMER_NAME": "paymentCompletionTimer",
     "PAYMENT_COMPLETION_TIMEOUT_MS": 120000,
-    "DEV_PAYMENT_COMPLETION_TIMEOUT_MS": 900000
+    "DEV_PAYMENT_COMPLETION_TIMEOUT_MS": 900000,
+    "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
+    "QUICK_SEARCH_PUBLIC_ID": "",
+    "QUICK_SEARCH_PUBLIC_SECRET": ""
```
severinbeauvais, line 25: Don't forget to add these to the config maps in Dev/Test/Prod. (You could probably do it now as it won't break anything.)

client/src/components/new-request/analyze-pending.vue
```diff
@@ -11,8 +11,11 @@
         <NameInput is-search-again="true"/>
       </v-row>
       <v-row justify="center" class="mt-n7 mb-n3 py-0">
+        <v-col cols="12" class="detailed-search-title">
+          <b>Detailed Search</b>
+        </v-col>
         <v-col cols="12" class="text-center text-body-2">
-          <b>We are attempting to find businesses using a similar name before you send this name for review.</b>
+          We are attempting a detailed analysis of this name before you send it for review.
           <p>This search will run for up to two minutes. Note: We might not obtain results for certain names.</p>
```
severinbeauvais, line 19: Fix this line according to uxpin also.

client/src/components/new-request/analyze-pending.vue
```diff
@@ -11,8 +11,11 @@
         <NameInput is-search-again="true"/>
       </v-row>
       <v-row justify="center" class="mt-n7 mb-n3 py-0">
+        <v-col cols="12" class="detailed-search-title">
+          <b>Detailed Search</b>
```
severinbeauvais, line 15: Invision has this as "Details Analysis".

client/src/components/new-request/analyze-pending.vue
```diff
@@ -83,4 +81,13 @@ export default class AnalyzePending extends Vue {
...
 <style scoped lang="scss">
+@import '@/assets/scss/theme.scss';
+.detailed-search-title {
+  font-size: 1rem;
+  color: $gray9;
+  text-align: center;
+  padding: 0;
+  margin: 0;
+  padding-top: 1.5625rem;
+}
```
severinbeauvais, line 92: We may have default styles (in base.scss) for some of these properties... but we can fix this later.

client/public/config/configuration.json
kialj876: doing it now

client/src/components/new-request/analyze-pending.vue
kialj876: okay will update

client/src/components/new-request/quick-search-names.vue
```diff
@@ -0,0 +1,42 @@
...
+    border-radius: 0.3125rem;
+    display: block;
+    overflow: auto;
+}
+.quickSearchNames {
+    font-size: 1rem;
+    color: $gray7;
+    padding-left: 0.3125rem;
+    padding-top: 0.9375rem;
+}
```
severinbeauvais, line 41: Generally, I prefer to see margins/padding in the HTML using Vuetify's helper classes. This makes it way easier to trouble-shoot and tweak layout.
https://vuetifyjs.com/en/styles/spacing/

Again, we can fix this later.

kialj876: okay sounds good, I'll keep this in mind for future PRs

client/src/components/new-request/quick-search-results.vue
```diff
@@ -0,0 +1,135 @@
...
+        You are searching for a name for a
+        {{ entityText === ' BC Corporation' && location.text === ' BC' ? '' : ' ' + location.text }}
+        {{ entityText }}
+      </v-col>
+    </template>
+    <template v-slot:content>
+      <v-row no-gutters class="pt-3">
+        <NameInput/>
+      </v-row>
+      <v-row justify="center" class="ma-0 pa-0">
```
severinbeauvais, line 14: FYI, "no-gutters" is basically equivalent to "ma-0 pa-0" for the row as well as the columns in it. It's not always the best thing to do but FYI.

client/src/components/new-request/quick-search-results.vue
```diff
@@ -0,0 +1,135 @@
...
+          We did not find any names using these exact or similar words within the same business category.
+        </v-col>
+      </v-row>
+      <v-row justify="center">
+        <v-col cols="auto" class="pb-0 pr-1">
+          <v-btn id="search-again-button" :disabled="nameChanged" @click="searchAgain">Search Again</v-btn>
+        </v-col>
+        <v-col cols="auto" class="pb-0 pr-1">
+          <v-btn class="continue-search-button" color="primary" outlined @click="detailedSearch">
+            Continue to Detailed Analysis
```
severinbeauvais, line 42: Generally it's better if the text is on the same line as the > and < of the button element, or within a span on its own line, to facilitate testing (ie, you can select the exact button text without a newline before or after it).
We can fix this later.

client/src/components/new-request/quick-search-results.vue
```diff
@@ -0,0 +1,135 @@
...
+  get entity_type_cd () {
+    return newReqModule.entity_type_cd
+  }
+  get entityText () {
+    return newReqModule.entityTextFromValue
+  }
+  get location () {
+    let value = newReqModule.location
+    let options = newReqModule.locationOptions
+    return options.find((opt: any) => opt.value === value)
```
severinbeauvais, line 81: Is it OK if this returns undefined or null?

severinbeauvais: I just checked. If the value is not found then the HTML will throw an error. Either return a fallback object here or handle it in the HTML please.

client/src/store/new-request-module.ts
```diff
@@ -135,6 +135,7 @@ export class NewRequestModule extends VuexModule {
   addressSuggestions: object | null = null
   allowAutoApprove: boolean = false
   analysisJSON: AnalysisJSONI | null = null
+  analyzePending: Boolean = false
```
severinbeauvais, line 138: Don't use "Boolean".
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object

severinbeauvais: In fact, TS already knows this is a boolean, and in some environments the linter will throw an error about "redundant type", so I would just remove the type here.
(Rule of thumb: if you hover over the property and the proper type is displayed, it's good.)

client/src/components/new-request/quick-search-results.vue
kialj876: okay sounds good, I split it up because the line was too long

client/src/components/new-request/quick-search-results.vue
kialj876: hmm okay. This is the same code that we have in the other components, but I will see how they are handfling it

client/src/store/new-request-module.ts
severinbeauvais: (I'm assuming you have some VS Code packages installed. Vetur is probably the main one :)

kialj876: haha sorry will update

client/src/store/new-request-module.ts
```diff
@@ -2504,8 +2518,103 @@ export class NewRequestModule extends VuexModule {
...
+      synonymResp.data.exactNames = exactNames
+      const synonymNames = await this.parseSynonymNames(synonymResp.data)
+      this.mutateQuickSearchNames(exactNames.concat(synonymNames))
+      // check if they skipped
+      if (this.quickSearch) {
+        this.mutateDisplayedComponent('QuickSearchResults')
+      }
+    } catch (error) {
+      // handle silently and move on to detailed search
+      console.error('quick search failed: ', error) // eslint-disable-line no-console
```
severinbeauvais, line 2575: This is not silent: it will get picked up by Sentry. Maybe use console.log or console.warn instead.

client/src/components/new-request/quick-search-results.vue
kialj876: hey sev, this is an existing error in the analyze components used in the top section of each one so to fix it properly it should be in its own comp and fixed there, then imported into each one. Can we say tech debt ticket for this?

client/src/components/new-request/quick-search-results.vue
kialj876: I was wondering what that was haha

client/src/store/new-request-module.ts
```diff
@@ -2504,8 +2518,103 @@ export class NewRequestModule extends VuexModule {
...
+      this.mutateQuickSearch(false)
+      this.startAnalyzeName()
+    }
+  }
+  @Action
+  async startQuickSearch () {
+    if (this.name) {
+      let name = this.name
+      let exactMatchName = name.replace(' \/', '\/')
```
severinbeauvais, line 2584: You can use const for these 2 variables since you don't change them (you are changing their content, not their value).

client/src/store/new-request-module.ts
kialj876: I want sentry to pick this up, but I don't want the user to see it

kialj876: that's what I meant by silent. I don't think we should ever have a truly silent error

severinbeauvais:
In Openshift, you should store the keys in a secret.

client/src/store/new-request-module.ts
kialj876: sure thing will update

severinbeauvais:
Should this feature be Feature Flagged so we don't enable it right away? Do we want this feature to go in along with the other bugfixes? (Or at least don't merge and deploy this to Test/Prod until it's been well tested.)

kialj876:
> In Openshift, you should store the keys in a secret.

I don't think it matters because they are public anyways. Someone could take these out and use them if they knew what they were doing, but the token is view only and you can't do anything with it except get the solr search

client/src/store/new-request-module.ts
kialj876: not for exact match but I changed it to const for name and synonymMatch further down

kialj876:
> Should this feature be Feature Flagged so we don't enable it right away? Do we want this feature to go in along with the other bugfixes? (Or at least don't merge and deploy this to Test/Prod until it's been well tested.)

I think Kaine wants this in asap and it if it fails it will skip to the detailed analysis anyways

vysakh-menon-aot (approved):

client/src/components/new-request/quick-search-results.vue
severinbeauvais: Yeah let's do that for all components that can fail this way.

client/src/store/new-request-module.ts
severinbeauvais: In that case, I suggest a more detailed comment, and perhaps add the method name to the error like the other ones in this module do.

severinbeauvais:
> > In Openshift, you should store the keys in a secret.
> 
> I don't think it matters because they are public anyways. Someone could take these out and use them if they knew what they were doing, but the token is view only and you can't do anything with it except get the solr search

OK.

severinbeauvais:
> > Should this feature be Feature Flagged so we don't enable it right away? Do we want this feature to go in along with the other bugfixes? (Or at least don't merge and deploy this to Test/Prod until it's been well tested.)
> 
> I think Kaine wants this in asap and it if it fails it will skip to the detailed analysis anyways

OK.

client/src/store/new-request-module.ts
kialj876: ahhh I only saw your first comment -- github just updated me with the other 2. I set it to boolean (I think this is okay because there are other values set to 'boolean' as well so it is consistent at least)

severinbeauvais: Yeah, if we ever have to change this it will be a global effort. It's fine for now.

client/src/store/new-request-module.ts
kialj876: done

severinbeauvais: It's perfect.

PR #250, opened 2020-12-21
Disabled Error modal for payment cancellation

cameron-freshworks:
*Issue #:* /bcgov/entity#5885

*Description of changes:*

* Removed Error modal request upon a payment cancellation. 

*Notes:* 

* This approach could potentially change when the deeper solution to PAYMENTS is complete.

### DIFF
--- a/client/src/components/payment/payment-complete-modal.vue
+++ b/client/src/components/payment/payment-complete-modal.vue
@@ -82,11 +82,6 @@ export default class PaymentCompleteModal extends Mixins(NameRequestMixin, Payme
         // Then complete the payment
         await this.completePayment(nrId, sessionPaymentId, sessionPaymentAction)
       } else {
-        errorModule.setAppError({
-          id: 'payment-error',
-          error: 'Your payment could not be completed. Please try again at a later time.'
-        } as ErrorI)
-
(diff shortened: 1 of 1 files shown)


client/src/components/payment/payment-complete-modal.vue
```diff
@@ -82,11 +82,6 @@ export default class PaymentCompleteModal extends Mixins(NameRequestMixin, Payme
         // Then complete the payment
         await this.completePayment(nrId, sessionPaymentId, sessionPaymentAction)
       } else {
-        errorModule.setAppError({
```
severinbeauvais, line 85: Something to keep in mind here: it may be possible to look at `sbcPaymentStatus` and `paymentStatus` (as on line 81 above) and determine that, yes, the user returned from PayBC but their payment was cancelled or failed. And in this case, we could give the user a useful dialog.

severinbeauvais (approved):

client/src/components/payment/payment-complete-modal.vue
cameron-freshworks: Definitely, i agree 100%.
Question, should we do this now or after Payment updates are in?

vysakh-menon-aot (approved):

PR #255, opened 2020-12-22
Verbiage, Tooltip updates & MRAs bug Fix

cameron-freshworks:
*Issue #:* /bcgov/entity#6027 /bcgov/entity#6028 /bcgov/entity#5896

*Description of changes:*

* Implemented loading indicator on 'Send to analysis' btn.
* Updated BETA tooltip content.
* Updated verbiage in 'Sent to examination' component.
* Added checks for Multiple names in MRAs profile
* Updated Verbiage in Request Actions NEW and MOVE

### DIFF
--- a/client/src/components/new-request/quick-search-results.vue
+++ b/client/src/components/new-request/quick-search-results.vue
@@ -97,9 +97,9 @@ export default class QuickSearchResults extends Vue {
   mounted () {
     this.originalName = this.searchName
   }
-  searchAgain () {
+  async searchAgain () {
     newReqModule.mutateQuickSearch(true)
-    newReqModule.startAnalyzeName()
+    await newReqModule.startAnalyzeName()
   }
   async detailedSearch () {
     this.isLoadingAnalysis = true
--- a/client/src/components/new-request/quick-search-results.vue
+++ b/client/src/components/new-request/quick-search-results.vue
@@ -38,7 +38,11 @@
           <v-btn id="search-again-button" :disabled="nameChanged" @click="searchAgain">Search Again</v-btn>
         </v-col>
         <v-col cols="auto" class="pb-0 pr-1">
-          <v-btn class="continue-search-button" color="primary" outlined @click="detailedSearch">
+          <v-btn class="continue-search-button"
+                 color="primary"
+                 outlined
+                 @click="detailedSearch"
+                 :loading="isLoadingAnalysis">
             Continue to Detailed Analysis
           </v-btn>
         </v-col>
@@ -68,6 +72,7 @@ export default class QuickSearchResults extends Vue {
   }
   contents: string = ''
   originalName: string = ''
+  private isLoadingAnalysis: boolean = false
 
   get entity_type_cd () {
     return newReqModule.entity_type_cd
@@ -96,9 +101,11 @@ export default class QuickSearchResults extends Vue {
     newReqModule.mutateQuickSearch(true)
     newReqModule.startAnalyzeName()
   }
-  detailedSearch () {
+  async detailedSearch () {
+    this.isLoadingAnalysis = true
     newReqModule.mutateQuickSearch(false)
-    newReqModule.startAnalyzeName()
+    await newReqModule.startAnalyzeName()
+    this.isLoadingAnalysis = false
   }
 }
 
--- a/client/src/components/new-request/send-to-examination.vue
+++ b/client/src/components/new-request/send-to-examination.vue
@@ -13,15 +13,17 @@
         <NameInput :is-read-only="true" class="mb-n2"/>
       </v-row>
 
-      <v-row no-gutters justify="center">
+      <v-row no-gutters justify="center" class="text-center">
         <v-col cols="auto" class="h5">
           <v-icon class="action-icon" :class="actionConfig.class">{{ actionConfig.icon }}</v-icon>
-          &nbsp;
           <span :class="actionConfig.class">{{ actionConfig.text }}</span>
+        </v-col>
+        <v-col cols="12">
+          <p :class="actionConfig.class">{{ actionConfig.subText }}</p>
         </v-col>
         <v-col cols="12" class="py-2">
           <p class="text-center copy-normal">
-            This name cannot be analyzed here and must be sent to staff for review.
+            Submit your name to staff for review.
           </p>
         </v-col>
       </v-row>
@@ -49,9 +51,10 @@ import NameInput from '@/components/new-request/name-input.vue'
 })
 export default class SendToExamination extends Vue {
   private actionConfig = {
-    class: 'rejected',
-    icon: 'mdi-star-circle',
-    text: 'Further Review Required',
+    class: 'approved',
+    icon: 'mdi-check-circle',
+    text: 'Name Ready for Review',
+    subText: '(Detailed Analysis Coming Soon)',
     showNextLines: true
   }
   get entityText () {
--- a/client/src/components/new-request/send-to-examination.vue
+++ b/client/src/components/new-request/send-to-examination.vue
@@ -18,12 +18,12 @@
           <v-icon class="action-icon" :class="actionConfig.class">{{ actionConfig.icon }}</v-icon>
           <span :class="actionConfig.class">{{ actionConfig.text }}</span>
         </v-col>
-        <v-col cols="12" class="pb-2">
+        <v-col cols="12">
           <p :class="actionConfig.class">{{ actionConfig.subText }}</p>
         </v-col>
       </v-row>
 
-      <v-row class="text-center justify-center mt-n1 mb-n2">
+      <v-row class="text-center justify-center mb-n2">
         <ReserveSubmit />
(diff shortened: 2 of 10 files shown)


client/src/components/new-request/quick-search-results.vue
```diff
@@ -96,9 +101,11 @@ export default class QuickSearchResults extends Vue {
     newReqModule.mutateQuickSearch(true)
     newReqModule.startAnalyzeName()
   }
-  detailedSearch () {
+  async detailedSearch () {
+    this.isLoadingAnalysis = true
     newReqModule.mutateQuickSearch(false)
-    newReqModule.startAnalyzeName()
+    await newReqModule.startAnalyzeName()
```
severinbeauvais, line 107: There are many other places in the code where startAnalyzeName is called. Do you think it's worth changing those to awaits also?

client/src/components/new-request/send-to-examination.vue
```diff
@@ -13,16 +13,13 @@
...
         <v-col cols="auto" class="h5">
           <v-icon class="action-icon" :class="actionConfig.class">{{ actionConfig.icon }}</v-icon>
-          &nbsp;
           <span :class="actionConfig.class">{{ actionConfig.text }}</span>
         </v-col>
-        <v-col cols="12" class="py-2">
-          <p class="text-center copy-normal">
-            This name cannot be analyzed here and must be sent to staff for review.
-          </p>
+        <v-col cols="12" class="pb-2">
```
severinbeauvais, line 21: Would it be better to have pb-0 on the p element below?
(Negative margins/padding are always a red flag.)

client/src/store/new-request-module.ts
```diff
@@ -2646,7 +2646,10 @@ export class NewRequestModule extends VuexModule {
       if (!this.noCorpNum) {
         const profile = await this.fetchMRASProfile()
         if (profile) {
-          name = sanitizeName(profile?.LegalEntity?.names?.legalName)
+          const hasMultipleNames = profile?.LegalEntity?.names && profile?.LegalEntity?.names.constructor === Array
+          name = hasMultipleNames
+            ? sanitizeName(profile?.LegalEntity?.names[0]?.legalName)
+            : sanitizeName(profile?.LegalEntity?.names?.legalName)
```
severinbeauvais, line 2652: Please call this out as a required QA test in https://github.com/bcgov/entity/issues/6028.

client/src/views/landing.vue
```diff
@@ -142,4 +156,7 @@ export default class Landing extends Vue {
 .v-tooltip__content {
   pointer-events: auto !important;
 }
+.tooltip-extra-wide {
```
severinbeauvais, line 159: LOL

client/src/views/landing.vue
```diff
@@ -22,12 +22,26 @@
...
-                official name requests and can be used in the province of British Columbia.</p>
-              <p class="mb-0">As part of a new
-                <a class="white--text" :href="agileUrl" target="_blank">agile software development process</a>,
+              <p>The Name Request website is available as a Beta version. Name requests obtained through the Name
+                Request Beta website are official name requests and can be used in the province of British Columbia.</p>
+              <p class="mb-0">The Name Request Beta website contains all of the functionality of the previous Name
+                Requests Online site, as well as new functionality intended to improve the experience of starting and
+                managing businesses in B.C. Some features of the site may not function as intended, and if you encounter
+                this please contact us:</p>
+              <span class="tooltip-bullet mt-3">
```
severinbeauvais, line 31: No need to do this. Use ~components/common/contact-info.vue instead (with `direction="col"`).

client/src/views/landing.vue
severinbeauvais: Also check the minimum width for this tooltip (ie, narrow your window and verify that content wraps correctly).

client/src/views/landing.vue
```diff
@@ -22,12 +22,26 @@
...
+                <a class="white--text" href="tel:+1-877-370-1033" target="_blank">Toll Free: 1-877-370-1033</a>
+              </span>
+              <span class="tooltip-bullet">
+                <a class="white--text" href="tel:+1-250-370-1033" target="_blank">Victoria Office: 250-370-1033</a>
+              </span>
+              <span class="tooltip-bullet mb-3">
+                <a class="white--text" href="mailto:BCRegistries@gov.bc.ca" target="_blank">BCRegistries@gov.bc.ca</a>
+              </span>
+              <p class="mb-0">As part of a
+                <a class="white--text" :href="agileUrl" target="_blank">agile software development process</a>
```
severinbeauvais, line 41: Missing "modern" here.
https://projects.invisionapp.com/share/AWZPCO1H7ZD#/screens/440284868_Beta_Tag_Update

client/src/components/new-request/quick-search-results.vue
cameron-freshworks: I'm not sure.. 
I don't see a downside to changing, but i'm not sure it will have any unexpected implications. This would require further testing on each call i think.

client/src/components/new-request/send-to-examination.vue
cameron-freshworks: Ok, will check it out!

client/src/store/new-request-module.ts
cameron-freshworks: For sure, will do!

client/src/views/landing.vue
cameron-freshworks: ah interesting. Will see how this works within a tooltip.

client/src/views/landing.vue
cameron-freshworks: Good catch!

client/src/views/landing.vue
severinbeauvais: FWIW, I think the contact info component, with icons instead of bullets, is fine. (consistency)

vysakh-menon-aot (approved):

client/src/components/new-request/quick-search-results.vue
cameron-freshworks: Updated!

client/src/components/new-request/send-to-examination.vue
cameron-freshworks: Fixed

client/src/store/new-request-module.ts
cameron-freshworks: Done

client/src/views/landing.vue
cameron-freshworks: For sure, though the view as a whole doesn't quite compress that small.

client/src/views/landing.vue
cameron-freshworks: To use the contact info component here will require changes to the component itself, which will also affect it's other implementation.

severinbeauvais: Oh well. Hopefully we will remove the Beta label at some point :)

client/src/views/landing.vue
```diff
@@ -52,6 +66,7 @@ import { Vue, Component, Prop } from 'vue-property-decorator'
 import AnalyzeCharacters from '@/components/new-request/analyze-characters.vue'
 import AnalyzePending from '@/components/new-request/analyze-pending.vue'
 import AnalyzeResults from '@/components/new-request/analyze-results.vue'
+import ContactInfo from '@/components/common/contact-info.vue'
```
severinbeauvais, line 69: Obsolete code.

cameron-freshworks: Ahhhhh

client/src/components/new-request/send-to-examination.vue
```diff
@@ -13,20 +13,17 @@
...
-          <p class="text-center copy-normal">
-            This name cannot be analyzed here and must be sent to staff for review.
-          </p>
+        <v-col cols="12">
+          <p :class="actionConfig.class">{{ actionConfig.subText }}</p>
         </v-col>
       </v-row>
 
-      <v-row class="text-center justify-center mt-n1 mb-n2">
+      <v-row class="text-center justify-center mb-n2">
```
severinbeauvais, line 26: For next time...

To simplify layout and spacing, in the dev window, Elements tab, I hover on various elements and watch that they don't have any extra whitespace (esp below them).

Basically my rule is that all elements should be "tight" and only have explicit whitespace added (eg, above it for spacing to the previous element). It's almost never necessary to have whitespace below an element.

In this case, I suspect the element above this has bottom whitespace, which is why you have to rein some back in with this margin style. So you have updated THIS row to fix a problem in the previous row.

severinbeauvais (approved):

client/src/views/landing.vue
cameron-freshworks: Definitely. 
That being said, there is similar listed contact info in the lower section as well and an implementation of the contact info component in some grey-boxes etc
This is definitely a spot for some refactoring when we can.

PR #256, opened 2020-12-22
Cancel the payment modal timer

severinbeauvais:
*Issue #:* /bcgov/entity#6025

*Description of changes:*

This is a necessary first part of the subject ticket. My plan is to get this into Prod and then see what errors remain.

- updated app version to 0.2.5
- stop payment completion timer when modal is hidden
- stop payment completion timer when modal is confirmed
- commented out unused method, etc (may be reinstated later)
- commented out timer debugging (may be useful later)
- fixed some resolve warnings

### DIFF
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -46,6 +46,7 @@
     <PaymentModal
       :onActivate="onPaymentModalActivated"
       :onCancel="onPaymentCancelled"
+      :stopTimer="stopPaymentCompletionTimer"
     />
     <UpgradeModal />
     <ReapplyModal />
@@ -162,7 +163,7 @@ export default class App extends Mixins(SessionTimerMixin) {
     // Avoided redundant navigation to current location: "/"
     this.$router.replace('/').catch(() => {})
 
-    // Display the tabs
+    // Clear data and display the Tabs page
     await newRequestModule.resetAnalyzeName()
     await newRequestModule.mutateName('')
     await newRequestModule.mutateDisplayedComponent('Tabs')
@@ -171,6 +172,7 @@ export default class App extends Mixins(SessionTimerMixin) {
   async onTimerModalExpired () {
     const { nrId } = newRequestModule
     const componentName = newRequestModule.displayedComponent
+
     if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
       // Cancel the NR using the rollback endpoint if we were processing a NEW NR
       await newRequestModule.rollbackNameRequest({ nrId, action: RollbackActions.CANCEL })
@@ -184,6 +186,7 @@ export default class App extends Mixins(SessionTimerMixin) {
   async onTimerModalSessionExtended () {
     const { nrId } = newRequestModule
     const componentName = newRequestModule.displayedComponent
+
     if (['SubmissionTabs'].indexOf(componentName) > -1 && this.$NR_COMPLETION_TIMEOUT_MS > 0) {
       await this.startNewNrTimer()
     } else if (nrId && ['ExistingRequestEdit'].indexOf(componentName) > -1 && this.$EXISTING_NR_TIMEOUT_MS > 0) {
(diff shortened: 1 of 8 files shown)


client/src/App.vue
```diff
@@ -194,12 +197,11 @@ export default class App extends Mixins(SessionTimerMixin) {
   async onPaymentModalActivated () {
     const { nrId } = newRequestModule
     const componentName = newRequestModule.displayedComponent
+
     // Only do this for New NRs!!!
     if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
-      if (this.$NR_COMPLETION_TIMEOUT_MS > 0) {
```
severinbeauvais, line 199: Don't need to check this since "stopTimer" can be called when the timer doesn't exist.

vysakh-menon-aot (approved):

client/src/components/payment/payment-modal.vue
```diff
@@ -103,6 +107,11 @@ export default class PaymentModal extends Mixins(
 
   /** Clears store property to hide this modal. */
   async hideModal () {
+    // stop timer because we don't need it any more
+    const { stopTimer } = this.$props
+    if (typeof stopTimer === 'function') {
+      stopTimer()
+    }
```
cameron-freshworks, line 114: I still don't quite follow these styles of conditions. Why do we check it in this manner if it's defined above as a function?

client/src/modules/vx-timer/store/actions.ts
```diff
@@ -5,9 +5,9 @@ export const createAndStartTimer = ({ commit }, opts: TimerOptions) => {
   commit(types.CREATE_AND_START_TIMER, opts)
 }
 
-export const createPromiseTimer = ({ commit }, opts: TimerOptions) => {
-  commit(types.CREATE_PROMISE_TIMER, opts)
-}
+// export const createPromiseTimer = ({ commit }, opts: TimerOptions) => {
+//   commit(types.CREATE_PROMISE_TIMER, opts)
+// }
```
cameron-freshworks, line 10: For future use?

client/src/modules/vx-timer/store/mutations.ts
```diff
@@ -57,13 +58,13 @@ export default {
     const { timers } = state
     if (!id) throw new Error(`[Error] Timer ID was not provided`)
     if (!timers.hasOwnProperty(id)) throw new Error(`[Error] Timer with ID [${id}] does not exist`)
+    // console.info('Refreshing timer =', id) // eslint-disable-line no-console
```
cameron-freshworks, line 61: Intentional comment?

client/src/modules/vx-timer/store/mutations.ts
```diff
@@ -73,7 +74,9 @@ export default {
     if (!id) throw new Error(`[Error] Timer ID was not provided`)
     const timer = timers[id]
     if (timer) {
+      // console.info('Stopping timer =', id) // eslint-disable-line no-console
```
cameron-freshworks, line 77: Same as above

cameron-freshworks (approved):
Looks good, just a few questions!

client/src/components/payment/payment-modal.vue
severinbeauvais: In case it's not passed in as a prop, I think.
I haven't verified whether this working properly (existing code).

cameron-freshworks: ok cool, thanks!

client/src/modules/vx-timer/store/actions.ts
severinbeauvais: Yea, maybe. It's not used now but someone wrote it for a reason.

client/src/modules/vx-timer/store/mutations.ts
severinbeauvais: Yes. I used these console.info()s to verify the timer behaviours... But atm we're only used the 1 timer (and I think it's OK now). When we enable the other timers, we may need to verify their behaviours too...

PR #259, opened 2020-12-24
5768 MOVE(continue) assumed and request fixes

cameron-freshworks:
*Issue #:* /bcgov/entity#5768

*Description of changes:*

* This is a fix for MVE requests. Currently they were not working as XPRO, so we have now mapped (CA AND IN) back to BC.. so we can see the BC entities and make a BC request. 

* This also fixes the assumed named flow for MVE requests, it will now follow the normal BC flow.

*Notes:*
* Personally if we map back to BC we should have made the location selection BC and updated the verbiage, but this is what was asked for..

* This can create an issue because we have an INTERNATIONAL Selection that essentially does nothing.. and a Canada selection that maps back to BC.. 

* This is a quick(NOT QUICK) fix that will allow the request to complete but designers and BA's need to look at this issue and come up with a long term solution as i don't like this much..

### DIFF
--- a/client/src/components/new-request/search.vue
+++ b/client/src/components/new-request/search.vue
@@ -348,6 +348,7 @@ export default class NewSearch extends Vue {
     newReqModule.mutateIsPersonsName(value)
   }
   get location () {
+    if (['MVE'].includes(this.request_action_cd)) newReqModule.mutateLocation('BC')
     return newReqModule.location
   }
   set location (location: LocationT) {
@@ -384,9 +385,6 @@ export default class NewSearch extends Vue {
     const request = this.requestActions.find(request => request.value === value)
     if (request.value !== 'NEW') {
       newReqModule.mutateExtendedRequestType(request)
-    }
-    if (request.value === 'MVE') {
-      newReqModule.mutateLocation('CA')
     }
     newReqModule.mutateRequestAction(value)
   }
--- a/client/src/store/new-request-module.ts
+++ b/client/src/store/new-request-module.ts
@@ -245,6 +245,13 @@ export class NewRequestModule extends VuexModule {
         `Reported separately as Corporate tax`,
         `Has name protection in BC`
       ],
+      mveBlurbs: [
+        `A company that may have one or more people who own shares with some personal responsibility for
+        debt and liabilities.`,
+        'Has many of the same rights of an individual',
+        'Reported separately as Corporate tax',
+        'Has name protection in BC'
+      ],
       rehBlurbs: [
         'Restore a limited company that is no longer active with BC Registries.'
       ],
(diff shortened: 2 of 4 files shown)


vysakh-menon-aot (approved):

client/src/components/new-request/search.vue
```diff
@@ -343,6 +343,7 @@ export default class NewSearch extends Vue {
     newReqModule.mutateIsPersonsName(value)
   }
   get location () {
+    if (['MVE'].includes(this.request_action_cd)) newReqModule.mutateLocation('BC')
```
severinbeauvais, line 346: Could you please add code comments explaining things here?

client/src/store/new-request-module.ts
```diff
@@ -245,6 +245,13 @@ export class NewRequestModule extends VuexModule {
         `Reported separately as Corporate tax`,
         `Has name protection in BC`
       ],
+      mveBlurbs: [
+        `A company that may have one or more people who own shares with some personal responsibility for
+        debt and liabilities.`,
```
severinbeauvais, line 250: I think "debt" should be plural (as on line 281).

client/src/store/new-request-module.ts
```diff
@@ -376,6 +403,12 @@ export class NewRequestModule extends VuexModule {
         'Reported as Corporate tax',
         'Has name protection in BC'
       ],
+      mveBlurbs: [
+        'A type of corporations which has a benefit to  the community.  They are intended to bridge the gap between ' +
```
severinbeauvais, line 407: "corporations" should be singular?

Also, should be "xxx, which yyy" or "xxx that yyy", if possible. (Here, I prefer the "that" form.)

Second sentence refers to "they", which is plural. Can you use "it" to solve this?

client/src/store/new-request-module.ts
```diff
@@ -399,6 +432,14 @@ export class NewRequestModule extends VuexModule {
         'Has name protection in BC',
         'Must use Societies Online to register a name and incorporate'
       ],
+      mveBlurbs: [
+        `A non-profit organization that is also known as a non-share corporation.`,
+        'Any funds or profits must be used only for social or community benefit',
+        'When incorporated has independent legal status separate from its members',
```
severinbeauvais, line 438: Comma after "incorporated".

severinbeauvais (approved):
Looks ok. Just a few comments on existing grammar :)

client/src/components/new-request/search.vue
cameron-freshworks: Definitely!

client/src/store/new-request-module.ts
cameron-freshworks: for sure!

client/src/store/new-request-module.ts
cameron-freshworks: lol for sure :)

client/src/store/new-request-module.ts
cameron-freshworks: Will add!

PR #279, opened 2021-01-08
removed disable from continue buttons submission flow screens and trigger validation

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5858

*Description of changes:*

### DIFF
--- a/client/src/components/common/applicant-info-1.vue
+++ b/client/src/components/common/applicant-info-1.vue
@@ -321,7 +321,7 @@
                 class="mt-2 pt-0"
               />
             </v-col>
-            <ApplicantInfoNav :isValid="isValid" />
+            <ApplicantInfoNav @nextAction="nextAction" />
           </v-row>
         </v-col>
       </v-row>
@@ -330,10 +330,11 @@
 </template>
 
 <script lang="ts">
-import newReqModule from '@/store/new-request-module'
 import { Component, Vue, Watch } from 'vue-property-decorator'
+import newReqModule from '@/store/new-request-module'
 import ApplicantInfoNav from '@/components/common/applicant-info-nav.vue'
 import { Location } from '@/enums'
+import NameRequestMixin from '@/components/mixins/name-request-mixin'
 
 const _debounce = require('lodash/debounce')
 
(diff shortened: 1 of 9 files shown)


client/src/components/common/applicant-info-1.vue
```diff
@@ -321,7 +321,7 @@
                 class="mt-2 pt-0"
               />
             </v-col>
-            <ApplicantInfoNav :isValid="isValid" />
+            <ApplicantInfoNav @nextAction="nextAction" />
```
severinbeauvais, line 324: If possible add () to method (to make it clear to the reader that it's a method).

client/src/components/common/applicant-info-nav.vue
```diff
@@ -1,43 +1,37 @@
 <template>
   <v-col cols="5" class="text-right py-0">
     <v-btn x-large
-           :id="`submit-back-btn-${isValid}`"
+           :id="`submit-back-btn-true`"
```
severinbeauvais, line 4: I recommend you just remove "-true" here and elsewhere (and update styles/queries accordingly).
Otherwise this is a leftover artifact of a more complicated implementation. Simple is better! 😁

client/src/components/mixins/name-request-mixin.ts
```diff
@@ -118,4 +121,76 @@ export default class NameRequestMixin extends Vue {
...
+  async submit () {
+    // FUTURE: fix error handling in case of newReqModule (app or api) error (#5899)
+    const { nrId } = this
+    if (this.editMode) {
+      // stop timer first so it doesn't expire during network requests
+      timerModule.stopTimer({ id: this.$EXISTING_NR_TIMER_NAME })
+      if (await newRequestModule.patchNameRequests()) {
+        if (await newRequestModule.checkinNameRequest()) {
+          newRequestModule.mutateDisplayedComponent('Success')
+          await sleep(1000)
```
severinbeauvais, line 165: Please add comment why delay for 1 second here :)

severinbeauvais (approved):
Nice work! I love your solution of moving some of the code to the mixin. It really makes the other components much simpler!

client/src/components/common/applicant-info-nav.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <v-col cols="5" class="text-right py-0">
     <v-btn x-large
-           :id="`submit-back-btn-true`"
+           :id="`submit-back-btn`"
```
severinbeauvais, line 4: You don't need the back-ticks any more, just `:id="submit-back-btn"`.

severinbeauvais (approved):
Ah, even cleaner! 👍

PR #282, opened 2021-01-13
6086 drop box tooltip updates

kialj876:

*Issue #:* /bcgov/entity#6086

*Description of changes:*
- see ticket

### DIFF
--- a/client/src/components/new-request/search.vue
+++ b/client/src/components/new-request/search.vue
@@ -23,7 +23,6 @@
                         :hide-details="!errors.includes('request_action_cd')"
                         :items="requestActions"
                         @change="clearErrors()"
-                        class="select-drop-down"
                         label="Select an Action"
                         filled
                         v-model="request_action_cd">
@@ -57,7 +56,6 @@
                         :items="locationOptions"
                         :disabled="!request_action_cd || jurisdictionDisabled"
                         @change="clearErrors()"
-                        class="select-drop-down"
                         filled
                         label="Select a Jurisdiction"
                         v-model="location">
@@ -95,7 +93,6 @@
                         :label="isConversion ? 'Select an Alteration Type' : 'Select a Business Type'"
                         :disabled="!request_action_cd || !location"
                         @change="clearErrors()"
-                        class="select-drop-down"
                         filled
                         v-model="entity_type_cd">
                 <template slot="item" slot-scope="data">
@@ -135,7 +132,6 @@
                   :items="jurisdictionOptions"
                   label="Select business's home jurisdiction"
                   @change="clearErrors()"
-                  class="select-drop-down"
                   filled
                   v-model="jurisdiction">
           <template slot="item" slot-scope="data">
@@ -447,9 +443,6 @@ export default class NewSearch extends Vue {
 }
 .list-item:hover {
   color: $app-blue;
-}
-.select-drop-down {
-  padding-top: 0 !important;
 }
 #search-name-btn {
   min-height: 45px !important;
--- a/client/src/components/new-request/search.vue
+++ b/client/src/components/new-request/search.vue
@@ -10,7 +10,7 @@
     </v-row>
 
     <v-row class="mt-4" no-gutters>
-      <v-col cols="5">
+      <v-col cols="4">
         <!--request_action_cd-->
         <v-tooltip top
                    id="search-type-options-select"
@@ -19,16 +19,20 @@
                    :disabled="request_action_cd !== 'CNV'">
           <template v-slot:activator="scope">
             <div v-on="scope.on">
-              <v-select :error-messages="errors.includes('request_action_cd') ? 'Please select a type' : ''"
+              <v-select :error-messages="errors.includes('request_action_cd') ? 'Please select an action' : ''"
                         :hide-details="!errors.includes('request_action_cd')"
                         :items="requestActions"
                         @change="clearErrors()"
+                        class="select-drop-down"
+                        label="Select an Action"
                         filled
                         v-model="request_action_cd">
                 <template slot="item" slot-scope="data">
                   <v-tooltip :disabled="!data.item.blurbs" right transition="fade-transition">
                     <template v-slot:activator="scope">
-                      <span v-on="scope.on" class="list-item">{{ data.item.text }}</span>
+                      <span v-on="scope.on" class="list-item">
+                        {{ data.item.text.replace('Relocate into', 'Relocate into BC') }}
+                      </span>
                     </template>
                     <span>{{ data.item.blurbs }}</span>
                   </v-tooltip>
@@ -36,26 +40,29 @@
               </v-select>
             </div>
           </template>
-          <span>{{requestText}}</span>
+          <span>{{ requestText }}</span>
         </v-tooltip>
       </v-col>
-      <v-col cols="2" class="px-3">
+      <v-col cols="4" class="px-3">
         <!--location-->
         <v-tooltip id="location-options-select"
                    top
                    content-class="top-tooltip"
                    transition="fade-transition"
-                   :disabled="location === 'BC'">
+                   :disabled="!location || location === 'BC'">
           <template v-slot:activator="scope">
             <div v-on="scope.on">
               <v-select :error-messages="errors.includes('location') ? 'Please select a location' : ''"
                         :hide-details="!errors.includes('location')"
                         :items="locationOptions"
+                        :disabled="!request_action_cd || jurisdictionDisabled"
                         @change="clearErrors()"
+                        class="select-drop-down"
                         filled
+                        label="Select a Jurisdiction"
                         v-model="location">
                 <template slot="item" slot-scope="data">
-                  <v-tooltip :disabled="!data.item.blurbs" right transition="fade-transition">
+                  <v-tooltip :disabled="!request_action_cd || !data.item.blurbs" right transition="fade-transition">
                     <template v-slot:activator="scope">
                       <span v-on="scope.on" class="list-item">{{ data.item.text }}</span>
                     </template>
@@ -73,7 +80,7 @@
           <span>{{ locationText }}</span>
         </v-tooltip>
       </v-col>
-      <v-col cols="5">
+      <v-col cols="4">
         <!--entityConversionType-->
         <v-tooltip id="entity-type-options-select"
                    top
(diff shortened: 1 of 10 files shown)


client/src/components/new-request/search.vue
```diff
@@ -10,7 +10,7 @@
     </v-row>
 
     <v-row class="mt-4" no-gutters>
-      <v-col cols="5">
+      <v-col cols="4">
```
cameron-freshworks, line 13: Just curious here, but does going down to 4 columns cut off a few of verbiage after selecting an item?
For example, does `Change the name of an existing`.. selection still fit in the box after selection? 
I ask only because if it doesn't, UX might want to add a tooltip for any selections that are not fully visible just like the `Change (alter) business type of a`.. selection. 
Just a heads up!

client/src/components/new-request/search.vue
```diff
@@ -259,19 +266,13 @@ export default class NewSearch extends Vue {
...
-    }
-    if (Object.keys(xproMapping).includes(newVal) && ['CA', 'IN'].includes(this.location)) {
-      let { value } = newReqModule.entityTypesXPROData.find(ent => ent.rank === 1)
-      newReqModule.mutateEntityType(value)
-    }
     // Set default location to BC for the requests where BC is the only location option
     if (['CNV', 'MVE'].includes(newVal)) {
       this.location = 'BC'
+      this.jurisdictionDisabled = true
       return
```
cameron-freshworks, line 273: Is the jurisdiction field visible under either of these two requests? 
From what i recall, these were BC only, so the jurisdiction field might not even be viewable. 
Nothing major, but it might not be required to disable it if it's not rendered, not sure 🤔

client/src/components/new-request/search.vue
```diff
@@ -259,19 +266,13 @@ export default class NewSearch extends Vue {
...
     }
+    this.jurisdictionDisabled = false
```
cameron-freshworks, line 275: We might not need this property if the comment above is true.. though i could be wrong!

cameron-freshworks (approved):
Just had a few questions/comments, otherwise looks good!
Did you want us to pull and run this?

client/src/components/new-request/search.vue
```diff
@@ -120,7 +126,7 @@
     </v-row>
 
     <v-row class="mt-4" no-gutters>
-      <v-col cols="5" v-if="isXproMras">
+      <v-col cols="4" v-if="isXproMras">
```
severinbeauvais, line 129: You need to update the other v-col in this row to compensate for this change (line 144).

client/src/store/new-request-module.ts
```diff
@@ -1344,7 +1351,7 @@ export class NewRequestModule extends VuexModule {
   }
 
   get requestTextFromValue () {
-    if (this.request_action_cd) {
+    if (this.request_action_cd && this.requestTypeOptions.find(req => req.value === this.request_action_cd)) {
```
severinbeauvais, line 1354: This line is too long (per linter), and several others in this file.
Please run `npm run lint` to check.

client/tests/unit/analyze-pending.spec.ts
```diff
@@ -1,16 +1,34 @@
...
-import Comp from '@/components/new-request/analyze-pending.vue'
+import AnalyzePending from '@/components/new-request/analyze-pending.vue'
+import newReqModule from '@/store/new-request-module'
 
 const localVue = createLocalVue()
 const vuetify = new Vuetify()
 
 localVue.use(Vuetify)
 
+function setState () {
```
severinbeauvais, line 11: This should be `async`.

client/tests/unit/analyze-pending.spec.ts
```diff
@@ -1,16 +1,34 @@
...
-    localVue,
-    vuetify
+  let wrapper: any
+  beforeEach(async (done) => {
+    await setState()
+    wrapper = mount(AnalyzePending, {
+      localVue,
+      vuetify
+    })
+    done()
```
severinbeauvais, line 31: You don't need to call done(). We can chat separately on when done() is needed. (In a nutshell, never. It's better to use async/await.)

client/tests/unit/analyze-pending.spec.ts
```diff
@@ -1,16 +1,34 @@
...
 const vuetify = new Vuetify()
 
 localVue.use(Vuetify)
 
+function setState () {
+  newReqModule.mutateEntityType('CR')
+  newReqModule.mutateLocation('BC')
+  newReqModule.mutateRequestAction('NEW')
+  newReqModule.mutateAnalyzePending(true)
+  return new Promise(resolve => {
```
severinbeauvais, line 16: If you need an actual delay, I created a plugin, "sleep", for that purpose.

If you need to wait for Vue to finish its internal processing, `await Vue.nextTick()` is better.

But actually, here, you should just await each of the mutate calls (which are actually async). I tested this and it works :)

client/tests/unit/landing.spec.ts
```diff
@@ -64,6 +67,9 @@ describe('landing.vue', () => {
   let wrapper: any
 
   beforeEach(() => {
+    newReqModule.mutateRequestAction('NEW')
```
severinbeauvais, line 70: If these mutations only need to be done once in this describe block, they can be put in a "beforeAll()" method.

client/src/store/new-request-module.ts
```diff
@@ -1217,16 +1217,25 @@ export class NewRequestModule extends VuexModule {
   get locationOptions () {
     // To save template conditional logic, some locations have duplicate descriptions to align with there request
     let options = [
-      { text: 'BC', value: 'BC' },
+      {
+        text: 'BC',
+        value: 'BC',
+        blurbs: [
+          `Your business does not currently exist anywhere (i.e. it is a new business).`,
```
severinbeauvais, line 1224: "i.e." should have a comma after it.

client/src/components/new-request/search.vue
kialj876: Yeah I did this so that 'Select a Jurisdiction' wouldn't be cut off, which is specified in the ticket. Unfortunately I can't set the middle column to 3 and have the other 2 on either side fit properly. For now I set them all to 4, but we will see if UX will let this by - I may have to set the width differently

client/src/components/new-request/search.vue
```diff
@@ -19,43 +19,48 @@
...
                         @change="clearErrors()"
+                        label="Select an Action"
                         filled
                         v-model="request_action_cd">
                 <template slot="item" slot-scope="data">
                   <v-tooltip :disabled="!data.item.blurbs" right transition="fade-transition">
                     <template v-slot:activator="scope">
-                      <span v-on="scope.on" class="list-item">{{ data.item.text }}</span>
+                      <span v-on="scope.on" class="list-item">
+                        {{ data.item.text.replace('Relocate into', 'Relocate into BC') }}
```
severinbeauvais, line 33: It's better to not have a lot of code in the template, especially since this is JS instead of TS... specifically, I would like to ensure that "data.item.text" is non-null before chaining "replace", something like `data?.item?.text?.replace(etc)`.

Long story short, if it's a lot of code then use a method instead.

*Maybe* this one is OK if we can guarantee that data.item.text is never null... so then we choose either:
- the code will protect itself from an exception by adding checks, or
- let the code fail, and hope we notice errors when we test, and fix the cause

severinbeauvais: (A lot of this template code will fail if `data.item` is null.)

client/src/components/new-request/search.vue
kialj876: hey sorry cam, this is for disabling the 'location'. UX had me change the label to 'select a jurisdiction' so I thought this was a good name, but I could change this to disableLocation to make it less confusing if you like?

kialj876:
> Just had a few questions/comments, otherwise looks good!
> Did you want us to pull and run this?

Thanks cam, you can if you like, but I don't think you need to

client/src/store/new-request-module.ts
kialj876: oh oops, will update

client/src/components/new-request/search.vue
kialj876: ah you're right, will update

client/src/store/new-request-module.ts
kialj876: oh interesting, how did this pass the gha CI check? I will run it locally and update

client/src/components/new-request/search.vue
kialj876: okay good point, I'll add a method for this

client/src/store/new-request-module.ts
severinbeauvais: Maybe because it's a lint warning not an error?

client/src/components/new-request/search.vue
cameron-freshworks: Sounds good!

client/src/components/new-request/search.vue
cameron-freshworks: ahhh, sorry i missed that :) I got confused because we have a dedicated jurisdiction field as well.
Hmm yea, it might be good to have a distinction between the two fields, but not major at all.

client/tests/unit/analyze-pending.spec.ts
kialj876: good point!

client/tests/unit/analyze-pending.spec.ts
```diff
@@ -1,16 +1,24 @@
...
+  let wrapper: any
+  beforeAll(async () => {
+    newReqModule.mutateEntityType('CR')
+    newReqModule.mutateLocation('BC')
+    newReqModule.mutateRequestAction('NEW')
+    newReqModule.mutateAnalyzePending(true)
+    wrapper = mount(AnalyzePending, {
+      localVue,
+      vuetify
+    })
```
severinbeauvais, line 21: If it works... then it works and no need to change anything...

But in this case, to be explicit, I would have "awaited" each of the mutate statements.

Also, usually, in unit tests, we need to `await Vue.nextTick()` after mounting a component. This is needed to give Vue time to asynchronously finish whatever the component needs to do when it's mounted, before we can verify any elements or variables.

client/tests/unit/landing.spec.ts
```diff
@@ -63,7 +66,10 @@ describe('landing.vue', () => {
...
   let wrapper: any
 
-  beforeEach(() => {
+  beforeAll(() => {
+    newReqModule.mutateRequestAction('NEW')
+    newReqModule.mutateLocation('BC')
+    newReqModule.mutateEntityType('CR')
     newReqModule.mutateDisplayedComponent('AnalyzeResults')
 
     wrapper = mount(Landing, {
```
severinbeauvais, line 75: Again, no need to change anything right now, but for next time keep in mind that usually:
- beforeEach() will mount the component under test
- afterEach() will destroy the wrapper (ie, unmount and clean up memory)

severinbeauvais (approved):

client/tests/unit/analyze-pending.spec.ts
kialj876: ah okay I will keep this in mind for next time

PR #284, opened 2021-01-13
Handled Pending Payment state (fixes putNameReservation() error)

severinbeauvais:
*Issue #:* /bcgov/entity#6170

*Description of changes:*
- updated version to 0.3.1
- used NrState (enum) for NR states (multiple places)
- deleted obsolete submit() (2 instances)
- used PaymentStatus for payment statuses (multiple places)
- replaced Payment Status "model" with new enum
- fixed typo in Cond Reserved state
- added Pending Payment state
- handled Pending Payment state on put


cameron-freshworks (approved):
Looks great! What was the specific culprit causing the sentry error?

vysakh-menon-aot (approved):

severinbeauvais:
> Looks great! What was the specific culprit causing the sentry error?

The PENDING_PAYMENT state was not handled, so `data` was not set to anything... which caused the network call to fail.

client/src/store/new-request-module.ts
```diff
@@ -2329,22 +2329,29 @@ export class NewRequestModule extends VuexModule {
...
           data = this.editNameReservation
           break
+        case NrState.PENDING_PAYMENT:
+          // The user clicked Review and Confirm, which POSTed the draft NR.
+          // Then they closed the modal (eg, so they could fix something),
+          // and now they clicked Review and Confirm again.
+          // Treat this like a new NR, but keep the same state.
+          data = this.draftNameReservation
+          data['stateCd'] = NrState.PENDING_PAYMENT
+          break
```
severinbeauvais, line 2354: @cameron-freshworks This missing case is the root case.

PR #283, opened 2021-01-13
6148 Applicant Info UI update

vysakh-menon-aot:
*Issue #:* /bcgov/entity#6148

*Description of changes:*
  - Included review comments rfrom tracey on /bcgov/entity#6137

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the namerequest license (Apache 2.0).

PR #287, opened 2021-01-14
6173 Submit a Name Request for Review Screen UI update

vysakh-menon-aot:
*Issue #:* /bcgov/entity#6173

*Description of changes:*

### DIFF
--- a/client/src/components/common/names-capture.vue
+++ b/client/src/components/common/names-capture.vue
@@ -70,16 +70,25 @@
                               :name="Math.random()"/>
               </v-col>
               <v-col cols="4" class="py-0" v-if="designationAtEnd">
-                <v-select :error-messages="des1Message"
-                          :hide-details="hide"
-                          :items="items"
-                          :menu-props="props"
-                          :value="nameChoices.designation1"
-                          @blur="handleBlur(); showDesignationErrors.des1 = true"
-                          @input="editChoices('designation1', $event, true)"
-                          filled
-                          id="designation-1-select"
-                          label="Designation" />
+                <v-tooltip bottom transition="fade-transition" content-class="bottom-tooltip">
+                  <template v-slot:activator="{ on }">
+                    <div v-on="on">
+                      <v-select :error-messages="des1Message"
+                                :hide-details="hide"
+                                :items="items"
+                                :menu-props="props"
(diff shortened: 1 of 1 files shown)


severinbeauvais (commented):
Looks good but this is only a partial PR for the ticket right? I don't see the "Start Over" or colour changes here.

client/src/components/common/names-capture.vue
```diff
@@ -70,16 +70,28 @@
...
+                                @input="editChoices('designation1', $event, true)"
+                                filled
+                                id="designation-1-select"
+                                label="Designation" />
+                    </div>
+                  </template>
+                  <p>Your company name must include a designation.
+                  There are no legal differences between these designations.
+                  You can choose whichever you prefer.</p>
+                </v-tooltip>
```
cameron-freshworks, line 94: Not sure a refactor needs to happen in this PR but all of the selectors are very similar. It could be a good instance where we could break this out into a mini sub component to avoid duplication. 
Maybe i'm overlooking some complexity tho! Just a thought :)

vysakh-menon-aot:
> Looks good but this is only a partial PR for the ticket right? I don't see the "Start Over" or colour changes here.

That's already been covered in the previous PR which is yet to be merged

client/src/components/common/names-capture.vue
vysakh-menon-aot: I totally agree but it need more time

severinbeauvais:
> That's already been covered in the previous PR which is yet to be merged

[This PR](https://github.com/bcgov/namerequest/pull/283)? It refers to 2 different tickets. I was looking at the description of 6173.

vysakh-menon-aot:
> > That's already been covered in the previous PR which is yet to be merged
> 
> [This PR](https://github.com/bcgov/namerequest/pull/283)? It refers to 2 different tickets. I was looking at the description of 6173.

"Start Over" is in main-container.vue which is the only place I have to change

severinbeauvais (approved):

client/src/components/common/names-capture.vue
cameron-freshworks: Ok, fair enough!

cameron-freshworks (approved):
Merge when ready :)

PR #291, opened 2021-01-18
updated drop down validation

kialj876:

*Issue #:* /bcgov/entity#6191

*Description of changes:*
- drop down validation tweaks
- tests

### DIFF
--- a/client/src/components/new-request/search.vue
+++ b/client/src/components/new-request/search.vue
@@ -20,7 +20,6 @@
           <template v-slot:activator="scope">
             <div v-on="scope.on">
               <v-select :error-messages="errors.includes('request_action_cd') ? 'Please select an action' : ''"
-                        :hide-details="!errors.includes('request_action_cd')"
                         :items="requestActions"
                         @change="clearErrors()"
                         label="Select an Action"
@@ -51,8 +50,7 @@
                    :disabled="!location || location === 'BC'">
           <template v-slot:activator="scope">
             <div v-on="scope.on">
-              <v-select :error-messages="errors.includes('location') ? 'Please select a location' : ''"
-                        :hide-details="!errors.includes('location')"
+              <v-select :error-messages="errors.includes('location') ? 'Please select a jurisdiction' : ''"
                         :items="locationOptions"
                         :disabled="locationDisabled"
                         :readonly="!request_action_cd"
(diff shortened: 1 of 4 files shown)


client/src/store/new-request-module.ts
```diff
@@ -2675,7 +2675,7 @@ export class NewRequestModule extends VuexModule {
       name = sanitizeName(this.name)
     }
     ['entity_type_cd', 'request_action_cd', 'location'].forEach(field => {
-      if (this[field] === 'INFO') {
+      if (!this[field] || this[field] === 'INFO') {
```
severinbeauvais, line 2678: A comment describing what the code is doing here would be helpful. Eg, why check for empty?

client/tests/unit/new-search.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    })
+    newReqModule.mutateEntityType('CR')
+    newReqModule.mutateLocation('BC')
+    newReqModule.mutateRequestAction('NEW')
+    newReqModule.mutateName('test valid name')
+    await wrapper.vm.$nextTick()
+    wrapper.vm.clearErrors()
+  })
+
+  it('Resists submission when there is no action selected', async () => {
```
severinbeauvais, line 27: "Resists" LOL! (It's totally fine though.)

severinbeauvais (approved):

client/tests/unit/new-search.spec.ts
kialj876: haha there's a test for the name input component that is similar and it uses this wording lol

client/src/store/new-request-module.ts
kialj876: sure I can add something

client/src/components/new-request/search.vue
```diff
@@ -457,6 +449,9 @@ export default class NewSearch extends Vue {
 #goto-corporate-btn {
   min-height: 45px !important;
 }
+#name-input-component {
+  margin-top: 0 !important;
```
cameron-freshworks, line 453: Since this element only has this one style, you might be able to get rid of this and apply the class of `mt-0` on the element itself. I think* it does the same as `!important`

Either way, super small :)

cameron-freshworks (approved):
Super small comment, but looks good!

client/src/components/new-request/search.vue
kialj876: I did try that, but for some reason it gets overwritten by a different !important with mt=-20 after the errors are added. I didn't figure out why that was happening, but this gave the mt=0 more priority

PR #296, opened 2021-01-27
5757 NR Results pdf download

cameron-freshworks:
*Issue #:* /bcgov/entity#5757

*Description of changes:*

* Download PDF outputs for a given NR iD

* Loading Indicator


vysakh-menon-aot (approved):

client/src/components/existing-request/existing-request-display.vue
```diff
@@ -528,6 +530,9 @@ export default class ExistingRequestDisplay extends Mixins(
         case NrAction.RETRY_PAYMENT:
           this.navigateToPaymentPortal()
           break
+        case NrAction.RESULTS:
```
severinbeauvais, line 533: Does this button have a loading indicator while the results are being created/downloaded?

client/src/components/mixins/output-mixin.ts
```diff
@@ -0,0 +1,48 @@
...
+      const headers = { 'Accept': 'application/pdf' }
+
+      // show spinner since the network calls below can take a few seconds
+      this.$root.$emit('showSpinner', true)
+
+      // Request PDF for specified id
+      const response: any = await axios.get(url, { headers: headers, responseType: 'blob' as 'json' })
+
+      // Create a new blob object with mime-type explicitly set, otherwise only Chrome works
+      const blob = new Blob([response.data], { type: 'application/pdf' })
```
severinbeauvais, line 22: I think this code already exists in one of the services/modules. Can you just reuse it?

lekshmimallika-aot (approved):

client/src/components/existing-request/existing-request-display.vue
cameron-freshworks: I utilized the global spinner opposed to locking out the individual button. I figured since there was multiple actions a user could click on, the global spinner was a good choice. Any thoughts on this?

client/src/components/mixins/output-mixin.ts
cameron-freshworks: ah interesting! I'll see if i can find this.

cameron-freshworks: I see some code that is related to payments for receipts, it is different however. Is this the code you were referring too?

client/src/components/existing-request/existing-request-display.vue
severinbeauvais: Ah, that works.

I have no preference, other than having a common mechanism -- but which is the common mechanism in NR?

TLDR - you don't need to change this (unless, say, Scott/Tracey have a preference).

severinbeauvais: PS I think the UI would support downloading the results AND doing something else concurrently (unless there are global state conflicts). Again, I have no pref whether to block just this button or the whole UI.

PPS What does the error handling look like?

client/src/components/mixins/output-mixin.ts
severinbeauvais: That sounds like it. Doesn't it just download a PDF? I would rather see a single method for download a PDF. Maybe the receipt code needs to be refactored to use a common download method?

client/src/components/existing-request/existing-request-display.vue
cameron-freshworks: Oh, interesting, right!
Currently were logging the error but i will add something in for the UI.
ie a modal, although i didn't see a design for that. Possibly use a generic confirm, as a failure would/could just be tried again with no detriment.

client/src/components/mixins/output-mixin.ts
cameron-freshworks: I think if it was just hitting a different endpoint, making it generic could work but the request for outputs was purposely removed from the `/payments` endpoint and payment module. 
Imo, i think we leave payments ( receipts ) standalone, but in the future if we find we need a generic download, we modify this mixin method to handle that, is that ok?

client/src/components/existing-request/existing-request-display.vue
severinbeauvais: A simple error modal would be OK. The user can simply retry from the main page.

severinbeauvais: (The error modal may already display.)

cameron-freshworks: Added! Thanks, good catch 😁

client/src/components/mixins/output-mixin.ts
severinbeauvais: OK, but could you move the network code to one of the 2 places where there is other network code?

Or do you feel strongly that we should update the architecture to have separate mixins for separate actions?

cameron-freshworks: I don't feel strongly enough about it :) We can move it anywhere, although i prefer the mixin approach. 
What are the options, i don't see a place where it really fits other than the `newRequestModule`?
The requests made from `existing-Request-Display` refer to `newrequestModule`, so i guess that's the most logical.
(Although NOT a new request at this point 😆 )

cameron-freshworks: kk, relocated!

severinbeauvais (approved):
Looks good.

PR #303, opened 2021-02-12
updated environmental / config files

severinbeauvais:
*Issue #:* None

*Description of changes:*

I'm hoping that configuring this web app similar to our other Entity UIs will make it behave similarly. In particular I'm hoping these changes fix some of the "device specific" issues (which may be due to browser version / polyfills). But this is really a test. I don't expect this will break anything.

- updated browser list (same as filings)
- added missing babel compat-data package
- updated app patch version
- added dev packages (same as filings)
- excluded dist folder from ts files


severinbeauvais:
So this is weird. Locally, after my code changes, `npm run build` reported the same "lint" errors that have cause the checks to fail above. But... when I ran `npm run build` a second time, all was well.

Now, the GHA is reporting these build errors. So I will update the code to fix the "errors"...

cameron-freshworks (approved):
Looks good!

PR #305, opened 2021-02-12
Refactor local anchor to prevent page reload

cameron-freshworks:
*Issue #:* /bcgov/entity#6333

*Description of changes:*

* Refactor local anchor to prevent page reload

### DIFF
--- a/client/src/components/common/link-row.vue
+++ b/client/src/components/common/link-row.vue
@@ -12,7 +12,9 @@
     <a id="name-build-link"
        class="middle-link pt-2"
        :class="{ 'no-selector-link' : !entitySelectorUrl }"
-       href="#name-build-info"
+       href="#"
+       onclick="return false;"
+       @click="scrollTo('name-build-info')"
     >
       <v-col>
         <v-icon color="primary">mdi-help-circle-outline</v-icon> Learn how to build a name
@@ -33,11 +35,12 @@
 </template>
 
 <script lang="ts">
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Mixins } from 'vue-property-decorator'
 import newReqModule from '@/store/new-request-module'
+import CommonMixin from '@/components/mixins/common-mixin'
 
 @Component({})
-export default class LinkRow extends Vue {
+export default class LinkRow extends Mixins(CommonMixin) {
   /** Entity Selector Tool */
   private get entitySelectorUrl (): string {
     return sessionStorage.getItem('ENTITY_SELECTOR_URL')
--- a/client/src/components/mixins/common-mixin.ts
+++ b/client/src/components/mixins/common-mixin.ts
@@ -62,6 +62,6 @@ export default class CommonMixin extends Vue {
 
   /** Scroll to given element Id */
   private scrollTo (id) {
-    document.getElementById(id).scrollIntoView()
+    document.getElementById(id)?.scrollIntoView()
   }
 }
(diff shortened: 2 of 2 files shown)


client/src/components/common/link-row.vue
```diff
@@ -12,7 +12,9 @@
     <a id="name-build-link"
        class="middle-link pt-2"
        :class="{ 'no-selector-link' : !entitySelectorUrl }"
-       href="#name-build-info"
+       href="#"
+       onclick="return false;"
```
severinbeauvais, line 16: I think you can also do `onclick=""`.

(Or onclick.prevent or onclick.stop.)

Try it.

client/src/components/mixins/common-mixin.ts
```diff
@@ -59,4 +59,9 @@ export default class CommonMixin extends Vue {
     return [EntityType.XCR, EntityType.XUL, EntityType.RLC, EntityType.XLP, EntityType.XLL,
       EntityType.XCP, EntityType.XSO].includes(nr?.entity_type_cd)
   }
+
+  /** Scroll to given element Id */
+  private scrollTo (id) {
+    document.getElementById(id).scrollIntoView()
```
severinbeauvais, line 65: I would add conditional chaining to avoid problems if the id isn't found.

severinbeauvais: If possible, test this with Safari as well. I had to do the following to get scrolling working previously:
https://github.com/bcgov/business-filings-ui/blob/1c6a86ab2154646f806eb3d866bddd6beb56e32e/src/views/StandaloneDirectorsFiling.vue#L840

client/src/components/common/link-row.vue
cameron-freshworks: Kk, will do.

client/src/components/mixins/common-mixin.ts
cameron-freshworks: ok, good call. Will do!

client/src/components/common/link-row.vue
cameron-freshworks: I'm not entirely sure why, but neither had the same effect.
From what i read, it was one of two options that were recommended. 
`onclick="return false;"` or `href="javascript:void(0)"`

client/src/components/mixins/common-mixin.ts
cameron-freshworks: So it navigates to the section, but safari doesn't support SMOOTH scrolling

cameron-freshworks: To support smooth scrolling for Safari it is either..
a) Some custom javascript
b) Add package to handle this.

client/src/components/common/link-row.vue
severinbeauvais: OK, thanks for checking.

client/src/components/mixins/common-mixin.ts
severinbeauvais: Ask UX / POs where to go from here.

severinbeauvais: CSS transition?

severinbeauvais (approved):

client/src/components/mixins/common-mixin.ts
cameron-freshworks: Copy that, will do!

PR #306, opened 2021-02-12
more defensive code to prevent errors in Sentry

severinbeauvais:
*Issue #:* None

*Description of changes:*
- added conditional chaining to prevent "TypeError: Cannot read property 'end' of undefined"
https://sentry.io/organizations/registries/issues/2097009288/events/46ee541686f3469c9f8f072c459e1adf/

### DIFF
--- a/client/src/components/common/names-capture.vue
+++ b/client/src/components/common/names-capture.vue
@@ -814,12 +814,10 @@ export default class NamesCapture extends NameRequestMixin {
               newReqModule.mutateNameChoices({ key: `name${choice}`, value: newName })
             }
           }
-        } else {
-          if (this.$designations[this.entity_type_cd].end && nameChoices[`designation${choice}`]) {
-            if (!nameChoices[`name${choice}`].endsWith(nameChoices[`designation${choice}`])) {
-              let newName = nameChoices[`name${choice}`] + ' ' + nameChoices[`designation${choice}`]
-              newReqModule.mutateNameChoices({ key: `name${choice}`, value: newName })
-            }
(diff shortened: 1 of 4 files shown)


client/src/components/common/names-capture.vue
```diff
@@ -814,12 +814,10 @@ export default class NamesCapture extends NameRequestMixin {
               newReqModule.mutateNameChoices({ key: `name${choice}`, value: newName })
             }
           }
-        } else {
-          if (this.$designations[this.entity_type_cd].end && nameChoices[`designation${choice}`]) {
-            if (!nameChoices[`name${choice}`].endsWith(nameChoices[`designation${choice}`])) {
-              let newName = nameChoices[`name${choice}`] + ' ' + nameChoices[`designation${choice}`]
-              newReqModule.mutateNameChoices({ key: `name${choice}`, value: newName })
-            }
+        } else if (this.designationAtEnd && nameChoices[`designation${choice}`]) {
```
severinbeauvais, line 817: Having now replaced the 2 conditions with the getter, it now becomes obvious that this code can never run, because it is the else case for line 810 which is `if (this.designationAtEnd)`. Anyway that can be looked at in a different scope. Today I'm just trying to fix some easy Sentry errors.

PS
```
  get designationAtEnd () {
    if (this.entity_type_cd && this.$designations[this.entity_type_cd]) {
      return this.$designations[this.entity_type_cd].end
    }
    return false
  }

```

cameron-freshworks (approved):
Looks good

PR #308, opened 2021-02-19
changed compiler options

severinbeauvais:
*Issue #:* /bcgov/entity#6499

*Description of changes:*
- updated compiler options to target ES5/ES2015 (to support older browser engines)
- updated app version to 0.6.0

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "0.5.3",
+  "version": "0.6.0",
   "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


cameron-freshworks (approved):
Looks good. Is this a test?

I'm curious if TS will be impacted here.

severinbeauvais:
> Looks good. Is this a test?
> 
> I'm curious if TS will be impacted here.

Yes it's a test, because I really don't know why we're getting some of the errors we're seeing in Prod.

The TS is not affected. The generated JS is affected.

PR #310, opened 2021-02-22
added checks to prevent misc errors

severinbeauvais:
*Issue #:* None

*Description of changes:*
- added checks for valid nr.applicants object, which should fix this error:
   https://sentry.io/organizations/registries/issues/2098142767/events/90a90f68303243b29277a68f6233ec34/
- added check for valid summary.receipt object, which should fix this error:
   https://sentry.io/organizations/registries/issues/2098875580/events/332ddfeba28d47929f2eff9f516eb63c/
- added checks for valid querySelector function, which should fix these errors:
   https://sentry.io/organizations/registries/issues/2229040415/events/d3183f0ec1ad42fca073711a6bf19e2b/
   https://sentry.io/organizations/registries/issues/2227735742/events/77c9f37ef78f4a0d92db15c48f84986d/
   https://sentry.io/organizations/registries/issues/2227735759/events/a37fbf1afd984f08b5476f1fedde7351/

### DIFF
--- a/client/src/components/payment/payment-summary.vue
+++ b/client/src/components/payment/payment-summary.vue
@@ -1,7 +1,7 @@
 <template>
   <v-expand-transition>
-    <!-- do not display until payments are fetched -->
-    <section class="payment-summary" v-if="summary">
+    <!-- check for valid receipt, as the payment may exist but not be completed -->
+    <section class="payment-summary" v-if="summary.receipt">
       <v-row class="py-5" no-gutters>
         <div class="col1 align-self-center">{{receiptNumber}}</div>
         <div class="col2 align-self-center">{{receiptDate}}</div>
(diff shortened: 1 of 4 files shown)


cameron-freshworks (approved):
LGTM!

client/src/components/payment/payment-summary.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <v-expand-transition>
-    <!-- do not display until payments are fetched -->
-    <section class="payment-summary" v-if="summary">
+    <!-- check for valid receipt, as the payment may exist but not be completed -->
+    <section class="payment-summary" v-if="summary.receipt">
```
jordiwes, line 4: Quick question, just to be thorough, should this be v-if="summary && summary.receipt" ?

jordiwes (commented):
Other than above comment LGTM

client/src/components/payment/payment-summary.vue
severinbeauvais: Good question!

This component is in a v-for loop that iterates over summaries (and summary, a prop, is each item). So in theory this component is not rendered if the loop is empty, and summary is always defined. I tested a few scenarios locally to verify that this component doesn't have an undefined/null summary object.

Note that some components in Namerequest are pre-mounted and are then conditionally displayed. In that case, the component may exist without valid props. But not in this case (I think).

If you insist, I can add the original check back in. But I'd be more curious to see if it ever fails and then handle THAT situation :)

jordiwes: got it & do not insist ;)

jordiwes (approved):

PR #314, opened 2021-02-26
added submit button loading behaviour + updated email validation

severinbeauvais:
*Issue #:* /bcgov/entity#5110, /bcgov/entity#6574

*Description of changes:*
Commit 1:
   - updated app version

Commit 2:
   - updated email rules on page 2
   - updated email rules on page 3

Commit 3:
   - moved property not visible outside of mixin to module instead
   - moved "next" actions inside pages that use it (code simplification)
   - moved loading flag set/clear inside pages that use it (better code)
   - added some missing awaits
   - added optional chaining to querySelector calls
   - deleted some unused imports


client/src/components/common/applicant-info-1.vue
```diff
@@ -589,7 +589,7 @@ export default class ApplicantInfo1 extends NameRequestMixin {
 
   @Watch('isValid')
   onValidChanged (val: boolean) {
-    if (val) {
+    if (val && this.$el?.querySelector) {
```
severinbeauvais, line 592: I added this in several places. Would you believe `this.$el` is sometimes undefined? It's true!

(It doesn't really matter if the code in this block is run -- it's only for better diagnostics.

client/src/components/common/applicant-info-1.vue
```diff
@@ -603,7 +603,7 @@ export default class ApplicantInfo1 extends NameRequestMixin {
   nextAction () {
     this.validate()
     if (this.isValid) {
-      this.next()
+      newReqModule.mutateSubmissionTabNumber(this.submissionTabNumber + 1)
```
severinbeauvais, line 606: This was in `next()` for this page only.

severinbeauvais: This wonky architecture requires a developer to update things in MORE places rather than FEWER places when something needs to be changed :(

cameron-freshworks: Good change then!

client/src/components/common/applicant-info-2.vue
```diff
@@ -361,11 +364,15 @@ export default class ApplicantInfo2 extends NameRequestMixin {
...
+  async nextAction () {
+    newReqModule.mutateIsLoadingSubmission(true)
     this.validate()
     if (this.isValid) {
-      this.next()
+      await this.submit()
     }
+    // hang on to the loading state for a bit
+    // to prevent users clicking button again while next component displays
+    setTimeout(() => newReqModule.mutateIsLoadingSubmission(false), 1000)
```
severinbeauvais, line 375: I hope this makes sense. It feels a bit kludgy but I found that I was able to click the Submit button during the period between `this.submit()` and when the next component displays. I think it's all the reactive architecture in this app that makes things slow to respond.

client/src/components/common/applicant-info-nav.vue
```diff
@@ -9,22 +9,27 @@
...
+import { Component, Emit } from 'vue-property-decorator'
+import newRequestModule from '@/store/new-request-module'
 import NameRequestMixin from '@/components/mixins/name-request-mixin'
 
 @Component({})
 export default class ApplicantInfoNav extends NameRequestMixin {
   @Emit('nextAction')
   private nextAction () : void {}
 
+  get loading (): boolean {
```
severinbeauvais, line 29: Just a simplification to avoid putting `newRequestModule` in the template.

client/src/components/common/applicant-info-2.vue
```diff
@@ -228,10 +228,13 @@ export default class ApplicantInfo2 extends NameRequestMixin {
...
+    (v: string) => !!v || 'Required field',
+    (v: string) => {
+      // eslint-disable-next-line max-len
+      const pattern = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+      return pattern.test(v) || 'Not a valid email'
+    },
+    (v: string) => (!v || v.length <= 75) || 'Cannot exceed 75 characters'
   ]
   phoneRules = [
     v => !!v || 'Required field',
```
cameron-freshworks, line 240: Not related to your PR, but was going to check for this. 
I just found out this was required in OUR UI's, where i'm not sure it was before 👍

client/src/components/mixins/name-request-mixin.ts
```diff
@@ -143,16 +143,6 @@ export default class NameRequestMixin extends Vue {
     newRequestModule.mutateSubmissionTabNumber(this.submissionTabNumber - 1)
   }
 
-  private isloadingSubmission: boolean = false
```
severinbeauvais, line 146: It looks like this should work but it doesn't. This file is a mixin, so all the methods are inserted into `this` for the pages. However I found that `isLoadingSubmission` was always undefined in the pages, so the button loading state never worked.

client/src/components/mixins/name-request-mixin.ts
```diff
@@ -143,16 +143,6 @@ export default class NameRequestMixin extends Vue {
...
-  async next () {
-    if (this.submissionTabNumber === 3) {
-      this.isloadingSubmission = true
-      await this.submit()
-      return
-    }
-    newRequestModule.mutateSubmissionTabNumber(this.submissionTabNumber + 1)
-  }
```
severinbeauvais, line 154: This is the code I simply moved to the pages that use it. KISS!!!

client/src/components/mixins/name-request-mixin.ts
```diff
@@ -176,8 +166,7 @@ export default class NameRequestMixin extends Vue {
         if (!this.editMode && [NrState.COND_RESERVED, NrState.RESERVED].includes(this.nrState)) {
           request = await newRequestModule.getNameRequest(nrId)
           if (request?.stateCd === NrState.CANCELLED) {
-            newRequestModule.setActiveComponent('Timeout')
-            this.isloadingSubmission = false
```
severinbeauvais, line 180: I moved the clearing of the state to the pages that set it.

client/src/store/new-request-module.ts
```diff
@@ -3368,6 +3369,11 @@ export class NewRequestModule extends VuexModule {
   mutateAnalyzePending (value: boolean) {
     this.analyzePending = value
   }
+
+  @Mutation
+  mutateIsLoadingSubmission (val: boolean) {
+    this.isLoadingSubmission = val
+  }
```
severinbeauvais, line 3376: Though I hate to add to this monstrous module, it is a suitable place to keep this state property, And BTW you have to use `@Mutation` for this -- a regular setter doesn't work.

I'm looking forward to replacing this module with a simple, global store/state object.

cameron-freshworks (approved):
Looks good to me, get er in!

A few things to note for when we are diving back in (NR2.0) 
(not your work, just general thoughts)
- feels like a near duplication of Applicant Info Components
- having to set artificial delays between to prevent duplicated requests to the module.

client/src/store/new-request-module.ts
cameron-freshworks: Agrreeeeeeeeddddd

severinbeauvais:
> * feels like a near duplication of Applicant Info Components
> * having to set artificial delays between to prevent duplicated requests to the module.

I'm OK with the duplication (you mean pages 2 and 3, right?), since they are different enough.

Instead of the delays, I could have clear the loading flag in the "next component", but again that's having a feature in MORE components instead of FEWER, so I decided against it.

I don't have an easy solution to the problem of the button being re-enabled before the dialog covers it up.

severinbeauvais:
By the way, I did test the "Edit NR" flow locally. It's fine. (After Submit, there is no dialog. You just go back to the Existing NR page.)

jordiwes:
LGTM as well, don't know codebase well enough to comment intelligently but thanks for keeping me in the review loop

PR #317, opened 2021-02-28
deleted timer stuff + misc cleanup

severinbeauvais:
*Issue #:* /bcgov/entity#6581

*Description of changes:*
- ordered configuration same as in OCP
- deleted timer keys
- deleted timeout modal component
- deleted session timer mixin
- deleted timed module
- deleted timer state properties
- renamed models -> interfaces
- deleted coundown timer component
- deleted time state, actions, mutations
- deleted unused Analyze Name module


src/store/new-request-module.ts
```diff
@@ -2360,14 +2360,6 @@ export class NewRequestModule extends VuexModule {
 
       if (response?.data) {
         this.setNrResponse(response.data)
-
-        const { dispatch } = this.context
```
severinbeauvais, line 2364: Let's get rid of this style of getting properties!

severinbeauvais: I think this fails silently, and I think when the Vue instance restarts -- for whatever reason -- these additional properties added to context or `Vue.prototype` get cleared, which may also be a root cause of various problems.

cameron-freshworks (approved):

PR #320, opened 2021-03-02
Store directory structure

cameron-freshworks:
*Issue #:* /bcgov/entity#6581

*Description of changes:*

* Built new directory structure for Store.

* Broke out ListData, Actions, Mutations, Getters and State from Module.

* Prep for STORE swap.

* Minor Dev config updates.

*Notes:*

* This commit is the foundation for the store swap, i've tested some parts and the overall structure will work.

* Following this commit, i will swap the store and start working through each component and updating its usage of Store values, Getters, Actions. 

* Some actions will be relocated to Getters and some existing Getters in the Module will need to be ported as well.

* This commit doesn't break anything. 


severinbeauvais:
Excellent start! 👍

cameron-freshworks:
> Excellent start! 👍
> 
> > * This commit doesn't break anything.
> 
> Very good. This will make it easier to find anything we break.

The breaks will come soon Muahhaha 😈

.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     '@vue/typescript'
   ],
   rules: {
+    'no-console': 0,
```
severinbeauvais, line 13: How about ...?
`'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',`

src/interfaces/request-or-consent-interface.ts
```diff
@@ -0,0 +1,26 @@
+export interface RequestOrConsentIF {
+  0: {
```
severinbeauvais, line 2: Maybe keep a todo to change the property/state that uses this to an array instead?

src/interfaces/store-interfaces/index.ts
```diff
@@ -0,0 +1 @@
+export * from './action-interface'
```
severinbeauvais, line 1: Maybe in future don't both with this file and just re-export this IF from the parent index.ts ?

src/store/getters.ts
```diff
@@ -9,3 +32,355 @@ export const getNameRequestUrl = (state: any): string => {
 export const isAuthenticated = (state: any): boolean => {
   return Boolean(sessionStorage.getItem(SessionStorageKeys.KeyCloakToken))
 }
+
+/** Returns the name. */
```
severinbeauvais, line 36: Semantic, but these aren't functions that return something. They're getters, and appear like properties :)

src/store/getters.ts
```diff
@@ -9,3 +32,355 @@ export const getNameRequestUrl = (state: any): string => {
...
+}
+
+/** Returns the nr. */
+export const getNr = (state: StateIF): Partial<NameRequestI> => {
+  return state.stateModel.newRequestModel.nr
+}
+
+/** Returns the nr id. */
+export const getNrId = (state: StateIF): string => {
+  return state.stateModel.newRequestModel.nr.id
```
severinbeauvais, line 58: OK, you didn't use optional chaining. I think that's very good for now as it won't hide errors. We'll have to be careful where we get NR id, NR state, etc to check that we have a valid NR first.

src/store/getters.ts
```diff
@@ -9,3 +32,355 @@ export const getNameRequestUrl = (state: any): string => {
...
+  return state.stateModel.newRequestModel.doNotAnalyzeEntities
+}
+
+// /** Return the cancelled analysis boolean. */
+// export const getUserCancelledAnalysis = (state: StateIF): boolean => {
+//   return state.stateModel.newRequestModel.userCancelledAnalysis
+// }
+
+/** Return the BC Entity Types. */
+export const getEntityTypesBC = (state: StateIF): Array<EntityI> => {
```
severinbeauvais, line 152: For future: theoretically getters are cached (as long as nothing they depend on, of perhaps `state` doesn't change).

Does this create a static list of entity types in BC? If so, maybe this should be done up-front, once, somewhere else.

severinbeauvais: Ditto the other complex getters below.

src/store/getters.ts
```diff
@@ -9,3 +32,355 @@ export const getNameRequestUrl = (state: any): string => {
...
+    if (getLocation(state) === 'BC' && ['MVE'].includes(getRequestActionCd(state))) {
+      return 'mras'
+    }
+  }
+  return false
+}
+
+// MODAL GETTERS
+export const getIncorporateLoginModalVisible = (state: StateIF): boolean => {
+  return state.stateModel.newRequestModel.incorporateLoginModalVisible
```
severinbeauvais, line 385: FYI, component visibility state I hope to refactor out. 🙄

src/store/index.ts
```diff
@@ -17,5 +17,5 @@ export default new Vuex.Store<RootStateI>({
   state: {
     config: {} as EnvConfigI
   },
-  getters
+  getters: getters as any
```
severinbeauvais, line 20: Is this needed?

src/store/mutations.ts
```diff
@@ -0,0 +1,426 @@
...
+}
+
+export const mutateAnalysisJSON = (state: StateIF, analysisJSON: AnalysisJSONI) => {
+  state.stateModel.newRequestModel.analysisJSON = analysisJSON
+}
+
+export const mutateApplicant = (state: StateIF, appKV: any) => {
+  if (Array.isArray(appKV)) {
+    for (let address of appKV) {
+      state.stateModel.newRequestModel.applicant[address.name] = address.value
```
severinbeauvais, line 37: This is one of those properties that is sometimes undefined in Prod.

This is probably also why it's so slow to populate the address fields.

I'm not sure that I want to refactor this atm, but let's keep it in mind.

src/store/mutations.ts
```diff
@@ -0,0 +1,426 @@
...
+  state.stateModel.newRequestModel.tabNumber = tabNumber
+}
+
+export const mutateWaitingAddressSearch = (state: StateIF, appKV: WaitingAddressSearchI) => {
+  state.stateModel.newRequestModel.waitingAddressSearch = appKV
+}
+
+export const populateApplicantData = (state: StateIF) => {
+  for (let key in state.stateModel.newRequestModel.nr.applicants) {
+    Vue.set(
```
severinbeauvais, line 288: FYI, this looks like a work-around to make some things reactive, which IMO is an architectural issue :( which we are hopefully in the process of fixing :)

>Due to limitations in JavaScript, there are types of changes that Vue cannot detect. However, there are ways to circumvent them to preserve reactivity.

https://vuejs.org/v2/guide/reactivity.html

severinbeauvais (approved):

src/store/getters.ts
severinbeauvais: > We'll have to be careful where we get NR id, NR state, etc to check that we have a valid NR first.

... either with an explicit check, or better, the subject code doesn't get called if there is no NR in store.

src/interfaces/request-or-consent-interface.ts
cameron-freshworks: Cool i'll have a look!

src/interfaces/store-interfaces/index.ts
cameron-freshworks: My thoughts too, sort of redundant here.

severinbeauvais: Yeah, I did that with the 2 @/views/auth components.

src/store/getters.ts
cameron-freshworks: i see :) So should we avoid the doc-block comments or rephrase to `The Name` or `The Name Search` etc?

src/store/getters.ts
cameron-freshworks: You've read my mind, i recalled an earlier comment you made about hiding errors, which we didn't want quite yet.

src/store/getters.ts
severinbeauvais: In my IDE, hovering over these getters shows me the JsDoc stuff, so it's still valuable.

Yeah, I'd go with simply "The name." (but the name of what?)

src/store/getters.ts
cameron-freshworks: I know 😩  
The complex getters are 🤢 
I believe*, this list is dynamic based on the conditions in the Getter, so I'm not so sure it can be done once up front. 
These will be looked at further though!

src/store/getters.ts
severinbeauvais: PS Documenting the return type is redundant. Sometimes I'll use the @returns tag to give more context.

src/store/getters.ts
cameron-freshworks: Would be nice to have a singular action/mutation with params for all modal visibility states, but it sounds like we will be going another route. Using v-model maybe? What are you thinking?

src/store/getters.ts
severinbeauvais: OK, put it on your long-term todo list.

I'll RC you my list FYI.

src/store/index.ts
cameron-freshworks: It is temporarily due to all the added Getters where i define StateIF. 
In the updated index, the getters are typed correctly and not as `any`.

src/store/getters.ts
severinbeauvais: Or maybe a getter isn't the right pattern. Maybe we should these lists once, up-front, when the NR type is set/changed.

Medium-term future?

src/store/mutations.ts
cameron-freshworks: Absolutely! 
Leave all these comments open so we can look back :)

src/store/mutations.ts
cameron-freshworks: I was curious about the `Vue.set` approach. 
Wouldn't it be better to just build whatever object we wanted to set, then set it?

src/store/getters.ts
severinbeauvais: I've only made a todo list step so far 😁 

I don't have a solution yet.

src/store/mutations.ts
severinbeauvais: Not sure exactly what the problem is here, nor how to fix it.

I do know there's a thing here that doesn't work as expected and `Vue.set` is a way to work around it. Maybe it's a setter problem. Maybe it's a module problem. Maybe it's a deep-watch problem. 🤷‍♂️

PR #321, opened 2021-03-03
cleaned up dialogs, fixed a few things, etc

severinbeauvais:
*Issue #:* /bcgov/entity#6581

*Description of changes:*
- updated app version to 2.0.2
- added dev server paybc proxy
- added check before quick search to prevent errors
- moved modals -> dialogs
- cleaned up modals in App.vue
- started refactoring UI component architecture (WIP but working atm)
- moved some "not getters"
- commented out apparently-not-used module


public/config/configuration.json
```diff
@@ -13,7 +13,7 @@
     "NAME_REQUEST_URL": "https://dev.bcregistry.ca/namerequest",
     "NRO_URL": "https://dev.bcregistrynames.gov.bc.ca/nro/",
     "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
-    "QUICK_SEARCH_PUBLIC_ID": "namex-solr-service-account",
```
severinbeauvais, line 16: This is not a top secret but it doesn't need to be here either.

src/store/actions.ts
```diff
@@ -897,42 +897,50 @@ export const parseSynonymNames = (json: { names: [string], exactNames: [{ name:
...
-    commit('mutateDisplayedComponent', 'QuickSearchPending')
-    let encodedAuth = btoa(`${window['quickSearchPublicId']}:${window['quickSearchPublicSecret']}`)
+  const quickSearchPublicId = window['quickSearchPublicId']
+  const quickSearchPublicSecret = window['quickSearchPublicSecret']
 
-    const tokenResp = await axios.post(window['authTokenUrl'], 'grant_type=client_credentials', {
-      headers: { Authorization: `Basic ${encodedAuth}`, 'content-type': 'application/x-www-form-urlencoded' }
-    })
+  // only do quick search if we have id and secret
+  if (quickSearchPublicId && quickSearchPublicSecret) {
```
severinbeauvais, line 904: This is a fix to prevent an error when the quick search keys aren't populated (ie, when running locally). So it just skips this code.

I made this same fix to the old code in New Request Module.

severinbeauvais: (The diff here is mostly whitespace due to indentation.)

cameron-freshworks: ah was wondering about this one and why it didn't make the request locally. 👍

src/store/new-request-module.ts
```diff
@@ -2634,42 +2634,50 @@ export class NewRequestModule extends VuexModule {
 
   @Action
   async getQuickSearch (cleanedName: {exactMatch: string, synonymMatch: string}) {
-    try {
-      this.mutateDisplayedComponent('QuickSearchPending')
-      let encodedAuth = btoa(`${window['quickSearchPublicId']}:${window['quickSearchPublicSecret']}`)
+    const quickSearchPublicId = window['quickSearchPublicId']
+    const quickSearchPublicSecret = window['quickSearchPublicSecret']
```
cameron-freshworks, line 2638: A bummer you had to make these changes in 2 places! Next commit this module will go bye bye.

And by 'bye-bye', i mean placed to the side, because i'm sure we will need to reference it in case of weirdness 😁

vue.config.js
```diff
@@ -58,6 +58,13 @@ module.exports = {
         pathRewrite: {
           '/local-keycloak-config-url': ''
         }
+      },
+      // this is needed to avoid a PAYBC Not Found error when running locally
+      '/status/PAYBC': {
+        target: 'https://status-api-dev.apps.silver.devops.gov.bc.ca/api/v1/status/PAYBC',
```
cameron-freshworks, line 64: 👍

cameron-freshworks: Did they say how for long these urls will be changed?

cameron-freshworks (approved):

vue.config.js
severinbeauvais: They did not say, but as these are back-end URLs, they don't need to be vanity URLs and could just stay like that, I think.

PR #329, opened 2021-03-07
consistency, cleanup, fixes and defensive code

severinbeauvais:
*Issue #:* /bcgov/entity#6581

*Description of changes:*
- updated eslint rules (similiar to our other projects)
- added a few ignores (which we should verify/fix eventually)
- auto-fixed misc lint issues
- updated Sentry to latest version
- added Release tag to Sentry reporting
- added a temporary safety check to addrLine1
- refactored classname code to reduce errors
- fixed some rules errors
- other minor cleanup

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,13 +10,10 @@ module.exports = {
     '@vue/typescript'
   ],
   rules: {
-    'no-console': 0,
+    'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     'max-len': ['warn', { code: 120 }],
-    'no-unused-vars': 'off',
-    'quotes': 'off',
-    'no-useless-return': 'off',
-    'camelcase': 'off',
-    'no-useless-escape': 'off'
+    'camelcase': 'off' // FUTURE: fix code and remove this
   },
   parserOptions: {
     parser: '@typescript-eslint/parser',
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
   "dependencies": {
     "@babel/compat-data": "^7.12.13",
     "@mdi/font": "^4.5.95",
-    "@sentry/browser": "^5.29.1",
-    "@sentry/integrations": "^5.29.1",
+    "@sentry/browser": "^6.2.1",
+    "@sentry/integrations": "^6.2.1",
     "axios": "^0.21.1",
     "babel-plugin-transform-decorators": "^6.24.1",
     "core-js": "^3.8.0",
--- a/src/components/common/applicant-info-1.vue
+++ b/src/components/common/applicant-info-1.vue
@@ -347,7 +347,7 @@ export default class ApplicantInfo1 extends NameRequestMixin {
   isValid: boolean = false
   messages = {}
   provStateRules = [
-    v => typeof v === 'string' || 'Must be letters only',
+    v => (typeof v === 'string') || 'Must be letters only',
     v => (!v || v.length <= 2) || 'Max 2 characters'
   ]
   firstLastNameRules = [
@@ -589,13 +589,15 @@ export default class ApplicantInfo1 extends NameRequestMixin {
 
   @Watch('isValid')
   onValidChanged (val: boolean) {
-    if (val && this.$el?.querySelector) {
+    if (val) {
       this.$nextTick(() => {
-        // add classname to button text (for more detail in Sentry breadcrumbs)
-        const applicantBackBtn = this.$el.querySelector("#submit-back-btn > span")
-        if (applicantBackBtn) applicantBackBtn.classList.add("applicant-back-btn")
-        const applicantContinueBtn = this.$el.querySelector("#submit-continue-btn > span")
-        if (applicantContinueBtn) applicantContinueBtn.classList.add("applicant-continue-btn")
+        if (this.$el?.querySelector) {
+          // add classname to button text (for more detail in Sentry breadcrumbs)
+          const applicantBackBtn = this.$el.querySelector('#submit-back-btn > span')
+          if (applicantBackBtn) applicantBackBtn.classList.add('applicant-back-btn')
+          const applicantContinueBtn = this.$el.querySelector('#submit-continue-btn > span')
+          if (applicantContinueBtn) applicantContinueBtn.classList.add('applicant-continue-btn')
+        }
       })
     }
   }
--- a/src/components/common/applicant-info-2.vue
+++ b/src/components/common/applicant-info-2.vue
@@ -225,7 +225,7 @@ export default class ApplicantInfo2 extends NameRequestMixin {
   ]
   corpNumRules = [
     v => !!v || 'Required field',
-    v => !!this.getCorpNum(v) || 'Cannot validate number.  Please try again'
+    v => !!this.getCorpNum(v) || 'Cannot validate number. Please try again'
   ]
   emailRules = [
     (v: string) => !!v || 'Required field',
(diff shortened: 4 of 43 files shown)


.eslintrc.js
```diff
@@ -10,13 +10,10 @@ module.exports = {
...
-    'no-console': 0,
+    'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     'max-len': ['warn', { code: 120 }],
-    'no-unused-vars': 'off',
-    'quotes': 'off',
-    'no-useless-return': 'off',
-    'camelcase': 'off',
-    'no-useless-escape': 'off'
+    'camelcase': 'off' // FUTURE: fix code and remove this
```
severinbeauvais, line 16: I left this one because there were a ton of place to fix. Basically any props/methods that looked like `i_am_something` should be renamed `iAmSomething`, but this doesn't break anything.

I mean, it _could_ be fixed, but I'm not going to do it atm.

package.json
```diff
@@ -15,8 +15,8 @@
   "dependencies": {
     "@babel/compat-data": "^7.12.13",
     "@mdi/font": "^4.5.95",
-    "@sentry/browser": "^5.29.1",
-    "@sentry/integrations": "^5.29.1",
+    "@sentry/browser": "^6.2.1",
+    "@sentry/integrations": "^6.2.1",
```
severinbeauvais, line 19: I checked the release notes and did a bit of smoke testing, and it looks like nothing is broken.

src/components/common/applicant-info-1.vue
```diff
@@ -589,13 +589,15 @@ export default class ApplicantInfo1 extends NameRequestMixin {
...
   onValidChanged (val: boolean) {
-    if (val && this.$el?.querySelector) {
+    if (val) {
       this.$nextTick(() => {
-        // add classname to button text (for more detail in Sentry breadcrumbs)
-        const applicantBackBtn = this.$el.querySelector("#submit-back-btn > span")
-        if (applicantBackBtn) applicantBackBtn.classList.add("applicant-back-btn")
-        const applicantContinueBtn = this.$el.querySelector("#submit-continue-btn > span")
-        if (applicantContinueBtn) applicantContinueBtn.classList.add("applicant-continue-btn")
+        if (this.$el?.querySelector) {
```
severinbeauvais, line 594: My hope here is that, in the next tick, $el and querySelector exist (because for some unknown reason, they don't initially, which is why this check is here).

src/components/common/applicant-info-2.vue
```diff
@@ -238,7 +238,7 @@ export default class ApplicantInfo2 extends NameRequestMixin {
   ]
   phoneRules = [
     v => !!v || 'Required field',
-    v => (v.length <= 30) || 'Cannot exceed 30 characters'
+    v => (!v || v.length <= 30) || 'Cannot exceed 30 characters'
```
severinbeauvais, line 241: When v is null/undefined (which is weird because the first rule should have caught this, but it has definitely happened according to Sentry), this causes an error, which I think might be trashing the other rule validations, and letting required fields be empty.

So this may the fix for /bcgov/entity#6470.

src/components/dialogs/payment-complete.vue
```diff
@@ -127,7 +127,6 @@ export default class PaymentCompleteDialog extends Mixins(
     if (paymentSuccess) {
       this.$root.$emit('paymentComplete', true)
       await paymentModule.toggleReceiptModal(true)
-      return
```
severinbeauvais, line 130: Lint fix.

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -155,7 +155,7 @@ export default class PickEntityOrConversionDialog extends Vue {
     return `${210 * cols > maxThreshold ? maxThreshold : 210 * cols}px`
   }
 
-  entityBlurbs (entity_type_cd: string): Array<string> {
+  entityBlurbs (entity_type_cd: string): string[] {
```
severinbeauvais, line 158: So I've decided to use this array type declaration throughout this project.

src/components/existing-request/existing-request-display.vue
```diff
@@ -250,9 +250,15 @@ export default class ExistingRequestDisplay extends Mixins(
     })
   }
 
-  private get address () {
-    const fields = ['addrLine2', 'city', 'stateProvinceCd', 'countryCd', 'postalCd']
+  private get address (): string {
+    // FUTURE: delete this check as it hides an error that shouldn't happen
+    //         for now, report the error and don't crash
+    if (!this.nr.applicants) {
```
severinbeauvais, line 256: As per the comment, `this.nr` and/or `this.nr.applicants` should never be null/undefined, but according to Sentry this has definitely happened.

So this is a work-around for now and hides the error, but I added a console.error() to report it in Sentry so we remember to fix/verify this later.

src/components/existing-request/existing-request-search.vue
```diff
@@ -111,6 +111,7 @@ import newReqModule from '@/store/new-request-module'
 // import ErrorModule from '@/modules/error' // *** TODO: delete this?
 // import { NameRequestI, SearchDataI, NrDataResponseT, NrDataT } from '@/interfaces' // *** TODO: delete this?
 
+// eslint-disable-next-line no-useless-escape
```
severinbeauvais, line 114: Remove this override to see the error with the regex.

As long as the regex works, I'm not concerned about this. But does it work? 🤷‍♂️

src/components/existing-request/existing-request-search.vue
```diff
@@ -119,9 +120,11 @@ export default class ExistingRequestSearch extends Vue {
 
   mounted () {
     this.$nextTick(() => {
-      // add classname to button text (for more detail in Sentry breadcrumbs)
-      const retrieveNrBtn = this.$el.querySelector("#retrieve-name-btn > span")
-      if (retrieveNrBtn) retrieveNrBtn.classList.add("retrieve-nr-btn")
+      if (this.$el?.querySelector) {
```
severinbeauvais, line 123: This page, and a couple others, were missing this check... which I still think should not be necessary, but Sentry has reported errors here... 🤷‍♂️

severinbeauvais: BTW, it's not a big deal if this classname code doesn't run, as it's only to decorate the buttons for easier trouble-shooting in Sentry.

However, it is a big deal if code causes an error/exception, because nothing else in this method will run.

src/components/new-request/analyze-results.vue
```diff
@@ -823,7 +823,7 @@ export default class AnalyzeResults extends Vue {
         { insert: text }
       ])
     } catch {
-      return
+      // ignore errors
```
severinbeauvais, line 826: Lint fix.

src/main.ts
```diff
@@ -63,8 +64,11 @@ async function startVue () {
         'ResizeObserver loop limit exceeded',
         'ResizeObserver loop completed with undelivered notifications',
         // ignore Launch Darkly errors (partial match)
-        'LD: [error]'
-      ]
+        'LD: [error]',
+        // ignore Vuex errors (partial match)
+        '[vuex]'
```
severinbeauvais: I removed this vuex ignore line in a future commit, as I think we should report this in Prod. But I'm adding a comment to this commit because I found something interesting...

While testing the Release tag in Sentry, VS Code sent a patch to my browser, which output a ton of these "[vuex] duplicate getter" errors, followed by a bunch of "undefined this" and "undefined that" errors -- exactly what we see it Prod. So I have high hopes that the vuex module refactor will fix a ton of these errors!!!

(I don't know how users are getting these errors, since they're not using a debugger/patches in Prod, but the symptoms are the same so I expect the root cause is the same, too.)

cameron-freshworks (approved):

PR #332, opened 2021-03-07
Added Hotjar.

pwei1018:
*Issue #:* /bcgov/entity###

*Description of changes:*
1. add vue-hostjar library; 
2. add hotjar plugin into main.ts;
3. load hotjar id from configuration;

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "0.6.4",
+  "version": "0.6.5",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
@@ -29,6 +29,7 @@
     "style-loader": "^1.2.1",
     "vue": "^2.6.11",
     "vue-contenteditable": "^1.0.2",
+    "vue-hotjar": "^1.2.0",
     "vue-quill-editor": "^3.0.6",
     "vue-router": "^3.4.9",
     "vuetify": "^2.3.18",
--- a/client/public/config/configuration.json
+++ b/client/public/config/configuration.json
@@ -20,6 +20,7 @@
     "DEV_EXISTING_NR_TIMEOUT_MS": 0,
     "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
     "QUICK_SEARCH_PUBLIC_ID": "",
-    "QUICK_SEARCH_PUBLIC_SECRET": ""
+    "QUICK_SEARCH_PUBLIC_SECRET": "",
+    "HOTJAR_ID": "2284327"
   }
 ]
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -1,5 +1,6 @@
 import Vue from 'vue'
 import App from './App.vue'
+import Hotjar from 'vue-hotjar'
 import { getVueRouter } from '@/router'
 import store from './store'
 import { EnvConfigI, getConfig } from '@/plugins/getConfig'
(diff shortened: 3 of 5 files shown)


client/src/main.ts
```diff
@@ -72,6 +73,16 @@ async function startVue () {
   Vue.prototype.$USAStateCodes = USAStateCodes
   Vue.prototype.$xproMapping = mapping.xproMapping
 
+  // Setup Hotjar
+  // Initialize Launch Darkly
```
severinbeauvais, line 77: Obsolete comment.

client/src/main.ts
```diff
@@ -72,6 +73,16 @@ async function startVue () {
...
+  if (window['hotjarId']) {
+    console.info('Initializing Hotjar...') // eslint-disable-line no-console
+    Vue.use(Hotjar, {
+      id: window['hotjarId'],
+      isProduction: true
```
severinbeauvais, line 82: Should this be changed for non-production? Maybe add something like this:
isProduction: (process.env.NODE_ENV === 'production')

client/public/config/configuration.json
```diff
@@ -20,6 +20,7 @@
     "DEV_EXISTING_NR_TIMEOUT_MS": 0,
     "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
     "QUICK_SEARCH_PUBLIC_ID": "",
-    "QUICK_SEARCH_PUBLIC_SECRET": ""
+    "QUICK_SEARCH_PUBLIC_SECRET": "",
+    "HOTJAR_ID": "2284327"
```
severinbeauvais, line 24: We try to not include IDs in the local config (since it's committed to GH). This should be in the config maps only.
(And maybe different IDs for dev/test/prod ?)

client/package.json
```diff
@@ -29,6 +29,7 @@
     "style-loader": "^1.2.1",
     "vue": "^2.6.11",
     "vue-contenteditable": "^1.0.2",
+    "vue-hotjar": "^1.2.0",
```
severinbeauvais, line 32: Please also update app version by 0.0.1.

severinbeauvais:
Is this new feature temporary or ongoing? If it will be used on an ongoing basis, please port these changes to "release-2.0" branch as well. Thanks!

client/src/main.ts
pwei1018: fixed.

client/src/main.ts
pwei1018: We want to open this option in dev and test. Otherwise, we can get the data from hotjar.

isProduction:
If you would like to disable or enable tracking, pass in either true or false. It is advised to bind your Node ENV variable. This is an optional parameter and will default to true if not specified.

client/public/config/configuration.json
pwei1018: This id is for local site. All other environments are using different id.

client/package.json
pwei1018: Version bump.

client/src/main.ts
severinbeauvais: OK, so we this to run in all environments.

I suggest you delete this property, since it looks like an error, and it defaults to True anyway.

client/public/config/configuration.json
severinbeauvais: Oh OK.

Will you show us how to use Hotjar at the next tech review?

client/package.json
severinbeauvais: For some reason, the version in the package-lock file did not get updated. Please update manually or do `npm i` again.

pwei1018:
> Is this new feature temporary or ongoing? If it will be used on an ongoing basis, please port these changes to "release-2.0" branch as well. Thanks!

It should be ongoing. Right now, I just add it as "proof of concept" feature. I have to go over with Kaine to see what do we need to do for it next. Thanks.

severinbeauvais:
> > Is this new feature temporary or ongoing? If it will be used on an ongoing basis, please port these changes to "release-2.0" branch as well. Thanks!
> 
> It should be ongoing. Right now, I just add it as "proof of concept" feature. I have to go over with Kaine to see what do we need to do for it next. Thanks.

Sounds good. Please remember to port it to Rls 2 if needed.

client/package.json
pwei1018: Fixed.

client/src/main.ts
pwei1018: Removed as your recommend.

severinbeauvais (approved):
LGTM

PR #334, opened 2021-03-08
[entity6500] Adding Env Config / Feature Flag

jeznorth:
*Issue #:* /bcgov/entity#6500

Adding env config and feature flag

### DIFF
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -68,6 +68,7 @@
 </template>
 
 <script lang="ts">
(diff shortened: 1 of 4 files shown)


client/src/App.vue
```diff
@@ -68,6 +68,7 @@
 </template>
 
 <script lang="ts">
+import axios, { AxiosError, AxiosRequestConfig } from 'axios'
```
severinbeauvais, line 71: You don't need this.

PR #338, opened 2021-03-11
Store swap

cameron-freshworks:
*Issue #:* /bcgov/entity#6581

*CAUTION:* 

- I'm not sure we want to consider merges in the current state. Although it would be nice to have help working through all the breaks 😁 

- I'm open to ideas!

*Description of changes:*

* Ongoing store swap.

* Currently working through the flows and correcting errors, applying Typings as i go.

* Working on a way to integrate the modules into our store, otherwise it will require some re-working.


src/components/common/applicant-info-1.vue
```diff
@@ -329,10 +329,15 @@
...
-import newReqModule from '@/store/new-request-module'
+import { Action, Getter } from 'vuex-class'
+
 import ApplicantInfoNav from '@/components/common/applicant-info-nav.vue'
 import { Location } from '@/enums'
 import { NameRequestMixin } from '@/mixins'
+import { ActionBindingIF } from '@/interfaces/store-interfaces'
+import { ApplicantI, NameRequestI, SubmissionTypeT } from '@/interfaces'
+import $canJurisdictions from '@/store/list-data/canada-jurisdictions'
+import $intJurisdictions from '@/store/list-data/intl-jurisdictions'
```
severinbeauvais, line 340: What do you think about these -- should be part of the store (instead of just in the store folder)? Or perhaps a separate resource folder?

BTW I don't like their name. `$xxx` is used in other frameworks to identify an [observable](https://www.freecodecamp.org/news/what-are-observables-how-they-are-different-from-promises/). In Vue perhaps they refer to Vue's internal methods (eg, "$refs"). Naming them as they are here is misleading and giving them too much credit.

src/components/common/applicant-info-2.vue
```diff
@@ -179,7 +179,7 @@
             <template v-slot:activator="{ on }">
               <div v-on="on">
                 <v-checkbox
-                  v-if="showPriorityRequest"
+                  v-if="getShowPriorityRequest"
```
severinbeauvais, line 182: Don't change now, but if you get a change later, I recommend changing booleans getters to "isXXX" or "doXXX". (low priority)

src/components/common/applicant-info-2.vue
```diff
@@ -304,13 +298,26 @@ export default class ApplicantInfo2 extends NameRequestMixin {
...
+  get showAllFields () {
+    return (!this.getEditMode || this.getNrState === 'DRAFT')
   }
 
-  async getCorpNum (num) {
+  get xproJurisdiction () {
+    return (this.getNrData || {}).xproJurisdiction
+  }
+
+  async corpNumHandler (num) {
```
severinbeauvais, line 320: Rename to `validateCorpNum`.

src/components/common/applicant-info-nav.vue
```diff
@@ -9,58 +9,69 @@
...
+import { Component, Emit, Vue } from 'vue-property-decorator'
+import { Action, Getter } from 'vuex-class'
+import { SubmissionTypeT } from '@/interfaces'
+import { ActionBindingIF } from '@/interfaces/store-interfaces'
 
 @Component({})
-export default class ApplicantInfoNav extends NameRequestMixin {
-  @Emit('nextAction')
-  private nextAction () : void {}
+export default class ApplicantInfoNav extends Vue {
```
severinbeauvais, line 26: This is so so so so so so much better than what was there before!

I'm still sick to my stomach that there is so much left to clean up...

severinbeauvais: But we'll get there. I'm confident that what you're doing here will solve 80% of the issues, maybe more.

src/components/common/names-capture.vue
```diff
@@ -280,18 +275,50 @@
...
+  @Getter getEntityTypeOptions!: Array<EntityI>
+  @Getter getIsAssumedName!: boolean
+  @Getter getLocation!: LocationT
+  @Getter getLocationOptions!: Array<any>
+  @Getter getName!: string
+  @Getter getNameChoices!: NameChoicesIF
+  @Getter getNr!: Partial<NameRequestI>
+  @Getter getRequestActionCd!: string
+  @Getter getRequestTypeOptions!: RequestActionsI[]
+  @Getter getSubmissionTabNumber!: number
```
severinbeauvais, line 305: Wow, types :)

src/components/common/names-capture.vue
```diff
@@ -876,6 +858,55 @@ export default class NamesCapture extends NameRequestMixin {
...
+
+  @Watch('location')
+  handleLocation (newVal, oldVal) {
+    if (newVal !== oldVal) {
+      this.setNRData({ key: 'xproJurisdiction', value: '' })
+    }
+    if (newVal === 'INFO') {
+      let type = this.entity_type_cd
+      this.setLocationInfoModalVisible(true)
+      this.$nextTick(function () {
```
severinbeauvais, line 898: This is a strong code smell -- it's a workaround for something async not working right.

We should try to come back to this eventually and make it work right.

src/components/dialogs/exit.vue
```diff
@@ -22,18 +22,22 @@
...
+
   get showModal () {
-    return newReqModule.exitModalVisible
+    return this.getExitModalVisible
   }
   set showModal (value: boolean) {
   }
   hideModal () {
-    newReqModule.mutateExitModalVisible(false)
+    this.setExitModalVisible(false)
```
severinbeauvais, line 40: Have you formed an idea how to get rid of this visibility state? It feels wrong but I don't have a solution yet. Maybe we can treat like our other dialogs: the parent passes in the visibility prop? I dunno, is that any better? I don't mind our implementation.

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -81,13 +81,36 @@
...
+
+  // Global Actions
+  @Action setConversionType!: ActionBindingIF
+  @Action setConversionTypeAddToSelect!: ActionBindingIF
+  @Action setEntityTypeCd!: ActionBindingIF
+  @Action setEntityTypeAddToSelect!: ActionBindingIF
+  @Action setPickEntityModalVisible!: ActionBindingIF
+
+  private showSocietiesInfo = false
+  private conversionTypes = ConversionTypes
```
severinbeauvais, line 113: Haha, so much better than
```
get conversionTypes () {
    return newReqModule.conversionTypes
  }
```

If this is just to make ConversionTypes visible to the template, you can use:
`readonly ConversionTypes = ConversionTypes`

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -96,56 +119,76 @@ export default class PickEntityOrConversionDialog extends Vue {
...
-    return newReqModule.entity_type_cd
+    return this.getEntityTypeCd
   }
 
   set entity_type_cd (value) {
-    newReqModule.mutateEntityType(value)
-  }
-
-  get isConversion () {
-    return (newReqModule.request_action_cd === 'CNV')
```
severinbeauvais, line 112: You replaced this with `getIsConversion` but I didn't see where it's declared. Does it replace this code?

src/components/dialogs/pick-request-type.vue
```diff
@@ -47,48 +47,49 @@
...
   }
 
   set showModal (value: boolean) {
-    newReqModule.mutatePickRequestTypeModalVisible(value)
+    this.setPickRequestTypeModalVisible(value)
   }
 
   get tableData () {
-    let { length } = this.requestActions
+    let { length } = RequestActions
```
severinbeauvais, line 78: This looks odd because the right hand side is PascalCase, which usually means it's a a type. I'd be tempted to rename it into camelCase.

In this case, I gather it's an array. But can it really be destructured this way to get array length???

Or I'd be explicit that I'm doing `const length = RequestActions.length`.

src/components/new-request/stats.vue
```diff
@@ -58,26 +58,30 @@
...
     }
   }
 
-  get stats (): StatsI {
-    return newReqModule.stats
-  }
-
   get autoApprovedCount (): string | number {
-    return (this.stats?.auto_approved_count || '0')
+    return (this.getStats?.data?.auto_approved_count ?? '-')
```
severinbeauvais, line 84: I think you mean `||` not `??`. See also lines 93 and 103 below.

src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -58,32 +55,33 @@ export default class ReserveSubmitButton extends Vue {
...
   async sendToExamination () {
-    await newReqModule.userClickedStopAnalysis()
-    newReqModule.cancelAnalyzeName('NamesCapture')
+    await this.userClickedStopAnalysis(null)
+    this.cancelAnalyzeName('NamesCapture')
   }
 
   handleSubmit () {
-    let { setup, location, entity_type_cd, request_action_cd } = this
+    let { setup, getLocation } = this
```
severinbeauvais, line 63: We already have a getter for `getLocation`. Let's use that instead.

src/interfaces/models.ts
```diff
@@ -246,9 +246,11 @@ export interface SetupI {
 }
 
 export interface StatsI {
-  auto_approved_count: number
-  priority_wait_time: number
-  regular_wait_time: number
+  data: {
```
severinbeauvais, line 249: Nice!

src/interfaces/models.ts
```diff
@@ -96,8 +96,8 @@ export interface EntityI {
   intBlurbs?: string[]
   mveBlurbs?: string[]
   rehBlurbs?: string[]
-  amlBlurbs?: string[] | [string[], string[]?]
-  chgBlurbs?: string [] | [string[], string[]?]
+  amlBlurbs?: string[]
+  chgBlurbs?: string[]
```
severinbeauvais, line 100: The new code isn't equivalent to the old. Is it OK?

src/main.ts
```diff
@@ -83,7 +84,7 @@ async function startVue () {
   new Vue({
     vuetify: getVuetify(),
     router: getVueRouter(),
-    store,
+    store: store,
```
severinbeauvais, line 87: Old code was OK.

Does this test through to payment OK? I tried using `getVuexStore` but it broken something due to the dynamic store modules.

src/mixins/payment/payment-mixin.ts
```diff
@@ -60,7 +60,7 @@ export class PaymentMixin extends Vue {
 
   get paymentSummaries () {
     const payments = this.$store.getters[paymentTypes.GET_PAYMENTS]
-    const summaries = payments.map(payment => {
+    const summaries = payments && payments.map(payment => {
```
severinbeauvais, line 63: Could also use conditional chaining.

What should be returned if there are no payments, `[]` ?

src/modules/error/index.ts
```diff
@@ -1,8 +1,10 @@
 import { Module, VuexModule, getModule } from 'vuex-module-decorators'
-import store from '@/store'
+import { getVuexStore } from '@/store'
 
 import errorStore from './store'
 
+const store = getVuexStore()
+
 @Module({ dynamic: true, namespaced: false, store, name: 'errorModule' })
```
severinbeauvais, line 8: Yeah this is what I tried initially. Something was getting corrupt but I had to test all the way through to the payment summary to see the error. YMMV.

src/store/actions.ts
```diff
@@ -1132,14 +1136,205 @@ export const fetchMRASProfile: ActionIF = async ({ commit, getters }): Promise<a
   return null
 }
 
+/** Submits an edited NR or a new name submission. */
+export const submit: any = async ({ commit, getters }): Promise<any> => {
```
severinbeauvais, line 1140: Need to refactor this (after merge). I don't think a store action should **ever** make a network call !!!

If this functionality is truly needed, there should be a high-level function that does the store update and the network call. Maybe in a mixin?

src/store/actions.ts
```diff
@@ -1132,14 +1136,205 @@ export const fetchMRASProfile: ActionIF = async ({ commit, getters }): Promise<a
...
+    if (!getters.isEditMode && [NrState.COND_RESERVED, NrState.RESERVED].includes(getters.getNrState)) {
+      request = await getNameRequest(getters.getNrId)
+      if (request?.stateCd === NrState.CANCELLED) {
+        await setActiveComponent('Timeout')
+        return
+      }
+    }
+    request = await putNameReservation(getters.getNrId)
+  }
+  console.log('got Somewhere')
```
severinbeauvais, line 1164: Oh good. Better than getting nowhere.
😁

src/store/actions.ts
```diff
@@ -354,7 +361,7 @@ export const confirmAction: ActionIF = async ({ commit }, action: string): Promi
 
 export const findNameRequest: ActionIF = async ({ commit, getters }): Promise<void> => {
```
severinbeauvais, line 362: This looks like something else that should be a higher-level function (not in actions!).

src/store/actions.ts
```diff
@@ -977,20 +982,19 @@ export const startQuickSearch = async () => {
 }
 
 export const startAnalyzeName: ActionIF = async ({ commit, getters }) => {
```
severinbeauvais, line 984: Should not be in actions.

Lots of others like that in here.

src/store/actions.ts
```diff
@@ -977,20 +982,19 @@ export const startQuickSearch = async () => {
...
     name = sanitizeName(getters.getName)
   }
-  ['entity_type_cd', 'request_action_cd', 'location'].forEach(field => {
-    // These are initialized empty. Check to make sure user selected an option for each
-    if (!this[field] || this[field] === 'INFO') {
-      commit('setErrors', field)
-    }
-  })
+
+  if (!getters.getRequestActionCd) commit('setErrors', 'request_action_cd')
```
severinbeauvais, line 992: I'm not seeing how the old code became the new code here ...?

src/store/actions.ts
```diff
@@ -1029,10 +1033,10 @@ export const startAnalyzeName: ActionIF = async ({ commit, getters }) => {
       }
     }
   }
-  if (getters.getQuickSearch) {
-    await startQuickSearch()
-    return
-  }
+  // if (getters.getQuickSearch) {
```
severinbeauvais, line 1036: TODO: re-enable this eventually

src/store/getters.ts
```diff
@@ -107,11 +146,85 @@ export const getHasNoCorpNum = (state: StateIF): boolean => {
   return state.stateModel.newRequestModel.noCorpNum
 }
 
+export const getCorpNum = (state: StateIF): string => {
+  return state.stateModel.newRequestModel.corpNum
```
severinbeauvais, line 150: Does `newRequestModel` make sense to you?
It should almost be called `everythingModel`... I'm kinda wondering if it's helpful to have this object or not. Thoughts?

src/store/getters.ts
```diff
@@ -122,10 +235,191 @@ export const getDoNotAnalyzeEntities = (state: StateIF): string[] => {
...
+export const getAnalysisJSON = (state: StateIF): AnalysisJSONI => {
+  return state.stateModel.newRequestModel.analysisJSON
+}
+
+export const getIsPriorityRequest = (state: StateIF): boolean => {
+  return state.stateModel.newRequestModel.priorityRequest
+}
+
+export const getCurrentIssue = (state: StateIF): IssueI => {
+  if (!getAnalysisJSON(state)) return
```
severinbeauvais, line 316: return null?

also line 321

src/store/getters.ts
```diff
@@ -225,9 +519,33 @@ export const getEntityTypesXPRO = (state: StateIF): Array<EntityI> => {
...
+    }
+  }
+  return ''
+}
+
 /** Get entity type options. */
-export const getEntityTypeOptions = (state: StateIF) => {
-  let bcOptions: SelectOptionsI[] = getEntityTypesBC(state).filter(x => {
+export const getEntityTypeOptions = (state: StateIF): Array<EntityI> => {
+  let bcOptions: SelectOptionsI[] = getEntityTypesBC(state)?.filter(x => {
```
severinbeauvais, line 548: Did you add this because you found an error?

The only problem is this might be masking an error, and we should fix the root cause instead.

severinbeauvais: In a couple places, I added some "safety checks" like this, but I also put in a console.error() to report failures (and saw some, too). So there are Sentry errors that I can go back and look at and try to fix the root cause.

src/store/getters.ts
```diff
@@ -359,7 +691,201 @@ export const getShowCorpNum = (state: StateIF): 'colin' | 'mras' | false => {
...
+      }
+    }
+
+    return { ...requestName } as RequestNameI
+  })
+
+  return requestNames
+}
+
+export const getDraftNameReservation = (state: StateIF): DraftReqI => {
```
severinbeauvais, line 826: I think this should be out of the store and in a mixin or something. It can always use the getters to get its data. I'm nervous about it using `state`.... who knows if state is updated when this code runs?

severinbeauvais (approved):
Wow, so much work, and so much better !!!

A few comments here. Please triage them. I'd like you to verify some of them before merging. Others can be handled later (keep a list).

Do you want to fix the merge conflicts or do you want me to take a stab at it? Can be merged after that if you're happy with it.

src/components/common/applicant-info-2.vue
severinbeauvais: Don't change now.

src/components/common/applicant-info-2.vue
severinbeauvais: fixed

src/App.vue
```diff
@@ -56,7 +56,7 @@
     <LocationInfoDialog />
     <MrasSearchInfoDialog />
     <NrNotRequiredDialog />
-    <PaymentDialog :onCancel="onPaymentCancelled" />
+<!--    <PaymentDialog :onCancel="onPaymentCancelled" />-->
```
severinbeauvais, line 59: This needs to be uncommented eventually.

src/interfaces/models.ts
```diff
@@ -92,12 +92,12 @@ export interface DraftReqI {
 }
 
 export interface EntityI {
-  blurbs?: string[] | string
+  blurbs?: string | string[] | string[][] // TODO: fix wtf here
```
severinbeauvais, line 95: This doesn't look right but it's what TS wanted. Have a look.

Ditto on lines 99-100.

src/mixins/action-mixin.ts
```diff
@@ -0,0 +1,116 @@
...
+import querystring from 'qs'
+import axios from 'axios'
+import { Debounce } from 'vue-debounce-decorator'
+import canadaPostAPIKey from '@/store/config'
+import { ActionBindingIF } from '@/interfaces/store-interfaces'
+import { ApplicantI } from '@/interfaces'
+import { ApiErrorMixin } from '@/mixins'
+
+@Component
+export class ActionMixin extends Mixins(ApiErrorMixin) {
```
severinbeauvais, line 12: New mixin with stuff I pulled out of `store/actions.ts` (hence the name).

Do what you want with this. But we need somewhere for these methods to live where they have `this` in scope (so, not the store actions).

src/mixins/action-mixin.ts
```diff
@@ -0,0 +1,116 @@
...
+  qs: any = querystring
+
+  // Global getter
+  @Getter getApplicant!: ApplicantI
+
+  // Global action
+  @Action setApplicantDetails!: ActionBindingIF
+
+  // async fetchAddressDetails (id): Promise<void> {
```
severinbeauvais, line 21: This is commented out as I couldn't find anywhere it was used.

src/mixins/api-error-mixin.ts
```diff
@@ -0,0 +1,49 @@
+import { Component, Vue } from 'vue-property-decorator'
+import { AxiosError } from 'axios'
+
+@Component
+export class ApiErrorMixin extends Vue {
```
severinbeauvais, line 5: Another new mixin with `handleApiError()` for everyone to use.

severinbeauvais: I think I got rid of the other handleApiError but please verify.

tests/unit/analyze-pending.spec.ts
```diff
@@ -1,29 +1,29 @@
-import { createLocalVue, mount } from '@vue/test-utils'
-import Vuetify from 'vuetify'
-import AnalyzePending from '@/components/new-request/analyze-pending.vue'
-import newReqModule from '@/store/new-request-module'
+// import { createLocalVue, mount } from '@vue/test-utils'
```
severinbeauvais, line 1: I had to full-on comment the tests because they had errors importing things and then using them.

src/mixins/action-mixin.ts
severinbeauvais: (I think this used to be called `setAddressDetails`.)

src/App.vue
```diff
@@ -156,23 +161,24 @@ export default class App extends Vue {
...
-    // Only do this for New NRs!!!
-    if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
-      // Cancel the NR using the rollback endpoint if we were processing a NEW NR
-      await newRequestModule.rollbackNameRequest({ nrId, action: RollbackActions.CANCEL })
-      // Direct the user back to the start
-      await this.resetAppState()
-    }
-    await paymentModule.togglePaymentModal(false)
-  }
+  // TODO: Update Pay module
```
severinbeauvais, line 169: To update.

src/components/common/names-capture.vue
```diff
@@ -315,35 +342,35 @@ export default class NamesCapture extends NameRequestMixin {
...
-    if (this.isAssumedName && !this.editMode) {
+    this.setNameChoicesToInitialState(null)
+    if (this.getIsAssumedName && !this.getEditMode) {
       this.$nextTick(() => { this.hide = true })
       return
     }
 
     this.$nextTick(() => {
-      if (this.editMode) {
+      if (this.getEditMode) {
```
severinbeauvais, line 352: FWIW, this code allows execution to continue to lines 356-375, and after that it executes this arrow function. So quite likely this isn't working as intended.

This could initially be solved with an `await this.nextTick()`, or ultimately with better async design.

I'm going to fix this now because there's the same issue on line 347...

src/components/dialogs/pick-entity-or-conversion.vue
severinbeauvais: found it

src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: fixed

src/store/mutations.ts
```diff
@@ -360,7 +362,7 @@ export const setNrResponse = (state: StateIF, data: NameRequestI): boolean => {
     } else {
       // applicants is null/undefined
     }
-
+    console.log('Completed')
```
severinbeauvais, line 365: Remove.

src/App.vue
```diff
@@ -156,18 +164,19 @@ export default class App extends Vue {
...
-    const { nrId } = newRequestModule
-    const componentName = newRequestModule.displayedComponent
+    const { getNrId } = this
+    const componentName = this.getDisplayedComponent
     // Only do this for New NRs!!!
-    if (nrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
+    if (getNrId && ['SubmissionTabs'].indexOf(componentName) > -1) {
       // Cancel the NR using the rollback endpoint if we were processing a NEW NR
-      await newRequestModule.rollbackNameRequest({ nrId, action: RollbackActions.CANCEL })
+      await this.rollbackNameRequest({ getNrId, action: RollbackActions.CANCEL })
```
severinbeauvais, line 179: Check other places where an object is passed to a method... this one is broken because the method expects the first object property to be named `nrId`.

severinbeauvais: I fixed this one but there may be others.

src/store/getters.ts
```diff
@@ -76,6 +77,16 @@ export const getJurisdiction = (state: StateIF): string => {
   return state.stateModel.newRequestModel.request_jurisdiction_cd
 }
 
+export const getJurisdictionText = (state: StateIF): string => {
```
severinbeauvais, line 72: This is not really a getter. Should be in an external mixin (assuming said mixin can use a `@Getter` to get the location).

src/interfaces/models.ts
severinbeauvais: I updated the types for amlBlurbs and chgBlurbs but to me it still doesn't look right. Have a look.

src/store/actions.ts
severinbeauvais: I tagged this with `// *** TODO`.

src/mixins/action-mixin.ts
severinbeauvais: Obsolete comment. It is, in fact, used. And I uncommented the code accordingly.

src/mixins/api-error-mixin.ts
severinbeauvais: This didn't work ("cannot get components of undefined").
This file is currently named "~api-error-mixin.vue" to denote that it's not being used (and it's commented out in the index.ts).

src/components/common/applicant-info-1.vue
severinbeauvais: So I've decided that I'll continue to attach these "code tables" (static data) to the Vue instance, so that they don't get imported multiple times (in different components).
I'm still working on cleaning this up...

src/components/new-request/stats.vue
severinbeauvais: No. Nullish coalesce operator is used intentionally so that 0 is displayed (or - if empty).

PR #342, opened 2021-03-16
Re-enabled quick search

cameron-freshworks:
*Issue #:* /bcgov/entity#6581

*Description of changes:*
* Re enabled quick search.
* Misc Payment updates.


src/components/common/request-details.vue
```diff
@@ -67,28 +67,20 @@
...
+import { Getter } from 'vuex-class'
 
 @Component({})
 export default class RequestDetails extends Vue {
-  @Prop(Object)
-  readonly applicant: ApplicantI
-
-  @Prop(Array)
-  readonly nameChoices: {
+  @Prop() applicant: ApplicantI
```
severinbeauvais, line 76: Please restore the "readonly" prefix on the props.

src/store/actions.ts
```diff
@@ -886,8 +886,7 @@ export const startQuickSearch = async ({ commit, getters }) => {
       .replace(/¢/g, 'C')
       .replace(/(`|~|!|\||\(|\)|\[|\]|\{|\}|:|"|\^|#|%|\?|,)/g, '')
 
-    // TODO: Refactor into metho that calls an action
-    await this.getQuickSearch({ 'exactMatch': exactMatchName, 'synonymMatch': synonymsName })
+    await getQuickSearch({ commit, getters }, { 'exactMatch': exactMatchName, 'synonymMatch': synonymsName })
```
severinbeauvais, line 889: Are you sure we should keep methods that make network calls in the store? I'm a bit uncomfortable with that as it might be part of the async issues we've seen in Prod. I'd be more comfortable if store actions didn't call async things.

src/store/actions.ts
```diff
@@ -998,25 +996,17 @@ export const setAddressSuggestions: ActionIF = ({ commit }, addressSuggestions:
   commit('mutateAddressSuggestions', addressSuggestions)
 }
 
-// TODO: Not a real action
-export const fetchCorpNum = async ({ getters }, corpNum: string): Promise<any> => {
+// TODO: Not a real action?
+export const corpNumRequest = async ({ getters }, corpNum: string) => {
```
severinbeauvais, line 1000: I had renamed this to convey that it was a network action (fetch). Do you really prefer this name?
(Or was there a merge issue that knocked out my changes?)

src/store/actions.ts
```diff
@@ -1053,31 +1051,33 @@ export const fetchMRASProfile = async ({ commit, getters }): Promise<any> => {
...
+        request = await getNameRequest(getters.getNrId)
+        if (request?.stateCd === NrState.CANCELLED) {
+          await setActiveComponent('Timeout')
+          return
+        }
       }
+      request = await putNameReservation(getters.getNrId)
     }
-    request = await putNameReservation(getters.getNrId)
+    if (request) await paymentModule.togglePaymentModal(true)
```
severinbeauvais, line 1079: Is error handling (if the PUT failed) in putNameReservation() ?

src/store/actions.ts
severinbeauvais: I don't see a similar rename of the code that calls this method so I assume this rename is a merge snafu. Please double check this method's name and the places that call it.

severinbeauvais (approved):
A few comments. Looks like some of my changes were reverted. But feel free to merge if you feel like you're on top of things.

src/components/common/request-details.vue
cameron-freshworks: Oops, will do!

src/store/actions.ts
cameron-freshworks: I'm with you 100%.
I do have an urgency to get the app back to a functional state before refactoring too too much though. 

If you like, since we are still waiting for a discussion with Thor/Kial on what to do with nr creation/payment, my next commit can focus on this exact type of refactoring?

src/store/actions.ts
cameron-freshworks: kk will do

cameron-freshworks: Fixed

src/store/actions.ts
cameron-freshworks: Yes, turns out it is!

src/store/actions.ts
severinbeauvais: I'm content with quick merges at the moment, even if the code is partly broken or commented out or mid-refactor. As long as we're never far off a working codebase.

I guess the old code had these network calls in the store already, so ^^ isn't any worse.

src/store/actions.ts
severinbeauvais: General soapboxing: This is not a great architecture because, after all this time in the code, we still don't know how/where/when errors are handled. A better architecture would have small, concise network access methods/services and let the "logic code" (middle-tier, if UI code is upper tier and network code is lower tier) handle, well, the logic, including error handling.

There is value in having the network functions report their own errors -- as long as it's done consistently (which it isn't). But I still think low-level functions should not have middle- or high-level logic.

This can be something we clean up... later. Or as we go, if something's really bugging us. I dunno.

End rant.

src/store/actions.ts
severinbeauvais: Ah, reverted. That's fine.

src/store/actions.ts
cameron-freshworks: 😆 I think the rants are the only thing keeping us sane. 
Please, rant away!

And yes to ALL of the above.

src/store/actions.ts
cameron-freshworks: Already working on it!

PR #343, opened 2021-03-16
fix to review date logic

jordiwes:
*Issue #:* /bcgov/entity#6812

*Description of changes:*

Update logic to include submitted date instead of just today's date

### DIFF
--- a/client/src/components/existing-request/existing-request-display.vue
+++ b/client/src/components/existing-request/existing-request-display.vue
@@ -310,15 +310,20 @@ export default class ExistingRequestDisplay extends Mixins(
 
   private get reviewDate () {
     if (this.nr.waiting_time) {
-      let waitingTime = this.nr.waiting_time
+      let queueTime = this.nr.waiting_time
+      let submittedDate = Moment(this.nr.submittedDate)
+      let todaysDate = Moment()
+      // get number of days since the nr was submitted
+      let diffDays = todaysDate.diff(submittedDate, 'days')
+      let waitingTime = queueTime - diffDays + 1
+
       // make sure minimum waiting time is 1
       if (waitingTime < 1) {
         waitingTime = 1
       }
-      let reviewDate = new Date()
-      // add the number of days to the current date to get the review date
-      reviewDate.setDate(reviewDate.getDate() + waitingTime)
-      return Moment(reviewDate).tz('America/Vancouver')
+      // add the waiting time dats to the date submitted
+      let waitingDate = submittedDate.add(waitingTime, 'days')
+      return waitingDate.tz('America/Vancouver')
         .format('MMMM D[,] YYYY') + ' (' + this.nr.waiting_time + ' days)'
     }
     return ''
--- a/client/src/components/existing-request/existing-request-display.vue
+++ b/client/src/components/existing-request/existing-request-display.vue
@@ -312,15 +312,10 @@ export default class ExistingRequestDisplay extends Mixins(
     if (this.nr.waiting_time) {
       let queueTime = this.nr.waiting_time
       let submittedDate = Moment(this.nr.submittedDate)
-      let todaysDate = Moment()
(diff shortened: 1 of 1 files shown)


client/src/components/existing-request/existing-request-display.vue
```diff
@@ -310,15 +310,20 @@ export default class ExistingRequestDisplay extends Mixins(
 
   private get reviewDate () {
     if (this.nr.waiting_time) {
-      let waitingTime = this.nr.waiting_time
+      let queueTime = this.nr.waiting_time
+      let submittedDate = Moment(this.nr.submittedDate)
+      let todaysDate = Moment()
+      // get number of days since the nr was submitted
+      let diffDays = todaysDate.diff(submittedDate, 'days')
+      let waitingTime = queueTime - diffDays + 1
```
thorwolpert, line 318: this makes it a day long
let waitingTime = queueTime - diffDays + 1

let waitingTime = Math.max(queueTime - diffDays, 1)
same logic then as the wait time
and lines 314-317 can be removed

thorwolpert (commented):
changes

client/src/components/existing-request/existing-request-display.vue
```diff
@@ -310,15 +310,20 @@ export default class ExistingRequestDisplay extends Mixins(
 
   private get reviewDate () {
     if (this.nr.waiting_time) {
-      let waitingTime = this.nr.waiting_time
+      let queueTime = this.nr.waiting_time
+      let submittedDate = Moment(this.nr.submittedDate)
+      let todaysDate = Moment()
+      // get number of days since the nr was submitted
+      let diffDays = todaysDate.diff(submittedDate, 'days')
```
severinbeauvais, line 317: As Moment is large (and we don't REALLY need it nowadays), I want to get away from Moment actually. (And I've updated most of the code in NR Release 2.0 branch.)

See if src/components/mixins/date-mixin.ts::daysFromToday() will work for you.

image

client/src/components/existing-request/existing-request-display.vue
```diff
@@ -310,15 +310,20 @@ export default class ExistingRequestDisplay extends Mixins(
...
       if (waitingTime < 1) {
         waitingTime = 1
       }
-      let reviewDate = new Date()
-      // add the number of days to the current date to get the review date
-      reviewDate.setDate(reviewDate.getDate() + waitingTime)
-      return Moment(reviewDate).tz('America/Vancouver')
+      // add the waiting time dats to the date submitted
+      let waitingDate = submittedDate.add(waitingTime, 'days')
+      return waitingDate.tz('America/Vancouver')
```
severinbeauvais, line 326: I'm OK with Moment here because this branch doesn't have my new data format code ... It's just that I don't want to see MORE use of Moment. Here, Moment was already used.

client/src/components/existing-request/existing-request-display.vue
severinbeauvais: One problem with dates/times -- that Moment doesn't solve but my new code does -- is that the browser's date/time could be spoofed to try to get different UI/API behaviour. "Today" and "now" needs to be in Pacific time regardless of what the browser thinks. Moment solves the tz issue but doesn't solve this issue because it assumes the browser's UTC time is correct.

Some of my new code, if you want to have a look at how to do it with plain old TS/JS: https://github.com/bcgov/business-edit-ui/blob/master/src/mixins/date-mixin.ts

jordiwes: I changed the code to use daysFromToday

client/src/components/existing-request/existing-request-display.vue
jordiwes: Done

severinbeauvais (approved):

thorwolpert (approved):

PR #344, opened 2021-03-17
Fixed Name Analysis

cameron-freshworks:
*Issue #:* /bcgov/entity#6581

*Description of changes:*

* Fixed regular and Xpro analysis. 
* Left default flags to `false` but did test locally.

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -1,8 +1,8 @@
 import {
-  AnalysisJSONI,
+  AnalysisJSONI, ConditionalReqI, ConsentConflictI,
   ConversionTypesI, DraftReqI, IssueI,
   LocationT,
-  NameChoicesIF, RequestActionsI, RequestNameI, RequestOrConsentIF,
+  NameChoicesIF, NameDesignationI, RequestActionsI, RequestNameI, RequestOrConsentIF, ReservedReqI,
   StateIF,
   StatsI,
   SubmissionTypeT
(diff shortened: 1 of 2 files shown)


src/store/getters.ts
```diff
@@ -752,6 +756,62 @@ export const getPriorityRequest = (state: StateIF): boolean => {
   return state.stateModel.newRequestModel.priorityRequest
 }
 
+export const getDesignationObject = (state: StateIF): any => {
+  if (getEntityTypeCd(state) && this.$designations[getEntityTypeCd(state)]) {
+    return this.$designations[getEntityTypeCd(state)]
+  }
+  return ''
```
severinbeauvais, line 763: I think this should return null or an empty object instead.
(Normally this method will return an object, so when it fails ... ?)

severinbeauvais (approved):
Small comment. Can be fixed in next PR if you like.

src/store/getters.ts
cameron-freshworks: Good point, this will only every return `''` if there is no entity_type_cd selected or the `$designations` are unavailable.
That being said, this empty string here was probably there way of saying return false, since this method is used as a conditional in `getSplitNameDesignation` which in turn will handle that falsy value. 

Probably why they went with an empty string over an empty object, cuz i don't believe an empty object is falsy.

cameron-freshworks:
> Small comment. Can be fixed in next PR if you like.

K, sounds good!

src/store/getters.ts
severinbeauvais: You're right -- an empty object is not falsy. But null is :)

PR #345, opened 2021-03-17
Extended store with paymentModule

cameron-freshworks:
*Issue #:* /bcgov/entity#6581

*Description of changes:*

* Extends store with existing payment module.
* Refactors references to payment module.
* Fixed the Setters for Applicant info.

### DIFF
--- a/src/components/dialogs/cancel.vue
+++ b/src/components/dialogs/cancel.vue
@@ -39,14 +39,13 @@
 import { Component, Mixins } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
 
-import PaymentModule from '@/modules/payment'
 import { NameRequestMixin, PaymentMixin } from '@/mixins'
 import { sleep } from '@/plugins'
 
 import { NrAction } from '@/enums'
 import { ApplicantI } from '@/interfaces'
 import { ActionBindingIF } from '@/interfaces/store-interfaces'
-import * as PaymentTypes from '@/modules/payment/store/types'
+import { UPGRADE_MODAL_IS_VISIBLE } from '@/modules/payment/store/types'
 
 @Component({})
 export default class CancelDialog extends Mixins(NameRequestMixin, PaymentMixin) {
@@ -65,7 +64,7 @@ export default class CancelDialog extends Mixins(NameRequestMixin, PaymentMixin)
   }
 
   private get isVisible (): boolean {
-    return this.$store.getters['cancelModalIsVisible']
+    return this.$store.getters[UPGRADE_MODAL_IS_VISIBLE]
   }
 
   async showModal (): Promise<void> {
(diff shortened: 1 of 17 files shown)


src/components/dialogs/cancel.vue
```diff
@@ -64,11 +65,11 @@ export default class CancelDialog extends Mixins(NameRequestMixin, PaymentMixin)
   }
 
   private get isVisible (): boolean {
-    return PaymentModule[PaymentTypes.CANCEL_MODAL_IS_VISIBLE]
+    return this.$store.getters['cancelModalIsVisible']
```
severinbeauvais, line 68: So this is what you were talking about.

It feels too "low level" or detailed, but if it helps for now, ok.

Can you (re)use a type for the property name?

src/modules/payment/index.ts
```diff
@@ -1,15 +1,10 @@
...
-import { getVuexStore } from '@/store'
+import { Module, VuexModule } from 'vuex-module-decorators'
 import paymentStore from './store'
 
-const store = getVuexStore()
-
-@Module({ dynamic: true, namespaced: false, store, name: 'paymentModule' })
-export class PaymentModule extends VuexModule {
+@Module
+export default class PaymentModule extends VuexModule {
```
severinbeauvais, line 5: I like it.

src/store/actions.ts
```diff
@@ -1080,7 +1080,7 @@ export const submit: any = async ({ commit, getters }): Promise<any> => {
       }
       request = await putNameReservation(getters.getNrId)
     }
-    if (request) await paymentModule.togglePaymentModal(true)
+    if (request) await dispatch('togglePaymentModal', true)
```
severinbeauvais, line 1083: OK, tell me the different between dispatch and a commit or action?

src/store/index.ts
```diff
@@ -22,6 +23,9 @@ export function getVuexStore () {
   Vue.use(Vuex)
 
   const store = new Vuex.Store<any>({
+    modules: {
+      payment: PaymentModule
```
severinbeauvais, line 27: Oh, gotcha. Looks great!

severinbeauvais (approved):
Looks very nice so far. Way cleaner. Great job!

src/components/dialogs/cancel.vue
cameron-freshworks: It does for sure. 

I bet we can, good idea!

src/store/actions.ts
cameron-freshworks: So, from what i've read..
-Dispatch: To call an action ( after the next tick ) - Asynchronous
- Commit: When you are ready to commit a value to store. - Synchronous 

That being said. We have these compound methods that are doing both... 
I'm not entirely sure they are 'ACTIONS' but it is probably partially true.

severinbeauvais: Why did you choose dispatch here?

cameron-freshworks: Because `togglePaymentModal` is an action within the payment module.

severinbeauvais: Oh, so that's how you call an action in another module? Gotcha.

severinbeauvais:
Let's merge this soon as it's getting difficult to see new changes vs previous changes.

PR #346, opened 2021-03-17
cancel upgrade

kialj876:

*Issue #:* bcgov/entity#6507

*Description of changes:*
- see ticket
- backend dependency

### DIFF
--- a/client/src/components/existing-request/existing-request-display.vue
+++ b/client/src/components/existing-request/existing-request-display.vue
@@ -667,6 +667,19 @@ export default class ExistingRequestDisplay extends Mixins(
     }
   }
 
+  private cancelledUpgrade (status: string, payments: any): string {
+    let paymentId = null
+    if (status === 'PAYMENT_CANCELLED') {
+      for (let i = 0; i < payments.length; i++) {
+        if (payments[i].action === 'UPGRADE' && payments[i].sbcPayment.statusCode === 'CREATED') {
+          paymentId = payments[i].sbcPayment.id
+          break
+        }
+      }
+    }
+    return paymentId
+  }
+
   private get isVisible () {
     const componentName = newReqModule.displayedComponent
     return (componentName === 'ExistingRequestDisplay')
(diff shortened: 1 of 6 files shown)


client/src/components/existing-request/existing-request-display.vue
```diff
@@ -667,6 +667,19 @@ export default class ExistingRequestDisplay extends Mixins(
     }
   }
 
+  private cancelledUpgrade (status: string, payments: any): string {
+    let paymentId = null
+    if (status === 'PAYMENT_CANCELLED') {
```
severinbeauvais, line 672: Are you sure the status is this string exactly? It comes from the API right?

It would be nice to use a type/enum for this, so that we have a single place where the API codes (or URL query parameters, in this case) are declared. On the other hand, if this is the only place this code is used then that's overkill...

client/src/store/new-request-module.ts
```diff
@@ -45,7 +45,7 @@ import errorModule from '@/modules/error'
 import { ErrorI } from '@/modules/error/store/actions'
 import { NrAction, NrState, RollbackActions } from '@/enums'
 import { featureFlags } from '@/plugins/featureFlags'
-import { OK, BAD_REQUEST, NOT_FOUND, SERVICE_UNAVAILABLE } from 'http-status-codes'
+import { OK, BAD_REQUEST, NOT_FOUND, SERVICE_UNAVAILABLE, CREATED } from 'http-status-codes'
```
severinbeauvais, line 48: 👍

severinbeauvais (approved):
I have 2 requests for you:
1. please update the app patch version in package.json and package-lock.json (to make it easy to tell what's deployed)
2. please make this same change to the Namerequest release-2.0 branch (and PR there too)

I've pre-emptively approved so you can proceed when you are happy with this PR.

client/src/components/existing-request/existing-request-display.vue
kialj876: Yeah I was thinking that too haha. It is that string exactly and we don't use it anywhere else

kialj876:
> I have 2 requests for you:
> 
> 1. please update the app patch version in package.json and package-lock.json (to make it easy to tell what's deployed)
> 2. please make this same change to the Namerequest release-2.0 branch (and PR there too)
> 
> I've pre-emptively approved so you can proceed when you are happy with this PR.

will do this later tonight

client/src/components/existing-request/existing-request-display.vue
severinbeauvais: OK, it _should_ be an enum to be explicit that this is a code we receive from an API (pay)...

But don't change this now unless you're already updating that code.

PR #348, opened 2021-03-21
Ported in date and cancel changes from master branch

severinbeauvais:
*Issue #:* /bcgov/entity#6812 and /bcgov/entity#6507

*Description of changes:*
- ported in date logic/calc changes from master
- ported in cancel upgrade changes from master
- updated app version to 2.0.7

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "2.0.6",
+  "version": "2.0.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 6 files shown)


cameron-freshworks (approved):
LGTM. 

Should we be concerned we have payment related requests outside the payment module?
Or, is long term plan to port that whole module over in the main store perhaps?

severinbeauvais:
> Should we be concerned we have payment related requests outside the payment module?
> Or, is long term plan to port that whole module over in the main store perhaps?

Do you mean the code in `actions.ts` ? As you've commented in several places, some of these functions are not actions. As such, I'd love to see them better organized.

I see more complexity here in having multiple modules instead of just one. So, yes, I'm favour of coalescing these into a single store module. But non-actions should still be outside the store IMHO.

PR #355, opened 2021-03-26
Validations fix and additional format check

cameron-freshworks:
*Issue #:* /bcgov/entity#6944

*Description of changes:*

* Fixes email / phone validations
* Applies additional format check to assist* the api in its validations.

* The reason for the phone format check, is the back end only seems to validate the phone number if it is in phone format.
* If a random letter or number is passed in that field instead, it was returning the NR record without a phone match.

*Issue #:* /bcgov/entity#6942

*Description of changes:*

* Fix for editing an approved name where the NR is NOT checked out because the Name Request info itself is not editable.

* Fixing the scope references in the mras actions, fixed the mras profile retrieval. 

### DIFF
--- a/src/components/existing-request/existing-request-search.vue
+++ b/src/components/existing-request/existing-request-search.vue
@@ -66,7 +66,7 @@
                             filled
                             id="phone-number-text-field"
                             label="Applicant's Phone Number"
-                            hint="Example: 555-555-5555"
+                            hint="Example: 555-555-5555 or 555 555 5555"
                             persistent-hint
                             validate-on-blur
               />
@@ -149,21 +149,17 @@ export default class ExistingRequestSearch extends Vue {
     }
   }
 
-  private emailRules = [
-    v => (!!v || !!this.getExistingRequestSearch?.phoneNumber) || 'Please enter either the phone or the email',
-    v => !!this.getExistingRequestSearch?.phoneNumber || (!!v && /.+@.+\..+/.test(v)) ||
-      'Please be sure to enter a valid email'
-  ]
   private nrRules = [
     v => this.NR_REGEX.test(v) || 'Please enter a valid NR number'
   ]
-  private phoneRules = [
-    v => (!v || v.length <= 30) || 'Cannot exceed 30 characters'
-  ]
+  private emailRules = []
+  private phoneRules = []
 
   private async handleSubmit (): Promise<void> {
+    this.applyRules()
     this.$refs.existingNrForm.validate()
     await this.$nextTick()
+
     if (this.isValid) {
       await this.findNameRequest(null)
       if (this.getNr?.failed) {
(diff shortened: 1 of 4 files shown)


src/store/actions.ts
```diff
@@ -443,6 +443,9 @@ export const checkoutNameRequest: ActionIF = async ({ commit, getters }): Promis
 
 export const checkinNameRequest = async ({ getters }): Promise<boolean> => {
   try {
+    // Return if approved Name Request because it was never checked out
+    if (getters.getNrState === NameState.APPROVED) return true
```
severinbeauvais, line 447: Does an approved NR get checked out if I edit it?

src/components/existing-request/existing-request-search.vue
```diff
@@ -149,36 +149,47 @@ export default class ExistingRequestSearch extends Vue {
...
     }
   }
 
+  /** Apply the field rules to request search inputs. */
+  private applyRules (): void {
+    const { emailAddress, phoneNumber } = this.getExistingRequestSearch
+
+    this.phoneRules = (phoneNumber || !emailAddress)
+      ? [
+        v => !!v || 'Please enter a phone number',
```
severinbeauvais, line 179: Isn't this rule redundant, since we are in the if block where `phoneNumber` isn't falsy?

src/components/existing-request/existing-request-search.vue
```diff
@@ -66,7 +66,7 @@
                             filled
                             id="phone-number-text-field"
                             label="Applicant's Phone Number"
-                            hint="Example: 555-555-5555"
+                            hint="Example: 555-555-5555 or 555 555 5555"
```
severinbeauvais, line 69: Check with a PO whether this is OK. I've been see people just typing in `7786983639` and that doesn't validate any more (which may be correct -- I'm not saying it isn't -- just check to approve this change).

src/store/actions.ts
cameron-freshworks: It doesn't actually, just learned that as well. Seems like thats because you are unable to edit anything beyond applicant info.

src/components/existing-request/existing-request-search.vue
severinbeauvais: Oddly, this seems to work -- I can get both validation errors if I enter neither and click the button, which I think is correct.
I think it's odd that this works because this logic is working with 2 sets of variables: `v` in the validation functions which is injected by each component itself, and phoneNumber/emailAddress which comes from the store. I'm thinking there's still an async issue here (though it's hidden and so maybe we don't care).
Thoughts?

cameron-freshworks: Partly true. In the event the user hasn't entered either an email or phone this will trigger for both fields.

severinbeauvais (approved):
Couple of questions, but bottom line, this fix works.

src/store/actions.ts
severinbeauvais: Should it get checked out, to prevent concurrent changes? Or does that cause greater problems at this time?

Would you mind putting in a more details comment to explain the current thinking? A future developer would probably love to know the business/technical logic/decision behind this.

src/components/existing-request/existing-request-search.vue
cameron-freshworks: For sure, this is a really good point.
I think my main goal here was to protect what appears to be broken logic in the api, but we can run this by Linda.
I could also adjust the regex to allow 10 straight characters.

src/components/existing-request/existing-request-search.vue
severinbeauvais: Yup, yup. OK then.

src/components/existing-request/existing-request-search.vue
severinbeauvais: OK, I'll leave this with you to determine and adjust as needed.

This all looks like the right things to do. Well done. GTG

src/store/actions.ts
cameron-freshworks: That is a question i don't think i can answer. 
My understanding of the WHY on this whole check in / check out process is pretty high level. 
...To prevent changes happening on client side / staff side at the same time i imagine, although i think Thor already mentioned a mechanism that would prevent that. So i'm really not sure 🤔 

I will certainly expand the comment though.

severinbeauvais: Looks good.

src/components/existing-request/existing-request-search.vue
severinbeauvais: ^^^ it was decided to not do phone number validation in the UI.

There is a ticket to fix phone number validation in the namex API.

cameron-freshworks: Yes! Sorry, ticket #6950

PR #361, opened 2021-03-29
- moved KC init earlier in start sequence

severinbeauvais:
- fixed use of enums in action buttons
- fixed missing action text

*Issue #:* /bcgov/entity###

*Description of changes:*

### DIFF
--- a/src/components/existing-request/existing-request-display.vue
+++ b/src/components/existing-request/existing-request-display.vue
@@ -123,10 +123,12 @@
                   <v-col cols="12" v-if="isNotPaid">
                     <v-btn block
                            class="button button-blue"
-                           @click="handleButtonClick('RETRY_PAYMENT')">Retry Payment</v-btn>
+                           @click="handleButtonClick(NrAction.RETRY_PAYMENT)"
+                    >{{ actionText(NrAction.RETRY_PAYMENT) }}</v-btn>
                     <v-btn block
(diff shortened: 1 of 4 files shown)


src/components/existing-request/existing-request-display.vue
```diff
@@ -123,10 +123,12 @@
                   <v-col cols="12" v-if="isNotPaid">
                     <v-btn block
                            class="button button-blue"
-                           @click="handleButtonClick('RETRY_PAYMENT')">Retry Payment</v-btn>
```
severinbeauvais, line 126: This was a hack to display just 2 buttons when the NR is not paid, and it was even more of a hack due to hard-coded actions and action texts. It's a little less hacky now... :/

src/main.ts
```diff
@@ -88,16 +82,22 @@ async function startVue () {
     })
   }
 
+  // Initialize Keyloak Service
```
severinbeauvais, line 85: I moved Sentry first, so it can capture any following errors.
I moved KC second, so that maybe it finishes initializing while HJ and LD init. (see 5970 for details)

severinbeauvais: > I moved Sentry first, so it can capture all subsequent errors.
> I moved KC second, so that maybe it finishes initializing while HJ and LD init. (see 5970 for details)

cameron-freshworks (approved):

PR #367, opened 2021-04-09
7089 Edit rejected state bug

cameron-freshworks:
*Issue #:* /bcgov/entity#7089

*Description of changes:*

* Skip the checkin functionality for Rejected state NR's because they are never checked out.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "2.0.19",
+  "version": "2.0.20",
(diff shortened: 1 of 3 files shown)


severinbeauvais (approved):
Thanks for the quick fix.

Could you please update the app versions so we can tell when your fix is deployed? Thx!

cameron-freshworks:
> Thanks for the quick fix.
> 
> Could you please update the app versions so we can tell when your fix is deployed? Thx!

o ya, for sure!

PR #369, opened 2021-04-16
renew nr updates

kialj876:

*Issue #:* /bcgov/entity#6552

*Description of changes:*
- see ticket for details
- updated based on new renew designs
- fee summary restructured to make styling easier (currently was not matching designs)

### DIFF
--- a/src/components/dialogs/reapply.vue
+++ b/src/components/dialogs/reapply.vue
@@ -1,10 +1,19 @@
 <template>
-  <v-dialog max-width="40%" :value="isVisible" persistent>
+  <v-dialog max-width="35rem" :value="isVisible" persistent>
     <v-card>
 
       <v-card-title class="d-flex justify-space-between">
-        <div>Re-apply for Name</div>
+        <div>Renew Name Request</div>
+        <v-btn icon large class="dialog-close float-right pa-0" @click="hideModal()">
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
       </v-card-title>
+
+      <v-row no-gutters>
+        <div class="copy-normal pt-8">
+          Renew your Name Request for an additional 56 days from the current expiry date.
+        </div>
+      </v-row>
 
       <v-card-text class="copy-normal">
         <fee-summary
@@ -13,13 +22,16 @@
         />
       </v-card-text>
 
-      <v-card-actions>
-        <v-spacer></v-spacer>
-        <v-btn @click="confirmPayment()" id="payment-pay-btn" class="primary" text :loading="isLoadingPayment">
-          Accept
-        </v-btn>
-        <v-btn @click="hideModal()" id="payment-close-btn" class="normal" text>Cancel</v-btn>
-      </v-card-actions>
+      <v-row no-gutters justify="center" class="pt-8">
+        <v-col cols="auto">
+          <v-btn @click="confirmPayment()" id="payment-pay-btn" class="primary" text :loading="isLoadingPayment">
+            Continue to Payment
+          </v-btn>
+        </v-col>
+        <v-col cols="auto" class="pl-3">
+          <v-btn @click="hideModal()" id="payment-close-btn" class="button button-blue" text>Cancel</v-btn>
+        </v-col>
+      </v-row>
 
     </v-card>
   </v-dialog>
--- a/src/components/dialogs/reapply.vue
+++ b/src/components/dialogs/reapply.vue
@@ -9,11 +9,9 @@
         </v-btn>
       </v-card-title>
 
-      <v-row no-gutters>
-        <div class="copy-normal pt-8">
-          Renew your Name Request for an additional 56 days from the current expiry date.
-        </div>
-      </v-row>
+      <div class="copy-normal pt-8">
+        Renew your Name Request for an additional 56 days from the current expiry date.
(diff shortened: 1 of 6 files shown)


src/components/dialogs/reapply.vue
```diff
@@ -1,25 +1,37 @@
 <template>
-  <v-dialog max-width="40%" :value="isVisible" persistent>
+  <v-dialog max-width="35rem" :value="isVisible" persistent>
```
severinbeauvais, line 2: Does this shrink OK if the user has a narrow viewport?

src/components/existing-request/existing-request-display.vue
```diff
@@ -315,7 +316,7 @@ export default class ExistingRequestDisplay extends Mixins(
     let ret: string
     if (this.nr.expirationDate) {
       const date = new Date(this.nr.expirationDate)
-      ret = this.dateToPacificDate(date)
+      ret = this.dateToPacificDateTime(date)
```
severinbeauvais, line 319: Date-mixin methods FTW! 😀

src/components/payment/fee-summary.vue
```diff
@@ -1,39 +1,61 @@
...
-            <li class="container fee-list__item"
-              v-if="lineItem.serviceFees"
-              :key="lineItem.filingTypeCode+'-transaction'"
-              >
-              <div class="fee-list__item-name">Service fee</div>
-              <div class="fee-list__item-value">${{lineItem.serviceFees.toFixed(2)}}</div>
-            </li>
-          </template>
-        </v-slide-y-transition>
+        <v-row no-gutters class="pt-8 pb-4 header">
```
severinbeauvais, line 5: Why use a v-row here with no v-col in it? Would a div work as well?

src/components/payment/fee-summary.vue
```diff
@@ -52,16 +74,24 @@
...
-          <!--<div class="fee-total__currency">CAD</div>-->
-          <div class="fee-total__value">
-            <v-slide-y-reverse-transition name="slide" mode="out-in">
-              <div><b>${{total.toFixed(2)}}</b></div>
-            </v-slide-y-reverse-transition>
-          </div>
-        </div>
+        <v-row no-gutters class="pt-4 pb-8">
+          <v-col cols="8">
+            <v-row no-gutters justify="start">
```
cameron-freshworks, line 79: Just curious, is this additional row required with only one element within?

cameron-freshworks (approved):
generally looks really good! Just a question about the additional rows within the columns and if they are required beyond giving access to vuetify's helper styling. 
Overall though, all good 👍

src/components/dialogs/reapply.vue
kialj876: yes it does (I tested it by narrowing my window - do you have a different way of testing this?)

src/components/payment/fee-summary.vue
kialj876: I thought it was required so I could use the justify "start", but maybe I could do this differently with the v-col. Will check

src/components/payment/fee-summary.vue
kialj876: updated - this one, checking to see if I can change the ones cam questioned below

src/components/dialogs/reapply.vue
severinbeauvais: That's how I test it!

src/components/payment/fee-summary.vue
severinbeauvais: There's probably a Verify helper class for that

kialj876: Alright so I can remove the v-rows for the ones I'm using for the ```justify="start"```, and for the other ones I will add ```style=float: right;``` to the div

severinbeauvais: There's probably a float-right class already somewhere.

kialj876: I got in the habit of always using v-row for every item on the page because I thought it made adjusting things on the page easier to follow, but I guess it's not really needed and just adds more lines

severinbeauvais: Your other option is to keep the v-rows so your code looks consistent, but add single v-cols under them.

kialj876: haha okay added the float-right class in. I'm good with this unless you would prefer the extra row/cols in there

src/components/payment/fee-summary.vue
severinbeauvais: Smells clean now!

(It was a code smell to use v-rows without v-cols, haha.)

src/components/payment/fee-summary.vue
severinbeauvais: I'm good with this.

I'll leave it to Cam to resolve this convo.

severinbeauvais:
Ah sorry, one more thing: please update the app version (so we can check it in various environments).

OK, one itty bitty other thing: wait for Cam to deploy his fix from Dev -> Test before you deploy to Dev. (To keep the builds separate, which will allow to deploy to Prod today maybe, without including your change.)

severinbeauvais (approved):
Approved but see my previous comment before merging.

PR #377, opened 2021-05-03
5730 SASS code to SCSS

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5730

*Description of changes:*


package.json
```diff
@@ -57,10 +57,8 @@
     "sass": "^1.29.0",
     "sass-loader": "^8.0.2",
     "sinon": "^8.1.1",
-    "style-resources-loader": "^1.4.1",
```
severinbeauvais, line 60: Yay, thanks for removing unused imports!

Also, please update the app version (on line 3) so we can tell what version is running when deployed.

vysakh-menon-aot: will do

src/assets/scss/base.scss
```diff
@@ -113,3 +113,350 @@ html {
 .cursor-default {
   cursor: default; /* arrow */
 }
+
+.bg-grey-1 {
```
severinbeauvais, line 117: Are these all being used, or can some of them be deleted?

vysakh-menon-aot: bg-grey-1 is been used. bg-grey-2 to bg-grey-5 not used

src/assets/scss/overrides.scss
```diff
@@ -84,3 +75,207 @@
...
+
+.grey-box-checkbox-button .v-input__slot {
+  position: relative;
+  left: 0px;
+  top: 0px;
+}
+
+.designation-buttons button {
+  margin: 0 !important;
+  padding: 0 !important;
```
severinbeauvais, line 91: Could you delete this and add `ma-0 pa-0` to the button itself?

src/assets/scss/overrides.scss
```diff
@@ -84,3 +75,207 @@
...
+  padding: inherit 28px !important;
+}
+
+.v-btn.normal {
+  color: white !important;
+  background-color: $gray6 !important;
+}
+
+.v-label {
+  color: rgba(73, 80, 87, .95) !important;
```
severinbeauvais, line 121: Theoretically, this will now affect v-labels that use SCSS (where previously it only affected v-labels that used SASS). Have you noticed any changes to the UI?

src/assets/scss/overrides.scss
```diff
@@ -84,3 +75,207 @@
...
+  color: $app-blue !important;
+  background-color: white !important;
+}
+
+.v-btn.warning {
+  color: white !important;
+  background-color: $caution !important;
+}
+
+.v-btn.error {
```
severinbeauvais, line 154: All these (primary, secondary, outliens, warning, error) will now affect more (all) buttons in theory. We need to make sure things looks the same as before.

src/assets/scss/overrides.scss
```diff
@@ -84,3 +75,207 @@
...
+.v-tab.active-tab:hover, .v-tab--active {
+  background-color: white !important;
+  pointer-events: none;
+}
+
+.v-tab:not(.active-tab):hover {
+  background-color: $app-blue !important;
+}
+
+.top-tooltip, .left-tooltip {
```
severinbeauvais, line 187: Same comment as above: all tooltips are now affected by this. We need to make sure things look the same as before.

src/assets/scss/theme.scss
```diff
@@ -59,6 +59,10 @@ $app-green: #2e8540;
 $border: #899eb2;
 $gold: $BCgovGold5;
 
+// Colours that should use standard palette instead:
+$other-bg: #e2e8ee;
+$caution: #bf4803;
```
severinbeauvais, line 64: Since these should use other colours instead, can you please check with Scott what these should be?

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -232,27 +232,42 @@ export default class PickEntityOrConversionDialog extends Vue {
...
+  text-align: left;
+}
+
+.clickable-cell {
+  min-width: 180px;
+  width: 100%;
+  padding: 10px;
+
+  &:hover {
+    background-color: $app-lt-blue;
```
severinbeauvais, line 264: Old colour is "$grey-2" but you have changed it. Is it intentional?

src/components/dialogs/pick-request-type.vue
```diff
@@ -94,8 +94,11 @@ export default class PickRequestTypeDialog extends Vue {
...
 
-<style lang="sass" scoped>
-.clickable-cell:hover
-  background-color: $grey-2
-  cursor: pointer
+<style lang="scss" scoped>
+@import '@/assets/scss/theme.scss';
+
+.clickable-cell:hover {
+  background-color: $app-lt-blue;
```
severinbeauvais, line 101: Is "app-lt-blue" correct?

src/assets/scss/base.scss
severinbeauvais: OK, please clean up (delete) what you can.
(But don't delete SCSS variables... we might still use those.)

src/assets/scss/base.scss
```diff
@@ -113,3 +113,350 @@ html {
...
+.colour-text {
+  color: $text !important;
+}
+
+.colour-dk-text {
+  color: $dk-text !important;
+}
+
+.colour-white {
+  color: white !important;
```
severinbeauvais, line 154: This one doesn't look very useful (since it's assigning a colour by name, not variable, which can be scoped instead of global).

As it's used in only 1 place, can you please move this to where it's used?

src/components/lower-info-area/name-build-sub-components/consent-words.vue
```diff
@@ -40,19 +40,25 @@ export default class ConsentWords extends Vue {
...
-      padding-left: 0
+  .consent-words-list {
+    background: inherit;
+
+    .consent-words-list-item {
+      color: $text;
+      font-size: 14px;
+      letter-spacing: 0;
+      line-height: 22px;
+      min-height: 0!important;
```
severinbeauvais, line 59: Should have space after "0".

severinbeauvais: Check other files for same thing.

src/components/lower-info-area/name-build-sub-components/unique-names.vue
```diff
@@ -41,9 +41,11 @@ import { Component, Vue } from 'vue-property-decorator'
...
-<style scoped lang="sass">
-  #unique-names
-    p
-      font-size: 14px
-      margin-bottom: 0
+<style scoped lang="scss">
+  #unique-names {
+    p {
+      font-size: 14px;
+      margin-bottom: 0;
```
severinbeauvais, line 48: Please remove the margin spec and add "mb-0" to the element's class in the template -- it is easier to have layout specs like this in the template, in case we need to move things around / fix things.

My rule of thumb:
- layout rules in the template
- local styles in the scoped block
- global styles in the assets

src/components/dialogs/pick-entity-or-conversion.vue
vysakh-menon-aot: yes. it was the same color code

severinbeauvais (commented):
This is awesome (and much needed) cleanup - thanks!

Please compare NR UI with your changes against NR UI without your changes (eg, Dev), looking at as many pages and scenarios as possible for display differences. (Or maybe just the pages directly impacted by SASS -> SCSS.) I wouldn't be surprised to find a few things that need to be updated.

If you keep a list of the pages/scenarios you have personally verified, please add that as objective evidence, which will help QA as well.

src/components/dialogs/pick-request-type.vue
severinbeauvais: same as above

src/assets/scss/overrides.scss
vysakh-menon-aot: Since it was imported in main.ts this could have no impact. right?

severinbeauvais: You are right! I confirmed this in NR UI Dev.

image

src/assets/scss/overrides.scss
vysakh-menon-aot: same as above

src/assets/scss/overrides.scss
vysakh-menon-aot: same as above

src/assets/scss/theme.scss
vysakh-menon-aot: Will update once I have response from Scott

severinbeauvais (approved):

severinbeauvais:
@kialj876 Please merge if OK with you!

vysakh-menon-aot:
> @kialj876 Please merge if OK with you!

Can you merge it. I don't have permission

PR #380, opened 2021-05-11
Added safety checks and logging + fixed button styles

severinbeauvais:
*Issue #:* /bcgov/entity#7490 and /bcgov/entity#7507

*Description of changes:*
- added safety checks in POST call
- added extra logging in POST error
- fixed button styles
- updated app version

### DIFF
--- a/src/services/namex.services.ts
+++ b/src/services/namex.services.ts
@@ -369,9 +369,9 @@ export default class NamexServices {
       throw new Error(`Invalid response = ${response}`)
     } catch (err) {
       // extra logging to help find errors
-      err.message && console.log('postNameRequests(), message =', err.message) // eslint-disable-line no-console
-      err.request && console.log('postNameRequests(), request =', err.request) // eslint-disable-line no-console
-      err.response && console.log('postNameRequests(), response =', err.response) // eslint-disable-line no-console
+      err?.message && console.log('postNameRequests(), message =', err.message) // eslint-disable-line no-console
+      err?.request && console.log('postNameRequests(), request =', err.request) // eslint-disable-line no-console
+      err?.response && console.log('postNameRequests(), response =', err.response) // eslint-disable-line no-console
       const msg = await this.handleApiError(err, 'Could not post name requests')
       console.error('postNameRequests() =', msg) // eslint-disable-line no-console
       await errorModule.setAppError({ id: 'post-name-requests-error', error: msg } as ErrorI)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "2.1.2",
```
severinbeauvais, line 3: Yes, 2.1.2 with broken button styles is in Prod.

src/assets/scss/base.scss
```diff
@@ -85,27 +85,6 @@ html {
...
-  border: thin solid $app-blue !important;
-}
-
-// red button: red on white
-.button-red {
-  color: $app-red !important;
-  background-color: transparent !important;
-  border: thin solid $app-red !important;
-}
-
```
severinbeauvais, line 108: This file is loaded before the overrides, so the overrides took effect (ie, normal button styles overrode these).

src/assets/scss/overrides.scss
```diff
@@ -107,9 +107,32 @@
...
+  background-color: transparent !important;
+  border: thin solid $app-blue !important;
+}
+
+// red button: red on white
+.button-red {
+  color: $app-red !important;
+  background-color: transparent !important;
+  border: thin solid $app-red !important;
 }
```
severinbeauvais, line 136: Here, I moved the "special" button rules to this file, and after the normal button rules, so that have effect.

(As you know, CSS rules work in the order they are declared, and/or according to specificity.)

Re: https://www.w3schools.com/css/css_specificity.asp

src/services/namex.services.ts
```diff
@@ -363,6 +368,10 @@ export default class NamexServices {
       }
       throw new Error(`Invalid response = ${response}`)
     } catch (err) {
+      // extra logging to help find errors
+      err.message && console.log('postNameRequests(), message =', err.message) // eslint-disable-line no-console
+      err.request && console.log('postNameRequests(), request =', err.request) // eslint-disable-line no-console
+      err.response && console.log('postNameRequests(), response =', err.response) // eslint-disable-line no-console
```
severinbeauvais, line 374: Changes in this file are to help us track down POST HTTP 500 issues.

severinbeauvais:
> I have one request to log out the payload if the request fails (in your added logging for 500 errors), but this can go ahead without that if needed

Yeah, that's what logging the `err` parts does (for any error).

src/assets/scss/overrides.scss
cameron-freshworks: hahah i have been battling the button styles in my current work as well. 

Thanks for this. I'm not 100% if all of this carries over, but i will use everything out of here i can and remove my local changes, after this PR goes in! 👍

cameron-freshworks (approved):
Looks great! Just had a comment about the button styles.
I'll re-use everything i can from here. Was battling these styles locally as well 😁

PR #381, opened 2021-05-11
Advanced Name Request Search Flow

cameron-freshworks:
*Issue #:* /bcgov/entity#7326

*Description of changes:*

* Implemented the advanced search tab component and flow. This flow includes..
* Search form
* Results table tab
* Results retrieval form and request
* Invalid results tab

* Added current date logic
* Minor style updates
* Moved local getter to store for expanded use

*NOTES:*
* This work still requires the search query to be implemented which is currently back end work. Will implement that aspect in ticket #7325.

* Currently disabled the link to access this flow.
* Also, the means by which this access is granted could be refined beyond whether a user is authenticated or not, down to specific roles perhaps, but that is yet to be implemented.


src/components/advanced-search/advanced-search-dates.vue
```diff
@@ -0,0 +1,139 @@
...
+            v-model="endDateText"
+            :max="currentDateString"
+            elevation="5"
+          >
+          </v-date-picker>
+        </v-col>
+      </v-row>
+      <v-card-actions class="my-0 py-0 float-right">
+          <v-btn id="btn-done" text color="primary" :ripple="false" @click="submitDates()"><strong>OK</strong></v-btn>
+          <v-btn id="btn-cancel" text color="primary" :ripple="false" @click="emitClose()">Cancel</v-btn>
```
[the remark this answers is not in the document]
severinbeauvais: This is the first time I have seen the OK button bolded.
Also, in other places the Cancel button has gray styling (see upgrade.vue).

Do we want to proceed with different styles than elsewhere in this app?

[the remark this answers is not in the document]
cameron-freshworks: Great question. I'm not actually sure about either, this is most likely due to habit of doing so since it was requested in UX feedback prior. 
I don't believe we've standardized this 😁

src/components/advanced-search/advanced-search-dates.vue
```diff
@@ -0,0 +1,139 @@
...
+// Mixins
+import { DateMixin } from '@/mixins'
+
+// Interfaces
+import { StartEndDatesI } from '@/interfaces/models'
+
+@Component({})
+export default class AdvancedSearchDates extends Mixins(DateMixin) {
+  /** Prop to display the dialog. */
+  @Prop() private dialog: boolean
```
severinbeauvais, line 49: Change private to readonly.

cameron-freshworks: kk, will do 👍

src/mixins/date-mixin.ts
```diff
@@ -150,4 +150,22 @@ export class DateMixin extends Vue {
...
   }
+
+  /**
+   * Converts a Date object to a date string (YYYY-MM-DD) in Pacific timezone.
+   * @example "2021-01-01 07:00:00 GMT" -> "2020-12-31"
+   * @example "2021-01-01 08:00:00 GMT" -> "2021-01-01"
+   * @example "2021-01-01 00:00:00 PST" -> "2021-01-01"
+   * @example "2021-01-01 23:59:59 PST" -> "2021-01-01"
+   */
+  dateToDateString (date: Date): string {
```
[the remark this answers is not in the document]
cameron-freshworks: For sure, i'll have a look at `dateToYyyyMmDd()` and see if we can replace with the one you wrote here.

src/components/existing-request/existing-request-search.vue
```diff
@@ -238,6 +268,15 @@ export default class ExistingRequestSearch extends Vue {
...
+  padding: 0 30px !important;
+  font-size: 1rem !important;
+}
+
+#advanced-search-btn {
+  color: $app-blue !important;
+  font-size: .875rem;
+  &:hover {
+    cursor: pointer;
+  }
```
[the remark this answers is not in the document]
cameron-freshworks: Will have a look!

src/components/existing-request/existing-request-search.vue
```diff
@@ -94,6 +102,19 @@
             </v-col>
           </v-row>
 
+          <!-- SIXTH LINE -->
+          <v-row v-if="false" class="mt-3" no-gutters> <!-- Condition to be: getIsAuthenticated -->
+            <v-col class="text-center">
+              <span
```
severinbeauvais, line 108: Why use a span instead of a button?

severinbeauvais (commented):
I've only partly reviewed... will continue later.

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: Do you think all the props should be readonly?

src/mixins/date-mixin.ts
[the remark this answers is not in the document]
cameron-freshworks: Replaced instances of `dateToYyy..` with `dateToDateString` as that is common amongst our other ui's

src/interfaces/models.ts
```diff
@@ -255,3 +255,16 @@ export interface WaitingAddressSearchI {
...
+
+export interface AdvancedSearchI {
+  businessName: string
+  applicantFirstName: string
+  applicantLastName: string
+  startDate: string
+  endDate: string
+}
+
+export interface StartEndDatesI {
```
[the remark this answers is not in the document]
cameron-freshworks: Updated 👍

src/components/existing-request/existing-request-search.vue
cameron-freshworks: Because `V-btn` in name request was rigidly styled.. probably my own doing/blunder. 
I think you improved this in your last PR. I will pull that in and apply the updated styling if possible 👍

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: Great points!

Absolutely. I think the really valuable ones are when you don't close dialogs correctly. As an example you open another dialog without actually closing the initial one, you get this same error.

src/assets/scss/overrides.scss
```diff
@@ -135,8 +135,17 @@
   border: thin solid $app-red !important;
 }
 
+// blue button: blue on white
+.button-text {
+  color: $app-blue !important;
```
severinbeauvais, line 140: Can you use `@extend` to keep the same styles as button-blue and just add what you need?

Example: https://github.com/bcgov/namerequest/blob/e6a86c444d2bc113db98c3b649585485164dd965/src/assets/scss/base.scss#L117

severinbeauvais: This code base just keeps getting better and better! This morning I felt :( about this app, but now I feel :)

src/components/dialogs/advanced-search.vue
```diff
@@ -0,0 +1,263 @@
...
+                  one search criteria and/or reducing your date range.</span>
+            </v-card-text>
+          </v-tab-item>
+
+        </v-tabs-items>
+      </v-tabs>
+
+      <v-card-actions id="advanced-search-actions" class="justify-center pa-0">
+          <v-btn
+            id="adv-search-btn-1"
```
severinbeauvais, line 86: This button's vertical alignment is off just a bit.

image

src/components/advanced-search/advanced-search-dates.vue
```diff
@@ -0,0 +1,139 @@
...
+            class="mt-2"
+            v-model="startDateText"
+            :max="currentDateString"
+            elevation="5"
+          >
+          </v-date-picker>
+        </v-col>
+        <v-col>
+          <span class="title-bold-16" :class="{ 'error-message': validate && !hasEndDate }">Select Start Date:</span>
+          <v-date-picker
```
severinbeauvais, line 18: Oddly, in this date picker, when I click on today's day, the styling is different than any other day. I was confused at first whether today is a valid value. Can this be changed?

image

src/components/dialogs/advanced-search.vue
```diff
@@ -0,0 +1,263 @@
...
+            @click="advSearchBtn1Actions()"
+          >
+            <span class="px-1">
+              <v-icon v-if="!isTabSearchForm">mdi-chevron-left</v-icon>
+              {{ advSearchBtn1 }}
+              <v-icon v-if="isTabSearchForm">mdi-chevron-right</v-icon>
+            </span>
+          </v-btn>
+          <v-btn
+            id="adv-search-btn-2"
```
severinbeauvais, line 100: This button's alignment looks a bit off too, in certain cases:

image

src/components/dialogs/advanced-search.vue
```diff
@@ -0,0 +1,263 @@
...
+
+  /** Prop to provide attachment selector. */
+  @Prop() readonly attach: string
+
+  // Local Properties
+  private advSearchTab = 0
+  private toggleSearchSubmit = false
+  private toggleRetrieveSubmit = false
+  private invalidSearch = false
+  private readonly AdvancedSearchTabs = AdvancedSearchTabs
```
severinbeauvais, line 144: I've been sticking these at the top of the class with comment `// Declarations for template`, if you want to copy that. (Also just readonly is fine.)

src/components/advanced-search/advanced-search-retrieve.vue
```diff
@@ -0,0 +1,170 @@
...
+      ? [
+        v => !!v || 'Please enter an email',
+        v => (/.+@.+\..+/.test(v)) || 'A valid email is required'
+      ]
+      : []
+  }
+
+  /** Clear field level validations. */
+  private clearValidations (): void {
+    this.$refs.advancedSearchRetrieveForm.resetValidation()
```
severinbeauvais, line 108: You may have to await this.

src/components/advanced-search/advanced-search-form.vue
```diff
@@ -0,0 +1,200 @@
...
+
+  // Local model
+  private businessName = ''
+  private applicantFirstName = ''
+  private applicantLastName = ''
+  private startDate = ''
+  private endDate = ''
+
+  /** Validation rules for Applicant last name. */
+  private applicantLastNameRules = []
```
severinbeauvais, line 119: Will this be used?

vysakh-menon-aot (approved):

src/components/dialogs/advanced-search.vue
cameron-freshworks: Hmm i'm having a hard time picking it up. Do you mean it's sitting a little high?

src/components/advanced-search/advanced-search-dates.vue
```diff
@@ -0,0 +1,139 @@
...
+            id="start-date-calendar"
+            class="mt-2"
+            v-model="startDateText"
+            :max="currentDateString"
+            elevation="5"
+          >
+          </v-date-picker>
+        </v-col>
+        <v-col>
+          <span class="title-bold-16" :class="{ 'error-message': validate && !hasEndDate }">Select Start Date:</span>
```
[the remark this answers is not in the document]
cameron-freshworks: Good catch, thank you!

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: hmm let me look into this, i didn't notice before!

src/components/dialogs/advanced-search.vue
cameron-freshworks: hmm... yea not sure here. I can see the alignment is centred.. but with no padding influencing the vertical height that i can tell. Didn't do anything unusual here.

src/components/dialogs/advanced-search.vue
cameron-freshworks: ahh yea! The horizontal alignment spacing looks too big on the back btn for sure.

src/components/advanced-search/advanced-search-table.vue
```diff
@@ -0,0 +1,46 @@
...
+      <template v-slot:item="row" class="advanced-search-results">
+        <tr>
+          <td></td>
+          <td></td>
+          <td></td>
+          <td></td>
+        </tr>
+      </template>
+    </v-data-table>
+    <v-btn @click="mockSelectNR">Mock Select NR</v-btn>
```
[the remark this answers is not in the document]
cameron-freshworks: I agree for sure, the tab heights are the same besides the Retrieve NR tab. This height difference was called out explicitly in designs, as was the sliding transition :)
I have noticed this as well though, will see what i can do to smooth it out. (Without churning too much time 😁 )

src/components/dialogs/advanced-search.vue
cameron-freshworks: For sure, i'm good with that!

src/components/advanced-search/advanced-search-retrieve.vue
cameron-freshworks: Will have a look but i didn't notice any issues here!

src/components/advanced-search/advanced-search-retrieve.vue
```diff
@@ -0,0 +1,170 @@
...
+    this.phoneRules = []
+    this.emailRules = []
+    this.errorMessage = ''
+  }
+
+  /** Apply rules dynamically and await form validation. */
+  private async validateForm (): Promise<void> {
+    this.applyRules()
+    this.$refs.advancedSearchRetrieveForm.validate()
```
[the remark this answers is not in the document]
cameron-freshworks: For sure, i think it would make sense to await it!
Is it 1 for 1 though, would awaiting this === waiting for the next tick in Vue?

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: Yes, they are applied ONLY if the client enters a first name, line 140.

src/components/advanced-search/advanced-search-dates.vue
```diff
@@ -0,0 +1,139 @@
...
+      <v-row>
+        <v-col>
+          <span class="title-bold-16" :class="{ 'error-message': validate && !hasStartDate }">Select Start Date:</span>
+          <v-date-picker
+            id="start-date-calendar"
+            class="mt-2"
+            v-model="startDateText"
+            :max="currentDateString"
+            elevation="5"
+          >
```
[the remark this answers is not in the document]
cameron-freshworks: I'm not up to speed on the Date situation, i just used the code written for the Edit UI with regards to retrieving the current server date. 

Set to store in the same manner in `App.vue` and we just retrieve it.
Screen Shot 2021-05-12 at 8 02 02 AM

Is this an issue in all the apps, or is something done differently here?

cameron-freshworks:
> This is very well written HTML and TS... good job!
> 
> Yes, I looked past the incomplete parts... that can wait until the next PR.
> 
> GTG after comments are addressed/closed/recorded for future.

Wonderful, thanks a lot!

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: Fixed! It is like this in the shared component, however here it was still a light gray opposed to the light blue. 
Updated tho, good eye 👍 

Screen Shot 2021-05-12 at 8 10 33 AM

src/components/dialogs/advanced-search.vue
cameron-freshworks: fixed!

src/components/dialogs/advanced-search.vue
cameron-freshworks: Yea, the more i look, the icon font size is affecting the alignment from what i can tell. Still looking!

severinbeauvais: It looks like the > is centered vertically but Find Name Request is a pixel or 2 high, compared to Cancel. You may need margin on it to fix it.

src/components/advanced-search/advanced-search-dates.vue
severinbeauvais: Much better thanks!

src/components/advanced-search/advanced-search-table.vue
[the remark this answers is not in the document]
severinbeauvais: Yes, I remember the designs. Maybe Scott/Tracey will have an idea after seeing it in action.

I don't recall if the 2 tab sizes transition with vertical expand/shrink, or whether they use the left/right panning like the other tabs? Maybe they're using the left/right transition and it would look better with vertical expand/shrink?

src/components/dialogs/advanced-search.vue
cameron-freshworks: Yea, the icon itself is increasing that inner elements height a bit. I just shrunk the icon a little, looks better 👍

src/components/advanced-search/advanced-search-retrieve.vue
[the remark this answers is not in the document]
severinbeauvais: In lib.dom.d.ts I see this:
```
function validate(bytes: BufferSource): boolean;
```
... It's written in WebAssembly so I'm not going to be able to see its inner workings.

ATM let's stick with what works: the await nextTick after it.

src/components/dialogs/advanced-search.vue
cameron-freshworks: This was updated with the icon size adjustment 👍

src/components/advanced-search/advanced-search-retrieve.vue
severinbeauvais: I believe there's no visible issue because the subsequent code doesn't depend on it, but I highly suspect resetting validation takes a few cycles -- like `validate()`.

I quickly searched the code base and there's actually no indication that `validate` is async, although we know from experience that we need to await the next tick to get the result we want... more in the comment below.

I can't find `resetValidation()` anywhere in the code base. (Are we sure it works? lol) On the assumption that it works like validation, I think it be would safest to add an "await nextTick" after the resetValidation call...

I see this app calls `resetValidation` in a few other places without waiting for it to finish -- again, no visible issue because of the current code flow. So I dunno: stay consistent and risk (another) async issue in this app, or go with a gut feel? What do you think?

src/components/advanced-search/advanced-search-retrieve.vue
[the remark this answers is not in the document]
severinbeauvais: Right now I'm thinking awaiting it won't work and we instead need to await the next tick to give it time to settle. In a true real-time system I'd ask "how long do we need to wait?" but here I think the single tick wait seems enough, right? Erm...

src/components/advanced-search/advanced-search-form.vue
severinbeauvais: Gotcha. I missed that the first time. You are correct :)

src/components/advanced-search/advanced-search-retrieve.vue
cameron-freshworks: I prefer to await as well. I think that would be the prudent approach for defensive coding :) 
Updated!

src/components/advanced-search/advanced-search-retrieve.vue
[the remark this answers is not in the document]
cameron-freshworks: The await appears to work. Updated in the meantime :)

src/components/advanced-search/advanced-search-dates.vue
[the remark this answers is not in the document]
severinbeauvais: Gaah, I totally missed line 11... Here, I _think_ this code is as good as it gets, though it may still confuse a user in a different timezone wondering wht their start/end date doesn't find a NR one day before/after... In other places we've been explicit that our system dates/times are "Pacific time". Is it possible to indicate that here? Maybe check with Scott/Tracey.

Yes, this timezone thing is an issue in all the apps, though I've been picking away at it bit by bit. (And I'm not thrilled about my current solution but we can talk about that separately.)

src/components/advanced-search/advanced-search-table.vue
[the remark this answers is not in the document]
cameron-freshworks: Still looking at ways to smooth this though. It's appears it is the buttons that make it look jumpy 🤔

src/components/advanced-search/advanced-search-dates.vue
severinbeauvais: I just go through it and see if it looks/functions as I, an experienced web user, expect. It's always easier reviewing someone else's code 😁

src/components/advanced-search/advanced-search-retrieve.vue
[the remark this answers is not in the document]
severinbeauvais: Awaiting it works? I can't explain why and that makes me a bit nervous, lol

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: Totally! Makes for a very good review as well, albeit a little more work.

src/components/advanced-search/advanced-search-retrieve.vue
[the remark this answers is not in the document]
cameron-freshworks: It could be because it doesn't need to await in the first place? 🤷  I dunno lol

src/components/advanced-search/advanced-search-dates.vue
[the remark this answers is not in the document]
cameron-freshworks: I think we are explicit to "Pacific time" in this instance as well. When we set the date in these pickers we use this method.

Screen Shot 2021-05-12 at 9 23 58 AM

src/components/advanced-search/advanced-search-retrieve.vue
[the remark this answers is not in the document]
severinbeauvais: You have a good point there. I noticed that `validate` returns a boolean. That can only work if it's not async.
🤷

src/components/advanced-search/advanced-search-dates.vue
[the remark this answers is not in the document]
severinbeauvais: Yes, our system works in Pacific time, but we don't tell the user that on the date picker.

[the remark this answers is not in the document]
cameron-freshworks: ahh i see.

cameron-freshworks:
Going to merge but the comments regarding Date management are to be left open for discussion/future improvements.

PR #383, opened 2021-05-14
Advanced Search Table & Misc Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#7325

*Description of changes:*

* Implemented the advanced search query.

* Updated Search results table.

* Updated property names to match api query params.

* Updated some form validations and resets.

*Notes:*

* Date range is currently untested due to the api not supporting yet, but it is a wip. 


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "2.1.5",
+  "version": "2.1.6",
```
severinbeauvais, line 3: This is OK, or we could consider it a "significant change" and set this to 2.2.0 instead. Thoughts?

cameron-freshworks: It's a full feature for sure, i think `2.2.0` makes sense!

src/components/advanced-search/advanced-search-form.vue
```diff
@@ -158,6 +160,15 @@ export default class AdvancedSearchForm extends Vue {
     if (!this.applicantFirstName) this.applicantLastNameRules = []
   }
 
+  /** Clear form. */
+  @Watch('formReset')
+  private clearForm (): void {
```
severinbeauvais, line 165: Would you want to do this only when `formReset` becomes True? (ie, `if (val) then ...`)

src/components/advanced-search/advanced-search-retrieve.vue
```diff
@@ -147,6 +150,14 @@ export default class AdvancedSearchRetrieve extends Vue {
     }
   }
 
+  /** Clear form. */
+  @Watch('formReset')
+  private clearForm (): void {
```
severinbeauvais, line 155: Same comment as above.

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: Not exactly. 
In this case i want to do this when the dialog closes or opens, so when the dialog state changes.

severinbeauvais: Is `formReset` tied to the dialog opening or closing?

severinbeauvais: Oh, it is...
```
:formReset="dialog"
```

severinbeauvais: Then why not call the prop something more topical?

As I reads now, I would expect this property to reset the form only when it's set to True.

cameron-freshworks: Sure

severinbeauvais: A name that makes more sense to me is "resetOnToggle" (but I don't like this name either 😛 ).

cameron-freshworks: Updated 👍

src/components/advanced-search/advanced-search-retrieve.vue
cameron-freshworks: Updated 👍

src/components/advanced-search/advanced-search-form.vue
severinbeauvais: Where I'm coming from is this...

From the parent's point of view, it's informing the component when `dialog` changes. The prop name should reflect that.

From the component's point of view, what it does with that prop (ie, reset the form) is an internal thing.

But a prop named "formReset" implies that the form should be reset when it becomes True.

cameron-freshworks: For sure! Updated to be more topical.

severinbeauvais (approved):
Nice code on the FF implementation.

Looks good.

cameron-freshworks:
> Nice code on the FF implementation.
> 
> Looks good.
Thanks!

and thanks for the good review, as always.

andrepestana-aot (approved):
Just some comments but looks very good to me!

src/enums/advanced-search-tabs.ts
```diff
@@ -3,4 +3,5 @@ export enum AdvancedSearchTabs {
   ADVANCED_SEACH_TABLE,
   ADVANCED_SEACH_RETRIEVE,
   ADVANCED_SEACH_INVALID,
+  ADVANCED_SEACH_NO_RESULTS
```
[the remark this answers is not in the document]
cameron-freshworks: 😁 Thank you! Good eye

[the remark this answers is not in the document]
cameron-freshworks: 4 typos
😆 Shame

src/interfaces/models.ts
```diff
@@ -258,11 +258,28 @@ export interface WaitingAddressSearchI {
...
-  endDate: string
+  compName: string
+  firstName: string
+  lastName: string
+  submittedStartDate: string
+  submittedEndDate: string
+}
+
+// The complete results from advanced nr search
+export interface AdvancedSearchResultsI {
```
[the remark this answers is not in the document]
cameron-freshworks: I agree! In Name request however, they've used `I` opposed to the IF we used throughout the Entities UI's. 
So in this case, i wanted to keep it consistent.

src/enums/advanced-search-tabs.ts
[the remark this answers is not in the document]
cameron-freshworks: All fixed up!

PR #385, opened 2021-05-18
Advanced Search UX Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#7326

*Description of changes:*

* Updated styling and spacing.

* Customized scroll bar for v-data-table.

* Validation tweaks.

* Enforce Date ranges. ie Start Date MUST come before End Date.

### DIFF
--- a/src/components/advanced-search/advanced-search-dates.vue
+++ b/src/components/advanced-search/advanced-search-dates.vue
@@ -10,6 +10,7 @@
             v-model="startDateText"
             :max="currentDateString"
             elevation="5"
+            @input="validateDates()"
           >
           </v-date-picker>
         </v-col>
@@ -19,8 +20,10 @@
             id="end-date-calendar"
             class="mt-2"
             v-model="endDateText"
+            :min="startDateText"
             :max="currentDateString"
             elevation="5"
+            @input="validateDates()"
           >
           </v-date-picker>
         </v-col>
@@ -69,6 +72,16 @@ export default class AdvancedSearchDates extends Mixins(DateMixin) {
   /** The current server date in string format. */
   get currentDateString (): string {
     return this.dateToDateString(this.getCurrentJsDate)
+  }
+
+  /** Is true when the End Date is less then the selected Start Date. */
+  get isEndDateInvalid (): boolean {
+    return new Date(this.startDateText) > new Date(this.endDateText)
+  }
+
+  /** Validate and handle end date out of range. */
+  private validateDates (): void {
+    if (this.isEndDateInvalid) this.endDateText = ''
   }
 
   /** Validate and submit form. */
--- a/src/components/advanced-search/advanced-search-form.vue
+++ b/src/components/advanced-search/advanced-search-form.vue
@@ -174,6 +174,7 @@ export default class AdvancedSearchForm extends Vue {
     this.clearDates()
   }
 
+  /** Emit form validity when the form is submitted or minimum requirements change thereafter. */
   @Watch('promptSubmit')
   @Watch('hasMinimumSearchCriteria')
   @Emit('isInvalid')
@@ -183,6 +184,7 @@ export default class AdvancedSearchForm extends Vue {
     return !this.hasMinimumSearchCriteria
   }
 
+  /** Emit submission data to parent if form passes validation. */
   @Watch('promptSubmit')
   @Emit('submitForm')
   private async emitSubmit () : Promise<AdvancedSearchI> {
--- a/src/components/advanced-search/advanced-search-form.vue
+++ b/src/components/advanced-search/advanced-search-form.vue
@@ -128,7 +128,7 @@ export default class AdvancedSearchForm extends Vue {
 
   /** True when the form is validating Names. */
   private get hasAppliedNameRules (): boolean {
-    return !!this.applicantFirstName && !this.applicantLastName
+    return this.applicantLastNameRules.length > 0
   }
 
   /** Is true when the minimum search criteria is met. */
@@ -151,6 +151,7 @@ export default class AdvancedSearchForm extends Vue {
     }
     await this.$refs.advancedSearchForm.validate()
 
+    this.emitInvalid()
     return this.hasMinimumSearchCriteria && this.isValid
   }
 
@@ -175,7 +176,6 @@ export default class AdvancedSearchForm extends Vue {
   }
 
   /** Emit form validity when the form is submitted or minimum requirements change thereafter. */
-  @Watch('promptSubmit')
   @Watch('hasMinimumSearchCriteria')
   @Emit('isInvalid')
   private emitInvalid (): boolean {
--- a/src/components/advanced-search/advanced-search-form.vue
+++ b/src/components/advanced-search/advanced-search-form.vue
@@ -20,7 +20,7 @@
       </v-col>
     </v-row>
 
-    <v-row no-gutters class="mt-n4">
+    <v-row no-gutters class="mt-n2">
       <v-col>
         <span class="title-bold-16">Applicant Name</span>
         <v-row>
@@ -126,6 +126,10 @@ export default class AdvancedSearchForm extends Vue {
     return !!this.startDate && !!this.endDate
   }
 
+  private get hasAppliedNameRules (): boolean {
+    return !!this.applicantFirstName && !this.applicantLastName
+  }
+
   /** Is true when the minimum search criteria is met. */
   private get hasMinimumSearchCriteria (): boolean {
     return !!this.compName || !!this.applicantLastName || (!!this.startDate && !!this.endDate)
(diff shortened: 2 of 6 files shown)


src/components/advanced-search/advanced-search-form.vue
```diff
@@ -169,9 +174,12 @@ export default class AdvancedSearchForm extends Vue {
     this.clearDates()
   }
 
+  @Watch('promptSubmit')
```
severinbeauvais, line 177: Since `emitInvalid()` doesn't depend on `promptSubmit`, why watch it? (If this is legitimate then please add a comment to explain 😃 )

src/components/advanced-search/advanced-search-form.vue
```diff
@@ -126,6 +126,11 @@ export default class AdvancedSearchForm extends Vue {
     return !!this.startDate && !!this.endDate
   }
 
+  /** True when the form is validating Names. */
+  private get hasAppliedNameRules (): boolean {
+    return !!this.applicantFirstName && !this.applicantLastName
```
severinbeauvais, line 131: I'm confused how we got from "the form is validating named" to "if first name and not last name", and also why this means we have applied name rules...?

src/components/dialogs/advanced-search.vue
```diff
@@ -16,12 +16,13 @@
...
-              Find an existing Name Request by searching for one or more of the following criteria:
+              Find an existing Name Request by searching for <strong>one or more</strong> of the following criteria:
             </v-card-text>
 
             <!-- Validation error message -->
-            <div v-if="invalidSearch" class="error-message copy-small my-n2">
-              Enter at least one search criteria
+            <div class="my-n2">
+              <span v-if="invalidSearch" class="copy-small error-message">Enter at least one search criteria</span>
+              <br v-else> <!-- Preserve Spacing when no errors present -->
```
severinbeauvais, line 25: Why a br instead of an empty span?

severinbeauvais:
Is this feature ready to be enabled permanently -- can the FF be deleted?

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: For sure, in this scenario it is legitimate. 

Since validations is tied to the submission we want to watch it, to not validate prematurely.

src/components/advanced-search/advanced-search-dates.vue
```diff
@@ -10,6 +10,7 @@
             v-model="startDateText"
             :max="currentDateString"
```
severinbeauvais, line 11: Maybe this should be `endDateText` ?

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: Added comment!

src/components/dialogs/advanced-search.vue
severinbeauvais: Or you could use `&nbsp;` to actually render a space character.

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: The latter is context when rules are applied. 
Another option is to check the rules array for length but i thought this was a little more explicit

cameron-freshworks: The why? I'm not sure, because Designs say so 🤷 hehe

src/components/dialogs/advanced-search.vue
cameron-freshworks: a span would require additional styling where a `br` does not.

src/components/advanced-search/advanced-search-form.vue
severinbeauvais: > The latter is context when rules are applied.
> Another option is to check the rules array for length but i thought this was a little more explicit

More direct, yes, but also more confusing (and won't work if other rules are added without updating this). In this case I'd prefer the length check -- it would answer all my questions!!!

src/components/dialogs/advanced-search.vue
severinbeauvais: OK. Weird. But OK.

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: I don't think so, since the start date max by default is the current date.
The endDate is dynamic and its range is dictated by the start date.

cameron-freshworks:
> Is this feature ready to be enabled permanently -- can the FF be deleted?

Yea! Actually i think so. Do we want to remove it now, or leave it in the event there is any issues with the queries in prod?

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: Okie doke, no prob!

src/components/advanced-search/advanced-search-dates.vue
severinbeauvais: So the min end date is the current start date? If so, this is fine and this conversation can be resolved.

andrepestana-aot (approved):
Just a question. Other than that looks good!

severinbeauvais:
> > Is this feature ready to be enabled permanently -- can the FF be deleted?
> 
> Yea! Actually i think so. Do we want to remove it now, or leave it in the event there is any issues with the queries in prod?

Your code is so nicely written... let's leave it in :)

(I'm serious about the first part.)

For the second part, sure I guess. Or ask PO.

src/components/advanced-search/advanced-search-table.vue
```diff
@@ -177,4 +179,28 @@ export default class AdvancedSearchTable extends Mixins(DateMixin) {
...
+::v-deep ::-webkit-scrollbar-thumb {
+  background: rgb(193,193,193);
+  border-radius: 10px;
+  height: 2rem;
+}
+
+/* Handle on hover */
+::v-deep ::-webkit-scrollbar-thumb:hover {
+  background: $gray6;
+}
```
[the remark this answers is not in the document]
cameron-freshworks: I'm not sure. Is there a FF alternative? 

I would have to check, but i'm assuming it would fallback to default in FF, which wouldn't be the worst thing.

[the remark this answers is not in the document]
severinbeauvais: > Those pseudo elements don't work for FF. Should we support FF?

Yes, we support "all major browsers".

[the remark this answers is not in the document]
andrepestana-aot: > I'm not sure. Is there a FF alternative?

The scrollbar customization is more limited but they are following the standard opposed to webkit browsers (most of them 🤔 )
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width

cameron-freshworks:
> > > Is this feature ready to be enabled permanently -- can the FF be deleted?
> > 
> > 
> > Yea! Actually i think so. Do we want to remove it now, or leave it in the event there is any issues with the queries in prod?
> 
> Your code is so nicely written... let's leave it in :)
> 
> (I'm serious about the first part.)
> 
> For the second part, sure I guess. Or ask PO.

Thanks 😁 

My only concern is the query doesn't perform the same in PROD as it does in our other environments. It would be the first time we've seen surprises in Prod 😆 

We can always remove it after, fairly easily.

src/components/advanced-search/advanced-search-dates.vue
cameron-freshworks: Exactly :) 

Screen Shot 2021-05-18 at 2 10 31 PM

src/components/advanced-search/advanced-search-form.vue
cameron-freshworks: Updated! 
Sorry it took so long, got pulled into something else at the end of the day there.

src/components/advanced-search/advanced-search-table.vue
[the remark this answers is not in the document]
cameron-freshworks: I'm not sure what to action here.
In FireFox this will fall back to default scrollbar values, yes. 

This difference is the small margin at the top of the scrollbar (ie so the scroll thumb doesn't start on the header), which was the design request for this customization.

If you think this is important i can start looking to this customization that is exclusive to this browser (FF)?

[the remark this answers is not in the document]
andrepestana-aot: The question is just a heads up for the difference in FF. If you think it's ok for FF, that's fine. I'm not sure if UX Assurance checks it for every major browser. I can't "Resolve conversations" here, so consider it resolved and thanks for checking!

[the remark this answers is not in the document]
cameron-freshworks: Definitely, i appreciate your opinion here! 
I will look into FireFox a bit here this evening and if i can come up with something that satisfies the browser, will 100% implement it :) 

That being said, if there was no default styling/behaviour it could be an issue, but luckily in this case the default is universal, so it's impact is a little smaller!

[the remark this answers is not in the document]
cameron-freshworks: After a little bit of research into FF, it appears the customization is unfortunately limited to bar width and colour :(

severinbeauvais (approved):
GTG!

PR #386, opened 2021-05-18
Implemented staff payment

severinbeauvais:
*Issue #:* /bcgov/entity#6554

**This is the first part of the UI implementation, so that it can be reviewed by the UX team. Although this is exactly according to the design comps, I expect some changes will be requested. Also, the implementation needs to be completed in a future PR when the namex API is updated to support staff payment data.**

*Description of changes:*

- imported shared component, enum and interface
- implemented tabs in Payment modal
- implemented Staff Payment "interface component" (between tab and shared component)
- added auth-getter plugin
- fetched keycloak roles
- added state/setter/getter for keycloak roles
- added state/setter/getter for staff payment data
- added staff-payment feature flag
- updated app version
- NB: needs completion when Namex API can handle staff payment data
- NB: this feature is guarded by a new feature flag (staff-payment-enabled) so that it can be tested in Dev before being completed and released in Test/Prod

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -69,7 +69,7 @@
 // libraries, etc
 import { Component, Mixins } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
-import { getFeatureFlag } from '@/plugins'
+import { getFeatureFlag, getKeycloakRoles } from '@/plugins'
 import { DateMixin } from '@/mixins'
 
 // dialogs and other components
@@ -130,6 +130,7 @@ export default class App extends Mixins(DateMixin) {
   @Action setIncorporateLoginModalVisible!: ActionBindingIF
   @Action togglePaymentModal!: ActionBindingIF
   @Action setCurrentJsDate!: ActionBindingIF
+  @Action setKeycloakRoles!: ActionBindingIF
 
   /** The Update Current JS Date timer id. */
   private updateCurrentJsDateId = 0
@@ -156,6 +157,17 @@ export default class App extends Mixins(DateMixin) {
 
     // listen for spinner show/hide events
     this.$root.$on('showSpinner', (flag = false) => { this.showSpinner = flag })
+
+    // in case user is already logged in,
+    // get and store keycloak roles
+    try {
+      const keycloakRoles = getKeycloakRoles()
+      this.setKeycloakRoles(keycloakRoles)
+      console.info('Got roles!') // eslint-disable-line no-console
+    } catch (error) {
+      // just log the error message
+      console.log(`Did not get roles (${error.message})`) // eslint-disable-line no-console
+    }
   }
 
   /** Fetches and stores the current JS date. */
--- a/src/components/dialogs/payment.vue
+++ b/src/components/dialogs/payment.vue
@@ -1,54 +1,96 @@
 <template>
   <v-dialog max-width="45rem" :value="isVisible" persistent>
     <v-card>
+      <v-tabs id="payment-tabs">
+        <v-tabs-items v-model="paymentTab">
 
-      <v-card-title class="d-flex justify-space-between">
-        <div>Confirm Name Request</div>
-        <v-btn icon large class="dialog-close float-right" @click="hideModal()">
-          <v-icon>mdi-close</v-icon>
-        </v-btn>
-      </v-card-title>
+          <v-tab-item>
+            <v-card-title class="d-flex justify-space-between">
+              <div>Staff Payment</div>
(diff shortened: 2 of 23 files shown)


src/App.vue
```diff
@@ -156,6 +157,17 @@ export default class App extends Mixins(DateMixin) {
...
+
+    // in case user is already logged in,
+    // get and store keycloak roles
+    try {
+      const keycloakRoles = getKeycloakRoles()
+      this.setKeycloakRoles(keycloakRoles)
+      console.info('Got roles!') // eslint-disable-line no-console
+    } catch (error) {
+      // just log the error message
+      console.log(`Did not get roles (${error.message})`) // eslint-disable-line no-console
```
severinbeauvais, line 169: I decided to just log a friendly-looking message here (ie, not the stack trace) since this will happen every time a non-logged-in user loads this app.

src/components/dialogs/payment.vue
```diff
@@ -62,39 +104,50 @@ import { ActionBindingIF } from '@/interfaces/store-interfaces'
...
-  },
-  props: {
-    onActivate: {
-      type: Function,
-      default: async () => undefined
-    },
-    onCancel: {
-      type: Function,
-      default: async () => {}
-    }
```
severinbeauvais, line 75: I replaced `onCancel` with a prop decorator/declaration below, and deleted `onActivate` which is no longer used.

src/components/payment/staff-payment.vue
```diff
@@ -0,0 +1,103 @@
+<template>
```
severinbeauvais, line 1: This is what I call an "interface component" that sits between the tab item and the shared component. It exists so that the tab item doesn't need to know about the low level details, and the shared component doesn't need to know about the high level getters/setters/etc. This interface component will be different in different apps.

src/interfaces/state-interface.ts
```diff
@@ -1,14 +1,17 @@
...
 export interface StateModelIF {
   common: {
-    currentJsDate: Date
-  },
+    currentJsDate: Date,
+    keycloakRoles: Array<string>
+  }
   newRequestModel: NewRequestIF
-  errorModel?: {},
+  staffPayment: StaffPaymentIF
```
severinbeauvais, line 14: This is where staff payment data is stored. It is not saved to the filing (NR) yet -- this is future work that will be dictated by the spike ticket and API ticket in this story (6288).

src/plugins/authHelper.ts
```diff
@@ -0,0 +1,33 @@
+import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
```
severinbeauvais, line 1: This is just a copy of this helper/util from our other apps.

src/store/getters.ts
```diff
@@ -991,7 +1005,7 @@ export const getDraftNameReservation = (state: StateIF): DraftReqI => {
         }
         if (!data['additionalInfo'].includes('*** Legal Name:')) {
           data['additionalInfo'] += '\n\n'
-          let notice = `*** Legal Name: ${this.nameChoices.name1} ***`
+          let notice = `*** Legal Name: ${getNameChoices(state)?.name1} ***`
```
severinbeauvais, line 1008: Cam, this is correct, right?

cameron-freshworks: Syntax wise? Yes, i think so when using a getter inside another getter. Does this give you any lint errors or is there a type on the return from the getter?

cameron-freshworks (approved):
Looks good to me. 
With some work pending and the fact it's FF, low impact. g2g!

src/store/getters.ts
severinbeauvais: No, the change from `this.nameChoices` (which showed an error due to "this") to the local getter (which I sort of copied from another method in here).

severinbeauvais: Ie, I fixed an error, but it feels weird to call a getter like this :P

cameron-freshworks: Referring the lack of `this`?

severinbeauvais: Yes.

cameron-freshworks: For sure, i'm not sure the terminology but the getters don't belong to a class so aren't scoped? Is that the reason maybe?

severinbeauvais: I think so. I think this is correct but I wanted to let you know.

andrepestana-aot (commented):
Sorry for the late review. Looks good to me too!

PR #388, opened 2021-05-27
Staff Payment validation and UX fixes

severinbeauvais:
*Issue #:* /bcgov/entity#6554

*Description of changes:*
- set Payment modal min width for better UX
- adjusted Payment modal whitespace
- updated Payment modal button order
- updated Staff Payment validation (state instead of prop)
- added Staff Payment error border
- removed Staff Payment extra whitespace
- moved specific base class to specific component
- changed button ids to classnames (due to multiple instances)
- moved Request Details class to parents
- removed unused component from Payment Complete
- updated unit tests
- updated app version
- updated shared dependencies
- updated browsers list

See the list of validation tests in ticket, if you're curious.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "2.2.3",
+  "version": "2.2.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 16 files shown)


package-lock.json
```diff
@@ -5205,9 +5205,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001164",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001164.tgz",
-      "integrity": "sha512-G+A/tkf4bu0dSp9+duNiXc7bGds35DioCyC6vgK2m/rjA4Krpy5WeZgZyfH2f0wj2kI6yAWWucyap6oOwmY1mg==",
+      "version": "1.0.30001230",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001230.tgz",
+      "integrity": "sha512-5yBd5nWCBS+jWKTcHOzXwo5xzcj4ePE/yjtkZyUV1BTUmrBaA9MRGC+e7mxnqXSA90CmCA8L3eKLaSUkt099IQ==",
```
severinbeauvais, line 5210: This is the update for "browser list" prompted by npm.

src/assets/scss/base.scss
```diff
@@ -288,14 +288,6 @@ button.reserve-submit-btn {
   margin: 0 !important;
 }
 
-#submit-back-btn {
```
severinbeauvais, line 291: I moved this to the component where it is declared.

src/components/common/applicant-info-nav.vue
```diff
@@ -1,16 +1,15 @@
 <template>
   <v-col cols="5" class="text-right py-0">
     <v-btn x-large
-           id="submit-back-btn"
-           class="mr-3"
+           class="submit-back-btn mr-3"
```
severinbeauvais, line 4: I changed id to a classname (and also below) to get rid of "duplicate id" messages.

This is actually the only place where this id was used, but apparently NR UI sometimes renders 2 of these components, so you would get a duplicate id.

src/components/common/request-details.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div id="request-details">
-    <div class="mt-4">
+    <section>
```
severinbeauvais, line 3: The whitespace shouldn't be in this component; it should be in its parent (which knows how this component is supposed to be laid out). I used a section element to retain the content in one chunk (and improve semantics).

src/components/dialogs/payment-complete.vue
```diff
@@ -35,7 +35,6 @@
 import { Component, Mixins, Watch } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
 import PaymentConfirm from '@/components/payment/payment-confirm.vue'
-import RequestDetails from '@/components/common/request-details.vue'
```
severinbeauvais, line 38: RequestDetails was not used here.

While in this component, I updated the element name.

src/components/dialogs/payment.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-dialog max-width="45rem" :value="isVisible" persistent>
+  <v-dialog min-width="32rem" max-width="45rem" :value="isVisible" persistent>
```
severinbeauvais, line 2: I was just playing around and found that setting the minimum width here really improves the UX on a small screen.

src/components/dialogs/payment.vue
```diff
@@ -12,25 +12,24 @@
               </v-btn>
             </v-card-title>
 
-            <v-card-text class="copy-normal pt-0">
+            <v-card-text class="copy-normal pt-5">
               <StaffPayment
-                :doValidate="validateStaffPayment"
+                ref="staffPaymentComponent"
```
severinbeauvais, line 17: I needed to make "validate" a state instead of a readonly prop, so I use this to call a method in this component.

src/components/dialogs/payment.vue
```diff
@@ -12,25 +12,24 @@
...
+                id="payment-close-btn"
+                class="button-blue"
+                :disabled="isLoadingPayment">Back</v-btn>
               <v-btn
                 @click="confirmPayment()"
                 id="payment-pay-btn"
                 class="primary px-4"
-                text
                 :loading="isLoadingPayment">Submit Name Request</v-btn>
-              <v-btn
```
severinbeauvais, line 29: I just moved this button up.

src/components/dialogs/payment.vue
```diff
@@ -61,14 +60,12 @@
                 @click="cancelPayment()"
                 id="payment-cancel-btn"
                 class="button-red px-4"
-                text
```
severinbeauvais, line 64: This is not a text button per se. It is already styled to look as we want it.

src/components/dialogs/payment.vue
```diff
@@ -159,12 +161,6 @@ export default class PaymentDialog extends Mixins(
 
   /** Clears store property to hide this modal. */
   async hideModal () {
-    // go to previous page
```
severinbeauvais, line 162: This is moved to its own method.

src/components/dialogs/payment.vue
```diff
@@ -193,19 +189,27 @@ export default class PaymentDialog extends Mixins(
     }
   }
 
+  /** Called when user clicks "Back" button. */
+  goBack () {
+    // disable validation
+    this.$refs.staffPaymentComponent && this.$refs.staffPaymentComponent.setValidation(false)
```
severinbeauvais, line 195: The component originally doesn't exist (until the tab is rendered).

This is a call into the component to set state.

src/components/payment/staff-payment.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-  <div id="staff-payment" class="pt-6">
-    <staff-payment-component
+  <div id="staff-payment" :class="{'invalid-section': showStaffPaymentInvalidSection}">
```
severinbeauvais, line 2: To keep the component styling simple, I set this class here, in its parent. (This is also what is done in Edit UI for this component.)

src/components/payment/staff-payment.vue
```diff
@@ -36,12 +36,27 @@ export default class StaffPayment extends Vue {
...
+  /** Whether to show the staff payment invalid section styling. */
+  private get showStaffPaymentInvalidSection (): boolean {
+    const option = this.getStaffPayment.option ?? StaffPaymentOptions.NONE
+    // True if no option is selected
+    return this.validate && (option === StaffPaymentOptions.NONE)
+  }
+
+  /** Called externally to set validation state. */
+  public setValidation (val: boolean) {
+    this.validate = val
```
severinbeauvais, line 50: This is now a state, so I can toggle it True or False as needed (instead of making it responsive like a prop). I need this state to "stick" so that I can enable it when the user tries to submit this NR, and clear it under various scenarios (like reloading this tab, selecting a different radio button, etc). I was not able to do that without a state.

src/components/payment/staff-payment.vue
```diff
@@ -100,4 +122,14 @@ export default class StaffPayment extends Vue {
 ::v-deep .v-input--radio-group__input {
   background-color: white;
 }
+
+// remove margin below radio group
+::v-deep .v-input--radio-group > .v-input__control > .v-input__slot {
+  margin-bottom: 0 !important;
```
severinbeauvais, line 128: Ideally, this (and the rules on line 116 and 132) should be in the shared component itself, but I didn't want to destabilize its use in other UIs.

vysakh-menon-aot (approved):

src/components/dialogs/payment.vue
cameron-freshworks: Is optionally chaining the reference an option here?
`this.$refs.staffPaymentComponent?...`

cameron-freshworks (approved):
Looks good!

src/components/dialogs/payment.vue
severinbeauvais: Hmm, I would think so, but I'm getting this:
```
error: Expected an assignment or function call and instead saw an expression (no-unused-expressions) at src/components/dialogs/payment.vue:195:5:
  193 |   goBack () {
  194 |     // disable validation
> 195 |     this.$refs.staffPaymentComponent?.setValidation(false)
      |     ^
```

severinbeauvais: I think that, technically, optional chaining works above, but the linter is complaining about it.

I think that, as in C, JS is fine with expressions that aren't assigned to anything, as above, but the "no-unused expression" is trying to save us from ourselves.

I suppose I could write `const f == this.blah?.func()` but that's not really any cleaner.

Another option is `if (this.blah) this.blah.func()`.

I think what I have is OK, but if you prefer something else, lmk.

cameron-freshworks: All good! Was just curious 👍

PR #394, opened 2021-06-03
Update sbc-common-components version number

eve-git:
*Issue #:* /bcgov/entity#6682

*Description of changes:*
Update sbc-common-components version number

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "launchdarkly-js-client-sdk": "^2.18.3",
     "lodash": "^4.17.20",
     "remove-accents": "^0.4.2",
-    "sbc-common-components": "^2.4.18",
+    "sbc-common-components": "2.4.18",
     "style-loader": "^1.3.0",
     "vue": "^2.6.12",
     "vue-contenteditable": "^1.0.2",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "launchdarkly-js-client-sdk": "^2.18.3",
     "lodash": "^4.17.20",
     "remove-accents": "^0.4.2",
-    "sbc-common-components": "2.4.17",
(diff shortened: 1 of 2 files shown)


eve-git:
Passed unit test

package.json
```diff
@@ -27,7 +27,7 @@
     "launchdarkly-js-client-sdk": "^2.18.3",
     "lodash": "^4.17.20",
     "remove-accents": "^0.4.2",
-    "sbc-common-components": "2.4.17",
+    "sbc-common-components": "^2.4.18",
```
severinbeauvais, line 30: Remove the ^. Its purpose is to auto-upgrade this package every time you npm install. But we don't want it to auto-upgrade without our knowledge (since we need to know/test whatever changed).

package.json
```diff
@@ -27,7 +27,7 @@
     "launchdarkly-js-client-sdk": "^2.18.3",
     "lodash": "^4.17.20",
     "remove-accents": "^0.4.2",
-    "sbc-common-components": "2.4.17",
```
severinbeauvais, line 30: I am OK with not updating the app version, since the Sbc Common Components version also displays in the footer "i" tooltip (so we can verify when it's deployed).

package.json
eve-git: done.

severinbeauvais (approved):
Looks good.

@andrepestana-aot If you approve, please merge. Thanks!

package.json
andrepestana-aot: I would ask about it :)

andrepestana-aot (approved):
LGTM

package.json
severinbeauvais: We implemented that feature for internal use. It still does what it needs to do. No need to ask anyone else :)

PR #395, opened 2021-06-08
6074 Update wait time tool tips

argush3:
*Issue #:* /bcgov/entity#6074

*Description of changes:*

* Updated both wait time tooltips to be 270px wide, text left-justified and arrow left nudge adjustment
* Added dotted underline styling to wait time labels to the right of the wait time circles
* Updated tooltip text for both wait times

### DIFF
--- a/src/components/new-request/stats.vue
+++ b/src/components/new-request/stats.vue
@@ -12,7 +12,7 @@
 <!--            No Wait Time-->
 <!--          </div>-->
 <!--        </div>-->
-      <v-tooltip bottom nudge-left="50"
+      <v-tooltip bottom nudge-left="45"
         content-class="bottom-tooltip wait-time-tooltip"
         transition="fade-transition"
       >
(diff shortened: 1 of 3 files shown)


src/components/new-request/stats.vue
```diff
@@ -123,6 +125,8 @@ export default class Stats extends Vue {
   padding-left: 8px;
   font-size: 12px;
   line-height: 18px;
+  text-decoration-line: underline;
+  text-decoration-style: dotted;
```
severinbeauvais, line 129: Use base class `dotted-underline` instead.

Also, should this use class `cursor-default`?

andrepestana-aot (commented):
LGTM!

andrepestana-aot (approved):

src/components/new-request/stats.vue
argush3: Not sure about your comment regarding `cursor-default`.  There is no mention of what the cursor needs to be in the designs from what I can see.

severinbeauvais: I think we don't handle the tooltip hover cursor consistently in our apps...
Anyway, in this app, we don't change the cursor for tooltips, so this is fine.

severinbeauvais:
I can't remember why we get the testing-coverage check error above (but I think we've seen it before). Please follow up with Patrick Wei. Don't merge until that's fixed please.

severinbeauvais (approved):
The code looks good!

argush3:
> I can't remember why we get the testing-coverage check error above (but I think we've seen it before). Please follow up with Patrick Wei. Don't merge until that's fixed please.

ok, will follow up with Patrick.

argush3:
@severinbeauvais talked to Patrick and it seems to have been just a connectivity issue.  He re-ran it and things checked out.

PR #406, opened 2021-06-21
7398 - Added Folio Number

andrepestana-aot:
*Issue #:* /bcgov/entity#7398

*Description of changes:*

- Added Folio Number
- Bumped up version

**Folio Number is optional only when the NR is not create on behalf of the person who is requesting it. This will not be in edit mode.**


src/components/common/applicant-info-1.vue
```diff
@@ -461,6 +462,7 @@ export default class ApplicantInfo1 extends Mixins(ActionMixin) {
   }
 
   set actingOnOwnBehalf (value) {
+    if (value) this.setFolioNumber('') // no folio number in case it is on own behalf
```
severinbeauvais, line 465: Are you sure about this?

I think someone creating their own NR might still want to enter something in the FN field...?

src/components/common/applicant-info-3.vue
```diff
@@ -68,6 +68,15 @@
                         hide-details="auto"
                         label="Last Name" />
         </v-col>
+        <v-col cols="2" v-if="!getEditMode"></v-col>
```
severinbeauvais, line 71: This should be in its own v-row (as there are 12 columns in a row, and here you are overflowing the first row in order to get another row... which is not the right way to do it).

src/components/common/request-details.vue
```diff
@@ -45,6 +45,7 @@
               <li>{{`${contactPerson}`}}</li>
               <li>{{applicant.emailAddress}}</li>
               <li>{{applicant.phoneNumber}}</li>
+              <li><span class="label" v-if="!!getFolioNumber">Folio: </span>{{getFolioNumber}}</li>
```
severinbeauvais, line 48: Just curious: why do you need the !! ?

src/components/common/applicant-info-1.vue
andrepestana-aot: This is because it should send a Folio Number only in case it is on someone else's behalf. Otherwise someone can add a Folio Number, come back and change to their own behalf. In this case the old value would be kept and and sent as Folio Number which is not correct.

src/mixins/payment-mixin.ts
```diff
@@ -159,6 +160,9 @@ export class PaymentMixin extends Mixins(ActionMixin) {
       const parsedAccountInfo = JSON.parse(accountInfo)
       headers['Account-Id'] = parsedAccountInfo.id
     }
+    if (this.getStaffPayment.option === StaffPaymentOptions.NONE) {
```
severinbeauvais, line 163: I don't understand this. Why only set FN when staff payment is none? Is this because it doesn't apply when staff are doing the payment? If so then maybe there's a more clear (and without assuming the option) way to do this. Or add a comment saying what you're doing (so it looks like you are doing this explicitly/on purpose).

src/components/common/request-details.vue
andrepestana-aot: I don't think we need. It was just to explicit. I'm gonna remove it.

src/components/common/applicant-info-1.vue
severinbeauvais: OK, I see this clears the value in case the user changes from someone else's behalf to their own behalf...

But why not allow a user to enter a FN on their own behalf?

src/components/common/applicant-info-3.vue
andrepestana-aot: ok. You're right. Changing...

src/components/common/request-details.vue
severinbeauvais: OK. I know we sometimes use that to cast something else to a boolean (to satisfy property/method typing), but it's probably not needed here.

src/components/common/applicant-info-1.vue
severinbeauvais: I see that there is discussion on FN for premium users, but that doesn't mean their won't create a NR on their own behalf.

Maybe I don't understand. But if the BA/PO say it is so, then ok.

src/mixins/payment-mixin.ts
andrepestana-aot: This is because in buildStaffPayment we set the folioNumber to the headers. If it's not StaffPayment we need to set it to headers. Maybe we'd better refactor this buildStaffPayment function's name to buildHeaders and include all the header related logic to it. What do you think?

severinbeauvais: Yes, that would make more sense :)

src/components/common/applicant-info-1.vue
andrepestana-aot: > But why not allow a user to enter a FN on their own behalf?

This is a requirement:
Scenario 1: Display folio field
GIVEN I am any type of user
WHEN I un-check "I am submitting this on my own behalf"
THEN I can enter a folio number

severinbeauvais: OK!

src/components/common/request-details.vue
```diff
@@ -104,12 +108,16 @@ export default class RequestDetails extends Vue {
...
+.label {
   font-weight: bold !important;
   font-size: 1rem;
   color: $dk-text;
 }
 
+h4 {
+  @extend .label
+}
+
```
cameron-freshworks, line 120: Did these not exist globally? Seems very generic!

cameron-freshworks (approved):
Minor comment, but looks good!

src/components/common/request-details.vue
andrepestana-aot: yes, I see that there's a class in base.scss that has those properties, then I changed to that other class. Thanks.

andrepestana-aot:
I changed to Draft to wait the environment be fixed and I can run a complete test again.

severinbeauvais (approved):
Looks good! I like the small comments here and there to explain what we're working with.

PR #411, opened 2021-06-30
Updated defensive coding when conflict date or id are missing

severinbeauvais:
*Issue #:* /bcgov/entity#6992

*Description of changes:*
- display "Unknown" if conflict date is null
- display "Unknown" if conflict id is null
- cleaned up logic
- updated app version

### DIFF
--- a/src/components/new-request/analyze-results.vue
+++ b/src/components/new-request/analyze-results.vue
@@ -82,12 +82,13 @@
                 <v-col cols="auto">
                   <div v-for="(corp, n) in conflicts" :key="'conflict-' + n">
                     <b>{{ corp.name }}</b><br>
-                    <span v-if="conflictDate && conflictId.startsWith('NR ')">
-                      <b>Submitted Date:</b> {{ conflictDate + ', ' }}</span>
-                    <span v-if="conflictDate && !conflictId.startsWith('NR ')">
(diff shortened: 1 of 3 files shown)


src/components/new-request/analyze-results.vue
```diff
@@ -82,12 +82,13 @@
...
-                      <b>Incorporation Date:</b> {{ conflictDate + ', ' }}</span>
-                    <span v-if="conflictId && !conflictId.startsWith('NR ')">
-                      <b>Corporation No.:</b> {{ conflictId }}</span>
+                    <template v-if="conflictId.startsWith('NR ')">
+                      <span><b>Submitted Date:</b> {{ conflictDate }}</span>
+                    </template>
+                    <template v-if="!conflictId.startsWith('NR ')">
+                      <span><b>Incorporation Date:</b> {{ conflictDate + ', ' }}</span>
+                      <span><b>Corporation No.:</b> {{ conflictId }}</span>
+                    </template>
```
severinbeauvais, line 91: The old code did not display the line when the value was null.
The new code always displays the line (with "Unknown" value if it's null). This is the concept of "letting the user know there's something there, even if we can't say what it is" (ie, don't hide error).

cameron-freshworks (approved):
Looks good!

PR #414, opened 2021-07-07
Cleaned up image source paths

severinbeauvais:
*Issue #:* /bcgov/entity#7116

### DIFF
--- a/src/views/landing.vue
+++ b/src/views/landing.vue
@@ -96,7 +96,7 @@ export default class Landing extends Vue {
 }
 
 #upper-row {
-  background: url('../assets/images/analyze-name-bg.jpg') no-repeat bottom;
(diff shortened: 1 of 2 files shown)


src/views/landing.vue
```diff
@@ -96,7 +96,7 @@ export default class Landing extends Vue {
 }
 
 #upper-row {
-  background: url('../assets/images/analyze-name-bg.jpg') no-repeat bottom;
+  background: url('~@/assets/images/analyze-name-bg.jpg') no-repeat bottom;
```
severinbeauvais, line 99: This is a weird-looking path but I tried without `~` and it didn't work; this does. So here it is.

cameron-freshworks: Interesting!

cameron-freshworks (approved):
Looks good

PR #420, opened 2021-07-10
7960 search combo

kialj876:
*Issue #:* /bcgov/entity#7960

*Description of changes:*
- search combo feature


public/config/configuration.json
```diff
@@ -14,7 +14,7 @@
   "NAMEREQUEST_LD_CLIENT_ID": "5fbec4b5d998e30990020007",
   "LEGAL_API_URL": "https://legal-api-dev.apps.silver.devops.gov.bc.ca",
   "LEGAL_API_VERSION": "/api/v1",
-  "NAMEX_API_URL": "https://namex-dev.apps.silver.devops.gov.bc.ca",
+  "NAMEX_API_URL": "https://namex-api-dev.apps.silver.devops.gov.bc.ca",
```
cameron-freshworks, line 17: Just trying to run this branch locally and the requests to the url are failing. I'm guessing Namex-Api will require a new build to Dev before this will take effect?

src/components/common/names-capture.vue
```diff
@@ -97,6 +97,7 @@
                               @blur="handleBlur()"
                               @input="editChoices('name1', $event, true)"
                               :filled="getIsAssumedName"
+                              :class="!getIsAssumedName ? 'read-only-mode' : ''"
```
cameron-freshworks, line 100: I think this could be simplified using vue class binding since there is no alternative class used in the ternary, ie
`{ 'read-only-mode': !getIsAssumedName }` or something along those lines.

src/components/dialogs/name-check-issues.vue
```diff
@@ -0,0 +1,78 @@
...
+            <v-icon>mdi-close</v-icon>
+          </v-btn>
+        </v-col>
+      </v-row>
+      <p class="dialog-title ma-0 pt-5">
+        {{ options.title }}
+      </p>
+      <v-card-text class="pt-10">
+        <p v-for="line, index in options.text"
+            :key="index"
```
cameron-freshworks, line 19: Might want to expand the id just a little to keep it unique, ie `options-text-${index}` or something like that? 
Just in the event if you had multiple keys using index, it could throw errors. 
Pretty minor!

src/components/dialogs/name-check-issues.vue
```diff
@@ -0,0 +1,78 @@
...
+        <v-col v-if="options.acceptText" class="pl-3" cols="auto">
+          <v-btn class="accept-btn" @click="proceed(true)">
+            {{ options.acceptText }}
+            <v-icon>mdi-chevron-right</v-icon>
+          </v-btn>
+        </v-col>
+      </v-row>
+    </v-card>
+  </v-dialog>
+</template>
```
cameron-freshworks, line 40: Template is nice, good job!

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
+<template>
+  <v-data-table :headers="headers"
+                fixed
+                hide-default-header hide-default-footer
+                :items="items"
+                item-key="problem"
+                single-expand
+                :expanded.sync="expanded"
+                no-data-text=""
+                show-expand>
```
cameron-freshworks, line 10: I don't mind this layout, but i think* we've been going with the next line with 1 tab for props. Up to you though!

public/config/configuration.json
kialj876: oops thank you haha. I have been hitting the api locally, but was just trying this and thought I had an auth issue

src/components/common/names-capture.vue
kialj876: sure I can update that

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
...
+      <tr v-else :class="isExpanded ? 'no-hover' : ''" :colspan="headers.length" @click="expand(!isExpanded)">
+        <v-row no-gutters class="conflict-row py-5 px-4 border-top">
+          <v-col cols="9">
+            <v-row no-gutters>
+              <v-col cols="auto" class="nudge-down">
+                <v-icon :color="item.iconColor">{{ item.icon }}</v-icon>
+              </v-col>
+              <v-col cols="auto" class="table-text pl-3 pt-1">
+                <span v-html="item.problem"/>
+                <b v-for="(word, index) in item.words" :key=index>
```
cameron-freshworks, line 55: Another instance where the key index might spit out a complaint if it's not unique, not major though!

src/components/dialogs/name-check-issues.vue
kialj876: Oh interesting, I didn't understand how that worked. So it sets the id to the key binding? So you're saying I should change the key binding to that right?

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: yeah I'll update it, otherwise we might run into it later

public/config/configuration.json
cameron-freshworks: oh right on! I will update locally and give it another try :)

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: ha, okay I've been wondering what the best way for this is. I've done it this way everywhere... Is it okay if I leave it so it's consistent throughout the namecheck vues? Otherwise I will have a lot of code to go though haha

cameron-freshworks: Totally, of course :) I'm easy!

src/components/dialogs/name-check-issues.vue
cameron-freshworks: Absolutely, it doesn't have to be the exact same as the id, just a means of making it unique from other potential `:key=index` values.

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
...
+      return 'Verifying Results ...'
+    }
+    if (this.tabIconVerdict.icon === 'mdi-check-circle') {
+      return 'Ready for Review'
+    }
+    return 'Attention Required'
+  }
+  get itemsConflict (): Array<NameCheckItemIF> {
+    let items = []
+    console.log('has exact errors: ', this.hasErrorsConflictsExact)
```
cameron-freshworks, line 460: Is this log for development only? or is the plan to leave this in?

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
...
+    return 'http://www.strategis.ic.gc.ca'
+  }
+  back () {
+    this.setActiveComponent('Tabs')
+  }
+  checkAgain () {
+    this.startAnalyzeName(null)
+  }
+  clearError (errorType: NameCheckErrorType) {
+    console.log(errorType)
```
cameron-freshworks, line 652: Same as above. Definitely has value for development but i'm not sure if we would want to commit this.

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
...
+      return { color: 'red darken-2', icon: 'mdi-alert-octagon' }
+    }
+    if (this.hasIssuesConflictCaution || this.hasIssuesStructureCaution) {
+      return { color: 'orange darken-2', icon: 'mdi-alert' }
+    }
+    return { color: 'green darken-2', icon: 'mdi-check-circle' }
+  }
+  get tradeMarkDBLink (): string {
+    // TODO: make it a config var?
+    return 'http://www.strategis.ic.gc.ca'
```
cameron-freshworks, line 643: If it's a static link i don't see any issue with keeping it in a getter.

cameron-freshworks:
Is quick search still a thing? I see a lot of calls to `getQuickSearch()` but there is configuration missing locally to run that.
Just curious if we should set that config to test locally as well?

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
...
+                      to submit signed consent to the Business Registry as part of the review process.
+                    </p>
+                  </v-tooltip>
+                  to use this name.
+                </p>
+              </div>
+              <div v-else v-html="item.expandedInfo1">{{ item.expandedInfo1 }}</div>
+              <div v-if="item.expandedInfoBlock1" class="pt-7">
+                <p v-for="line, index in item.expandedInfoBlock1" :key="index" class="ma-0" v-html="line">
+                  {{ line }}
```
cameron-freshworks, line 150: 1 more key instance :) 
Also, is `{{ line }}` required when templating raw html?

Vue ref: https://vuejs.org/v2/guide/syntax.html#Raw-HTML

If not, this would apply to the html templating below.

kialj876:
> Is quick search still a thing? I see a lot of calls to `getQuickSearch()` but there is configuration missing locally to run that.
> Just curious if we should set that config to test locally as well?

It is yeah, but it is within the combo search so you just need the id/secret in set your config

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
...
+        if (newSearch) this.expanded = []
+      }
+    } else {
+      this.expanded = []
+    }
+  }
+
+  mounted () {
+    this.updateExpanded(this.items)
+  }
```
cameron-freshworks, line 259: Super minor but we usually put the vue lifecycle hooks near the top of the script, after property definitions but before computed properties etc. No action required here, your call!

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
...
+                <v-tooltip top content-class="top-tooltip" transition="fade-transition">
+                  <template v-slot:activator="{ on, attrs }">
+                    <span class="dotted-underline" v-bind="attrs" v-on="on">Similar names</span>
+                  </template>
+                  <p class="ma-0">
+                    Similar names contain the same, synonymous, or similar looking or sounding words.
+                  </p>
+                </v-tooltip>
+                in different business categories may be approved at the discretion of the
+                Business Registry. If you feel your name is unique within your business category,
```
cameron-freshworks, line 116: This sentence seems a little off, with the comma maybe?

src/store/getters.ts
```diff
@@ -1076,3 +1065,74 @@ export const getStaffPayment = (state: StateIF): StaffPaymentIF => {
 export const getFolioNumber = (state: StateIF): string => {
   return state.stateModel.newRequestModel.folioNumber
 }
+/** Name Check Getters
+ * TODO: move existing getters used only for name check above to here
+ * TODO: eventually move this all out of vuex (if we refactor to composition api)
```
cameron-freshworks, line 1070: What would the scope of this look like?? Pretty large task?

cameron-freshworks (approved):
Amazing work sir, overall code is great.
Just a few comments on some vue syntax and structure but nothing that would stop this from going in.
Due to the scope of this i would also recommend @severinbeauvais have a look as well, if he has capacity 😁

src/store/getters.ts
kialj876: I would say probably a day or 2ish if the first batch of initial composition-api setup stuff is done first. Almost none of the name-check stuff is needed in the vuex global store and it would probably cut the variables/functions in getters/actions/mutations/state in half almost

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: haha - it's copy/pasted from the UX copy so it's up to them

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: okay yeah I was wondering about this too, I'll shift it up

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: Ah, you're saying every element I have within the raw html needs it's own bind?

src/components/new-request/name-check/name-check.vue
kialj876: will remove the comment

src/components/new-request/name-check/name-check.vue
kialj876: sorry will take these out

src/components/dialogs/name-check-issues.vue
```diff
@@ -0,0 +1,78 @@
...
+        <v-col>
+          <v-btn cols="6" class="close-btn outlined" icon :ripple="false" size="2.5rem" @click="proceed(false)">
+            <v-icon>mdi-close</v-icon>
+          </v-btn>
+        </v-col>
+      </v-row>
+      <p class="dialog-title ma-0 pt-5">
+        {{ options.title }}
+      </p>
+      <v-card-text class="pt-10">
```
severinbeauvais, line 17: Seems a bit weird to have
v-card
-- v-row
-- p
-- etc

instead of
v-card
-- v-card-text
---- v-row
---- p
---- etc

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
...
+  private expanded: Array<NameCheckItemIF> = []
+  private headers = [
+    { text: 'Icon', value: 'icon' },
+    { text: 'Problem', value: 'problem' },
+    { text: 'Words', value: 'words' },
+    { text: '', value: 'data-table-expand' }
+  ]
+
+  get ERROR_TYPES () {
+    return NameCheckErrorType
```
severinbeauvais, line 224: Unless you really want to use all-caps enums, another way to have the enum available in the HTML is like this:
https://github.com/bcgov/namerequest/blob/ae1da6cd7e7aa9dda104ce524fbbc9e7d865c135/src/components/common/applicant-info-1.vue#L400

src/components/new-request/name-check/name-check-conflicts.vue
```diff
@@ -0,0 +1,320 @@
...
+  }
+  get ITEM_TYPES () {
+    return NameCheckItemType
+  }
+
+  @Watch('expanded')
+  private triggerChange () { }
```
severinbeauvais, line 231: Obsolete?

src/components/new-request/name-check/name-check-tab-content.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <v-row class="check-tab-content" justify="center" no-gutters>
+    <v-col>
+      <span class="check-tab-title">{{ title }}</span>
+      <div v-if="loading" class="pt-2" no-gutters>
```
severinbeauvais, line 5: Does "no-gutters" do anything on a div?

src/components/new-request/name-check/quick-search-names.vue
```diff
@@ -0,0 +1,41 @@
...
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator'
+import MainContainer from '@/components/new-request/main-container.vue'
+
+@Component({
+  components: { MainContainer }
+})
+export default class QuickSearchNames extends Vue {
+  // Global getter
+  @Prop() namesList: Array<{ name: string, type: string }>
```
severinbeauvais, line 23: Seeing that you're expecting a specific value(s) for **type** (eg, "exact"), could you turn this into an enum?

src/components/new-request/name-check/resources/name-check-base-items-conflicts.ts
```diff
@@ -0,0 +1,145 @@
+import { NameCheckErrorType, NameCheckItemType } from '@/enums'
+
+export const baseItemsConflicts = {
+  'errorExact': {
```
severinbeauvais, line 4: I'd love to see an enum used for these keys.

src/components/new-request/name-check/name-check-conflicts.vue
cameron-freshworks: I think you can do without the templating, ie `{{ line }}` if that property is already contained in your raw html. 
Sorry, i didn't look into the html that you were applying. 

ie `<p v-for="line, index in item.expandedInfoBlock1" :key="index" class="ma-0" v-html="line"></p>` might work.

src/components/new-request/name-check/resources/name-check-base-items-structure.ts
```diff
@@ -0,0 +1,316 @@
+import { NameCheckErrorType, NameCheckItemType } from '@/enums'
+
+export const baseItemsStructure = {
+  'conflictsConditional': {
```
severinbeauvais, line 4: Add enum for this key?

src/components/new-request/name-check/resources/name-check-base-items-structure.ts
```diff
@@ -0,0 +1,316 @@
...
+    icon: 'mdi-alert',
+    iconColor: 'orange darken-2',
+    problem: 'You may need consent to use the following words:',
+    words: [],
+    expandedInfo1: 'Names that include well-known Names ' +
+    '(trade name and Trademarks), names of public figures, or reference ' +
```
severinbeauvais, line 10: I think our code style is to indent n-th lines. Is the linter OK with this?

src/store/getters.ts
cameron-freshworks: That would be nice! I need to get some more experience with composition-api ...😁

src/components/new-request/name-input.vue
```diff
@@ -1,35 +1,23 @@
...
-    </v-text-field>
-  </v-col>
+  <v-row no-gutters>
+    <v-col cols="12" class="pa-0">
+      <v-text-field :error-messages="message"
+                    @input="clearErrors()"
+                    @keydown.enter="handleSubmit"
+                    autocomplete="chrome-off"
+                    :filled="!isReadOnly"
+                    id="name-input-text-field"
```
severinbeauvais, line 9: Will this component be instantiated only once at a time? If not then this will create a duplicate ID browser warning; in that case use a classname instead.

src/components/new-request/search.vue
```diff
@@ -135,69 +135,32 @@
...
+             :cols="showDesignationSelect || (getIsXproMras) ? '8' : '12'">
         <NameInput v-if="!isFederal"
                    :class="inputCompClass"
                    :is-mras-search="(getIsXproMras && !noCorpNum)"
                    id="name-input-component"
                    class="pa-0"
                    @emit-corp-num-validity="corpNumValid = $event"/>
-        <p v-else class="pl-3 mt-n2 text-body-2">Federally incorporated businesses do not need a Name Request. You may
+        <p v-else class="pl-3 text-body-2">Federally incorporated businesses do not need a Name Request. You may
           register  your extraprovincial business immediately using its existing name at Corporate Online.</p>
```
severinbeauvais, line 147: Ask the UX folks if they want a link to Corp Online here.

src/components/new-request/search.vue
```diff
@@ -342,6 +295,24 @@ export default class NewSearch extends Vue {
   }
 
   // Local Getters
+  get isPremium () {
+    return JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.accountType === 'PREMIUM'
```
severinbeauvais, line 299: Would be great to have a const or enum for account type.

src/components/new-request/search.vue
```diff
@@ -342,6 +295,24 @@ export default class NewSearch extends Vue {
...
+  }
+
+  get designation () {
```
severinbeauvais, line 302: This return a string?

src/enums/name-check-error-type.ts
```diff
@@ -0,0 +1,7 @@
+export enum NameCheckErrorType {
+  errorDesignation = 'errorDesignation',
```
severinbeauvais, line 2: I think we usually use all-caps for the enum key.

src/interfaces/models.ts
```diff
@@ -1,5 +1,7 @@
 import { EntityType, Location, NrState, RequestCode } from '@/enums'
 
+export type AnalysisJurisdictionT = 'BC' | 'XPRO'
+export type AnalysisTypeT = 'designation' | 'structure'
```
severinbeauvais, line 4: I replaced other instances of these in my current ticket. These suck, because you can't use "constants" (enums) to assign or compare their values -- you're back to comparing strings. So I'd prefer to see enums for these.

src/interfaces/name-check-interfaces/restricted-response-interface.ts
```diff
@@ -0,0 +1,15 @@
+export interface RestrictedResponseIF {
+  restricted_words_conditions: Array<{
+    cnd_info: Array<{
+      allow_use: string,
```
severinbeauvais, line 4: Boolean?

src/interfaces/name-check-interfaces/restricted-response-interface.ts
```diff
@@ -0,0 +1,15 @@
...
+      consent_rrequired: string,
```
severinbeauvais, line 5: Typo?
Boolean?

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
+<template>
+  <v-container id="name-check-main-container" class="white pa-0 rounded">
+    <name-check-issues-dialog :attach="'#app'"
```
severinbeauvais, line 3: I think you can just use `attach="#app"` here.

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
...
+                              :display="showNameCheckIssuesDialog"
+                              :options="dialogOptions"
+                              @proceed="dialogSubmit"/>
```
severinbeauvais, line 6: I like to see parentheses on the method (just to be explicit and easier to recognize).

src/components/dialogs/name-check-issues.vue
kialj876: sure I can try changing that. Not sure if that'll mean cascading changes with v-card-text styling though. I might just take that out entirely if needed

kialj876: the reason it's like that is to separate the icons at the top from the text grid so that it displays correctly when the screen changes size

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: oh that's much better, thank you

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: no I need it otherwise the change doesn't trigger after it's been mounted

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
...
+        </v-col>
+      </v-row>
+      <v-row no-gutters class="pa-10 pt-7">
+        <v-col>
+          <v-btn @click="back()" class="outlined" outlined>
+            Start Over
+          </v-btn>
+        </v-col>
+        <v-col>
+          <v-btn id="name-check-submit-btn" style="float: right;" @click="dialogCheck()">
```
severinbeauvais, line 185: Try Vuetify class "float-right" instead of inline style.

https://vuetifyjs.com/en/styles/float/#classes

src/components/new-request/name-check/name-check-tab-content.vue
kialj876: nope - will update (this was a v-row at one point haha)

src/components/new-request/name-check/name-check.vue
```diff
@@ -0,0 +1,809 @@
...
+  config = {
+    modules: {
+      toolbar: false
+    },
+    placeholder: '',
+    scrollingContainer: true
+  }
+
+  private checks: string = null
+  private expandHelpTxt: boolean = false
```
severinbeauvais, line 261: Where the type can be inferred from its initial value, we should omit the type. This is because some linters will complain about "redundant type declaration".

src/components/new-request/name-check/quick-search-names.vue
kialj876: yeah I can haha. Also I will update that type to the interface

src/components/new-request/name-check/resources/name-check-base-items-conflicts.ts
kialj876: sure thing

src/components/new-request/name-check/resources/name-check-base-items-structure.ts
kialj876: will do

src/components/new-request/name-check/name-check.vue
severinbeauvais: If you need to leave this in, this probably needs a lint disable comment.

src/components/new-request/name-check/resources/name-check-base-items-structure.ts
kialj876: linter likes it. I'd rather not go through and tab them all since there's so many and probably won't be changed / looked at much in the future

src/components/new-request/name-check/name-check.vue
severinbeauvais: Or just a readonly string property?

src/store/actions.ts
```diff
@@ -883,3 +639,433 @@ export const setStaffPayment: ActionIF = ({ commit }, staffPayment: StaffPayment
 export const setFolioNumber: ActionIF = ({ commit }, folioNumber: string): void => {
   commit('mutateFolioNumber', folioNumber)
 }
+/** Name Check actions
+ * TODO: move these into a factory if converting to composition api
```
severinbeauvais, line 643: If you're not going to do this in this ticket/PR, I suggest you tag this with "FUTURE" instead 😁

src/store/actions.ts
```diff
@@ -883,3 +639,433 @@ export const setStaffPayment: ActionIF = ({ commit }, staffPayment: StaffPayment
...
+  try {
+    if (options.designationOnly) commit('mutateAnalyzeDesignationPending', true)
+    else commit('mutateAnalyzeStructurePending', true)
+    commit('resetRequestExaminationOrProvideConsent')
+    const params: NewRequestNameSearchI = {
+      name: name,
+      location: getters.getLocation,
+      entity_type_cd: getters.getEntityTypeCd,
+      request_action_cd: getters.getRequestActionCd,
+      jurisdiction: options.xpro ? 'XPRO' : 'BC',
```
severinbeauvais, line 703: Use enum values.

src/store/actions.ts
```diff
@@ -883,3 +639,433 @@ export const setStaffPayment: ActionIF = ({ commit }, staffPayment: StaffPayment
...
+      analysis_type: options.designationOnly ? 'designation' : 'structure'
```
severinbeauvais, line 704: Use an enum for this type?

src/store/actions.ts
```diff
@@ -883,3 +639,433 @@ export const setStaffPayment: ActionIF = ({ commit }, staffPayment: StaffPayment
...
+    // verify the user did not start a new search on a different name
+    if (name === getters.getFullName) {
+      const json = analysis
+      commit('mutateAnalysisJSON', json)
+      if (Array.isArray(json.issues) && json.issues.length > 0) {
+        for (let i = 0; i < json.issues.length; i++) {
+          switch (json.issues[i].issue_type) {
+            case 'add_descriptive':
+              commit('mutateMissingDescriptive', true)
+              continue
```
severinbeauvais, line 716: On the last "for" item, what does `continue` do? Just drop out of the for loop?

severinbeauvais (approved):
Looks really good. Good use of coding best practices (types, naming, comments, etc), Makes it easy to review :)

Just a few small niggles to make future dev work nicer, please and thanks.

~Q: Would it make sense to add a FF to bypass some of this for any reason?~ _Update: see comment below_

src/components/dialogs/name-check-issues.vue
severinbeauvais: It's just an expectation of not mixing row/col and card blocks.

severinbeauvais: Maybe v-card has other blocks you can use?

https://vuetifyjs.com/en/components/cards/#usage

src/components/new-request/name-check/name-check-conflicts.vue
severinbeauvais: Is this something that `{ immediate: true }` could solve?

If you keep this (with an empty body), DEFINITELY add a comment, otherwise some over-zealous programmer might delete this.

src/components/new-request/name-check/name-check.vue
kialj876: sure sounds good

src/components/new-request/name-input.vue
kialj876: hmm okay good point will update

src/components/new-request/name-check/quick-search-names.vue
severinbeauvais: Yes, I meant the type.

Well, the name, too, if it's a set of possible values :)

src/components/new-request/name-check/resources/name-check-base-items-structure.ts
severinbeauvais: OK, nvm

src/components/new-request/search.vue
kialj876: sure thing will update

src/components/new-request/search.vue
kialj876: ha - too late there's no time, could be an enhancement added in the UI ticket for updating this component

src/enums/name-check-error-type.ts
kialj876: will update

src/interfaces/models.ts
kialj876: sure thing

src/interfaces/name-check-interfaces/restricted-response-interface.ts
kialj876: nope -- it's from old namex-api code

src/interfaces/name-check-interfaces/restricted-response-interface.ts
kialj876: ah yes typo, but this value is never actually used by the UI

src/components/new-request/search.vue
severinbeauvais: ok

src/components/new-request/name-check/name-check.vue
kialj876: ha, yes

src/components/new-request/name-check/name-check.vue
kialj876: sure

src/interfaces/name-check-interfaces/restricted-response-interface.ts
severinbeauvais: ok

src/components/new-request/name-check/name-check.vue
kialj876: I knew there was something easier

src/components/new-request/name-check/name-check.vue
kialj876: oh okay sounds good

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: oh cool okay yes I will try that

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: Okay good point haha

src/components/dialogs/name-check-issues.vue
kialj876: Ah okay yes I will update it so I'm not mixing them

src/components/new-request/name-check/name-check.vue
kialj876: will remove

src/components/new-request/name-check/name-check-conflicts.vue
kialj876: Okay I am adding a comment. I'm not sure why it doesn't act reactive otherwise

src/components/new-request/name-check/quick-search-names.vue
kialj876: ha, the name can be anything, but I will add it for the type. I meant the interface for that object (it's used in actions)

src/store/actions.ts
kialj876: yes it was unnecessary

src/components/dialogs/name-check-issues.vue
severinbeauvais: Looks good!

src/components/new-request/name-check/resources/name-check-base-items-structure.ts
severinbeauvais: Looks good.

src/interfaces/models.ts
severinbeauvais: Hooray!

src/store/actions.ts
severinbeauvais: Oh no, that's not what I meant. (I was wondering what happens after the last iteration -- does the code execute after the `continue`? But I think it actually goes to the code after the for block.) So my original comment is not an issue.

Removing the `continue` in the default block is OK since that's the end of the for block anyway, so it will "continue" with any other for iterations, as desired.

src/components/new-request/name-check/name-check.vue
severinbeauvais: I confirmed this is removed, so resolving this conversation.

src/components/new-request/name-check/name-check-conflicts.vue
severinbeauvais: I confirm this is implemented, so I'm resolving this conversation.

src/components/new-request/name-check/quick-search-names.vue
severinbeauvais: Is this not fixable, or future?

src/components/new-request/name-check/resources/name-check-base-items-conflicts.ts
severinbeauvais: Looks good!

src/components/new-request/name-input.vue
severinbeauvais: Future?

src/enums/name-check-error-type.ts
severinbeauvais: Thanks.

(The reason we use all-caps is because these are essentially constants, which are traditionally all caps to distinguish them from regular variables.)

src/store/actions.ts
severinbeauvais: Thanks. This makes is easier to understand what "type" this field is (as well as managing what values we can assign).

👍

severinbeauvais:
> Q: Would it make sense to add a FF to bypass some of this for any reason?

Not needed?

kialj876:
> > Q: Would it make sense to add a FF to bypass some of this for any reason?
> 
> Not needed?

Na I don't think so. The worst case for things erring everywhere on the backend is covered by the code and the auto analyze distinctive/descriptive call is covered by an LD flag (this is the only call that could potentially take more than a couple seconds). Even if it took too long the design / code handles it so I don't see it being turned off unless it was +30 seconds for everyone (in dev its about 2 or 3 seconds usually)

kialj876:
fyi I am planning to merge this after the release tomorrow (just in case they need to put any hot fixes through)

PR #426, opened 2021-07-15
8164 Name Request edit sync issues

cameron-freshworks:
*Issue #:* /bcgov/entity#8164

*Description of changes:*

* Fixed some case sensitive fields syncing between NamEx UI -> Name Request UI.
* Removed Jurisdiction field being forced on MVE BC requests during an Edit.
* Simplified some jurisdiction getter logic.
* Made CorpNum optional for Amalgamation request types, this is work related to ticket /bcgov/entity#5941.


src/components/common/applicant-info-1.vue
```diff
@@ -390,6 +390,9 @@ import { ActionMixin } from '@/mixins'
 import { ActionBindingIF } from '@/interfaces/store-interfaces'
 import { NameRequestI } from '@/interfaces'
 
+// List Data
+import { CanJurisdictions, IntlJurisdictions } from '@/list-data'
```
severinbeauvais, line 394: You should have these in `this` already. See https://github.com/bcgov/namerequest/blob/ae1da6cd7e7aa9dda104ce524fbbc9e7d865c135/src/main.ts#L46.

cameron-freshworks: ah, sorry. I thought it was preferential to go without the globally scoped variables. 
Was i thinking backwards here?

src/components/common/applicant-info-1.vue
```diff
@@ -483,10 +486,10 @@ export default class ApplicantInfo1 extends Mixins(ActionMixin) {
 
   get jurisdictionOptions () {
     return (this.getLocation === Location.CA)
-      ? this.$canJurisdictions.filter(jur => jur.value !== Location.BC)
-        .map(jurisdiction => ({ value: jurisdiction.text, text: jurisdiction.text }))
-      : this.$intlJurisdictions.filter(jur => jur.value !== Location.CA)
-        .map(jurisdiction => ({ value: jurisdiction.text, text: jurisdiction.text }))
+      ? CanJurisdictions.filter(jur => jur.value !== Location.BC)
+        .map(jurisdiction => ({ value: jurisdiction.text.toUpperCase(), text: jurisdiction.text }))
```
severinbeauvais, line 490: What's the purpose of this? It will capitalize all keys in the v-select, right? Why have text in Title Case in the first place?

src/store/getters.ts
```diff
@@ -505,9 +505,7 @@ export const getEntityTypesXPRO = (state: StateIF): EntityI[] => {
 }
 
 export const getShowXproJurisdiction = (state: StateIF): boolean => {
-  if (getLocation(state) !== Location.BC) return true
-  if (getRequestActionCd(state) === RequestCode.MVE) return true
```
severinbeauvais, line 509: Why is it OK to remove this MVE check? Should this be commented in the ticket as something for QA to verify?

severinbeauvais: Never mind - I read your description of changes again. Looks fine.

src/components/common/applicant-info-1.vue
cameron-freshworks: This will only capitalize the value, as the display text is not full caps. (ie `value: ALBERTA, text: Alberta`)
This is because from Name Examination the Jurisdiction key value is capitalized.

The other option is changed the jurisdiction value properties to all capitalized just not sure of all the implications outside of this context.

src/store/getters.ts
cameron-freshworks: I confirmed with Linda this should never have been required in the EDIT flow for a MVE request. 

I believe, the initial implementation of MVE requests were thought of as XPRO but the flow changed to a BC request when creating a new request. I believe this was left over from that change.

cameron-freshworks: We also verified the same behaviour in Name Examination and NRO.

cameron-freshworks: Will add a comment to the ticket for QA, as to what to look for.

src/components/common/applicant-info-1.vue
severinbeauvais: In the spirit of "starting to do things the better way", your code is the preferred approach (ie, only import what you need). This will be one less place we have to fix when we eventually get rid of the globally scoped variables :)

As you were.

src/components/common/applicant-info-1.vue
severinbeauvais: Gotcha.

But the thing that bugs me here is that it's unclear which code needs what or what value is at any one time. I'd be tempted to add a property to every juridiction object named `namexValue` that is the upper case value. It's explicit and more clear IMO. What do you think?

cameron-freshworks: I like the idea actually! 
Although i might prefer to change the `value` to be capitalized so it's in sync with Name Ex, followed by exploring/fixing any implications of that, is that preferable vs a new property? 
Only concern was we would have the same value 3 times with just varying capitalization.

severinbeauvais: Yes, just making it all uppercase is preferable... except it will change the v-select items, won't it?

It's not worth a ton of time deliberating over this. Less duplication is usually better, except when it makes things harder to understand. Your call.

cameron-freshworks: Ah i see what you mean by an additional property now. 
Yea, changing this value would reformat the text in the list as well :/ 
I think i prefer converting to uppercase to match the model vs additional properties but i can see it is a little confusing so i added a comment!

severinbeauvais: I'm OK with that.

severinbeauvais (approved):

cameron-freshworks:
@severinbeauvais I added another commit that should be reviewed :) Sorry and thank you, no rush!

severinbeauvais:
Is there another ticket to link in the description?

cameron-freshworks:
> Is there another ticket to link in the description?

Indeed! Updated in comments above, thank you!

severinbeauvais (approved):

PR #428, opened 2021-07-22
name-check updates

kialj876:
*Issue #:* /bcgov/entity#8261

*Description of changes:*
- name check enhancements

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -22,6 +22,7 @@
 
       <!-- SBC Common Components header -->
       <SbcHeader
+        id="namerequest-sbc-header"
         class="sbc-header"
         :inAuth="false"
         :showActions="true"
--- a/src/App.vue
+++ b/src/App.vue
@@ -23,7 +23,6 @@
       <!-- SBC Common Components header -->
       <SbcHeader
         id="namerequest-sbc-header"
-        class="sbc-header"
         :inAuth="false"
         :showActions="true"
       />
@@ -225,11 +224,16 @@ export default class App extends Mixins(DateMixin) {
   display: flex;
   flex-flow: column nowrap;
   min-height: 100vh;
+}
 
-  .sbc-header {
-    .v-btn {
-      box-shadow: none !important;
-    }
+#namerequest-sbc-header {
+  .v-btn {
+    box-shadow: none !important;
+    background-color: rgba(0, 0, 0, 0) !important;
+    text-transform: none !important;
+    font-weight: normal !important;
+    letter-spacing: normal !important;
+    font-size: .875rem !important;
   }
 }
 
--- a/src/components/new-request/name-check/name-check.vue
+++ b/src/components/new-request/name-check/name-check.vue
@@ -42,9 +42,7 @@
         </v-col>
       </v-row>
       <div id="name-check-verdict" class="white mt-7 pa-2" no-gutters justify="center">
-        <v-progress-circular v-if="isAnalyzeConflictsPending ||
-                                   isAnalyzeDesignationPending ||
-                                   isAnalyzeStructurePending"
+        <v-progress-circular v-if="loadingStructureCheck"
                              color="primary"
                              indeterminate
                              size="30"
@@ -58,8 +56,7 @@
       </div>
       <v-container v-if="hasIssuesConflictAlert || hasIssuesConflictCaution ||
                          hasIssuesStructureAlert || hasIssuesStructureCaution ||
-                         isAnalyzeConflictsPending || isAnalyzeDesignationPending ||
-                         isAnalyzeStructurePending"
+                         loadingStructureCheck"
                    id="name-check-tabs-container"
                    class="ma-0 pt-7">
         <v-tabs no-gutters
@@ -77,9 +74,7 @@
                  class="upper-border px-0 pt-3"
                  :ripple="false">
             <div style="width: 100%;">
-              <name-check-tab-content :loading="isAnalyzeDesignationPending ||
-                                                isAnalyzeStructurePending ||
-                                                isAnalyzeConflictsPending"
+              <name-check-tab-content :loading="loadingStructureCheck"
                                       :subtitle="subTitleStructure"
                                       :tabIcon="tabIconStructure.icon"
                                       :title="'Name Structure Check'"/>
@@ -100,7 +95,7 @@
           </v-tab>
         </v-tabs>
         <div id="box-shadow-filler-tab-item"/>
-        <v-tabs-items class="rounded-b tab-items pa-5"
+        <v-tabs-items class="rounded-b tab-items pa-5 pt-6"
                       :class="checks === 'structure-check' ? 'rounded-top-right' : 'rounded-top-left'"
                       v-model="checks">
           <v-tab-item value="structure-check">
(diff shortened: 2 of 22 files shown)


src/components/new-request/name-check/name-check.vue
```diff
@@ -220,6 +215,7 @@ import { baseItemsConflicts, baseItemsStructure } from './resources'
 })
 export default class NameCheck extends Vue {
   @Getter getConflictsConditional!: Array<string>
+  @Getter getConflictsConditionalInstructions!: Array<{ word: string, instructions: string }>
```
cameron-freshworks, line 218: Although a simple interface, it could be made into it's own in the event you need to apply type or a return type based on this getter :)

src/components/new-request/name-check/name-check.vue
```diff
@@ -316,6 +318,9 @@ export default class NameCheck extends Vue {
   get conflictsConditional (): Array<string> {
     return this.getConflictsConditional
   }
+  get conflictsConditionalInstructions (): Array<{ word: string, instructions: string }> {
```
cameron-freshworks, line 321: Same type as above :)

src/components/new-request/name-check/name-check.vue
```diff
@@ -634,7 +661,7 @@ export default class NameCheck extends Vue {
       return { color: 'red darken-2', icon: 'mdi-alert-octagon' }
     }
     if (this.hasIssuesConflictCaution || this.hasIssuesStructureCaution) {
-      return { color: 'orange darken-2', icon: 'mdi-alert' }
+      return { color: '#F8661A', icon: 'mdi-alert' }
```
cameron-freshworks, line 664: looks like there is an app colour defined for this already called `$caution`

src/components/new-request/search.vue
```diff
@@ -297,6 +298,10 @@ export default class NewSearch extends Vue {
     })
   }
 
+  private activated () {
+    this.scrollTo('namerequest-sbc-header')
+  }
+
```
cameron-freshworks, line 304: Is this called in the component?

src/interfaces/name-check-interfaces/name-check-model-interface.ts
```diff
@@ -6,6 +6,7 @@ export interface NameCheckModelIF {
     analyzeStructurePending: boolean
     conflictsExact: Array<string>
     conflictsConditional: Array<string>
+    conflictsConditionalInstructions: Array<{ word: string, instructions: string }>
```
cameron-freshworks, line 9: I know tedious and simple, might be best to use an interface for all instances of this object.

src/App.vue
```diff
@@ -22,6 +22,7 @@
 
       <!-- SBC Common Components header -->
       <SbcHeader
+        id="namerequest-sbc-header"
         class="sbc-header"
```
severinbeauvais, line 26: Could you clean this up please?
- remove class
- change selector in local (scoped) style block from class to id
- move sbc-header styles from overrides.scss to local style block

cameron-freshworks (approved):
Looks good, just suggesting updating the hex colour usage as well as an interface for `conditionalInstructions` 👍

src/components/new-request/name-check/name-check.vue
```diff
@@ -109,18 +104,18 @@
                 <v-icon>mdi-information-outline</v-icon>
               </v-col>
               <v-col class="pl-2">
-                <span v-html="infoTextTab">{{ infoTextTab }}</span>
+                <span v-html="infoTextTab"/>
```
severinbeauvais, line 107: Add space before /.
Also on line 118.

src/components/new-request/name-check/name-check.vue
```diff
@@ -579,6 +603,9 @@ export default class NameCheck extends Vue {
     }
     return 'You can edit your name here and check it again'
   }
+  get numbersCheckUse (): Array<string> {
```
severinbeauvais, line 606: I think you don't need this. Just use `this.getNumbersCheckUse` where it's used.

src/components/new-request/name-check/name-check.vue
severinbeauvais, line 664: This hex code is already defined as `$caution`, but that's not a Vuetify theme color... please consider adding it (see vuetify.ts), or maybe replacing `warning` and then using that name here.

(It's bad practice to use hex colors in code, and good practice to have a defined palette for an app.)

src/components/new-request/name-check/resources/name-check-base-items-conflicts.ts
```diff
@@ -100,7 +100,7 @@ export const baseItemsConflicts = {
   },
   [NameCheckItemType.SIMILAR_MATCH]: {
     icon: 'mdi-alert',
-    iconColor: 'orange darken-2',
+    iconColor: '#F8661A',
```
severinbeauvais, line 103: Again, use named color here (and on line 119 and in other resource files).

src/components/new-request/name-check/name-check.vue
kialj876: how do I access that variable here?

src/interfaces/name-check-interfaces/parsed-restricted-response-interface.ts
```diff
@@ -1,4 +1,5 @@
 export interface ParsedRestrictedResponseIF {
+  conditionalInstructions: Array<{ word: string, instructions: string }>
```
severinbeauvais, line 2: This type is used in several places now... Please add an interface for it :)

src/components/new-request/search.vue
kialj876: no it's a life cycle hook isn't it? I think because of the way the Tabs work it gets triggered? I tried this and it worked, but I've never used it before haha

src/App.vue
severinbeauvais: Or... add a common mixin method to "scrollByClassname".

src/interfaces/name-check-interfaces/name-check-model-interface.ts
kialj876: haha yes I will add that in

src/App.vue
severinbeauvais: Meh, "scroll by id" is better since ids should be unique.

Another idea: scroll to "app" id instead of the sbc header???

src/components/new-request/name-check/name-check.vue
kialj876: ah okay thanks, I was wondering about this. I will replace 'warning'

src/components/new-request/search.vue
cameron-freshworks: Interesting!! I've never used it, i'll look it up :) Thanks!

src/App.vue
kialj876: yeah sure

src/components/new-request/name-check/name-check.vue
kialj876: nvm severin mentioned changing the vuetify 'warning' in override -- will do that

severinbeauvais: Add it to Vuetify theme, or replace warning colour. See vuetify.ts.

src/interfaces/name-check-interfaces/parsed-restricted-response-interface.ts
severinbeauvais: Resolving this convo since Cam already mentioned it.

src/components/new-request/name-check/name-check.vue
kialj876: right sorry that's what I meant

src/components/new-request/search.vue
severinbeauvais: Hmm, I don't see it here:
https://vuejs.org/v2/guide/instance.html#Lifecycle-Diagram

You might want to check where it's called in the order of things. Eg, is it only called once? Has the component been created? Rendered?

src/App.vue
kialj876: I did try scrolling to "app" but unfortunately it would go to the centre of it which is below the top page since it encompasses everything

src/components/new-request/name-check/name-check.vue
kialj876: I ended up creating a new theme: 'caution' because warning is used for the banner

src/components/new-request/search.vue
kialj876: https://vuejs.org/v2/api/#activated -- not a lifecycle hook I guess

kialj876: well it is a custom lifecycle hook for keep-alive I guess?

severinbeauvais (approved):
Looks good.
I'm ok with the `activated` hook as long as it works as intended :)

PR #430, opened 2021-07-28
7960 search combo update

kialj876:

*Issue #:* /bcgov/entity#7960

*Description of changes:*
- small updates to search combo

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -676,7 +676,7 @@ export const getMatchesSimilar = async (
   return synonymResp?.data ? parseSynonymNames(synonymResp.data) : []
 }
 export const getMatchesRestricted = async (
-  { commit },
+  { commit, getters },
   token: string,
   cleanedName: string
 ): Promise<ParsedRestrictedResponseIF> => {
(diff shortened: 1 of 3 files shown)


src/store/actions.ts
```diff
@@ -853,31 +853,46 @@ export const parseExactNames = (json: { names: [string] }): Array<ConflictListIt
...
   // restrictedWords: add any word with cnd_info[..].allow_use = N
   // conditionalWords: all other words in the list
-  for (let i = 0; i < words.length; i++) {
+  for (let i = 0; i < phrases.length; i++) {
+    const phrase = phrases[i].word_info.phrase.toUpperCase()
+
+    // rules for ignoring restricted/conditional phrases
+    const entityCd = getters.getEntityTypeCd
+    if (entityCd === EntityType.CR && Designations[EntityType.CC].words.includes(phrase)) continue
+    if (Designations[entityCd].words.includes(phrase)) continue
```
cameron-freshworks, line 871: I've never really used `continue` in this context. 
Will this `return` if the condition is NOT met?

kialj876: yeah it just means the loop will go to the next one (so all the code below it will be ignored if the condition is met and the loop will continue)

cameron-freshworks: Okie doke!

cameron-freshworks (approved):
Looks good!

severinbeauvais (commented):
Looks good. The only thing I note is that it's nice to update the app version so we can confirm what version is deployed in Dev/Test/Prod.

PR #431, opened 2021-07-29
8221

severinbeauvais:
*Issue #:* /bcgov/entity#8221

*Description of changes:*

**This is no riskier than the existing resubmit code in this UI... so I think we can merge it if OK. This may make it easier for Andre to test his API changes. I can always submit another PR for any further changes. -SB**

- added some base types
- added Priority checkbox to Resubmit dialog
- re-fetch fees when Priority changes
- call resubmit() before creating payment
- cancel edit mode where needed
- deleted space chars around some additional info messages
- implement resubmit() action
- misc cleanup
- updated app version


src/components/dialogs/resubmit.vue
```diff
@@ -51,6 +51,14 @@
               <FeeSummary
                 :fees="[...paymentFees]"
               />
+
+              <v-checkbox
+                hide-details
+                v-model="isPriorityRequest"
+                class="pre-wrap mt-8 pt-0 pl-2"
```
severinbeauvais, line 58: Note: need "pre-wrap" class to retain space char between "Request" and "($100)"

src/components/dialogs/resubmit.vue
```diff
@@ -51,6 +51,14 @@
...
+              >
+                <template v-slot:label>Make this a Priority Request <b>($100)</b></template>
```
severinbeauvais, line 60: Note: "label" slot is how you enter an HTML label

src/components/dialogs/resubmit.vue
```diff
@@ -204,11 +223,12 @@ export default class ResubmitDialog extends Mixins(
       }
     }
 
-    const success = await this.createPayment({
+    // if resubmit succeeded then create the payment
+    success = success && await this.createPayment({
       action: PaymentAction.RESUBMIT,
       nrId: this.getNrId,
       filingType: FilingTypes.NM620,
```
severinbeauvais, line 230: Outstanding question: should this be type NM620 or some other type for resubmit?

src/components/existing-request/existing-request-display.vue
```diff
@@ -2,7 +2,7 @@
   <MainContainer id="existing-request-display" class="pa-10">
     <template v-slot:container-header>
       <v-col cols="auto" class="py-0">
-        <span class="h3">{{ nr.nrNum }}</span>
+        <span class="h3 user-select-all">{{ nr.nrNum }}</span>
```
severinbeauvais, line 5: This CSS rule allows the user to click on this text and have it all auto-selected. Otherwise, if you double-click on the NR number, the next word is also selected (undesired).

src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -71,22 +71,18 @@ export default class ReserveSubmitButton extends Vue {
     this.cancelAnalyzeName('NamesCapture')
   }
 
-  getData (type: string): any {
-    if (this.getAssumedName) type = 'assumed'
-    let data: any
+  getData (type: NrType): DraftReqI | ConditionalReqI | ReservedReqI {
```
severinbeauvais, line 74: I cleaned this up when I intended to use this method to get "RESUBMIT" data, but then I found a better way.

src/components/payment/fee-summary.vue
```diff
@@ -20,22 +20,26 @@
...
+                <v-expand-transition>
+                  <v-row v-if="lineItem.priorityFees" no-gutters>
+                    <v-col cols="8">
+                      <div>Priority Request fee</div>
+                    </v-col>
+                    <v-col cols="4">
+                      <div class="float-right">${{lineItem.priorityFees.toFixed(2)}}</div>
+                    </v-col>
+                  </v-row>
+                </v-expand-transition>
```
severinbeauvais, line 32: I added this transition (where the HTML is conditionally displayed) so the fee summary grows/shrinks instead of updating abruptly.
I updated the service fees below as well since it also has a v-if.

src/enums/request-code.ts
```diff
@@ -11,4 +11,5 @@ export enum RequestCode {
   REN = 'REN', // restore with new name request (renew)
   INFO = 'INFO', // special value for sub-menu
   REST = 'REST', // restore
+  RESUBMIT = 'RESUBMIT', // resubmit
```
severinbeauvais, line 14: @andrepestana-aot FYI!

src/mixins/displayed-component-mixin.ts
```diff
@@ -41,7 +42,8 @@ export class DisplayedComponentMixin extends Vue {
       await this.cancelAnalyzeName('Tabs')
       this.redirectToStart()
     } else if (['ExistingRequestEdit'].indexOf(componentName) > -1) {
-      // We're editing
+      // Cancel edit mode
+      await this.setEditMode(false)
```
severinbeauvais, line 46: This is called if the user clicks Return or Exit during edit mode (or anytime probably, in which case this does nothing).

src/services/namex.services.ts
```diff
@@ -75,7 +75,7 @@ export default class NamexServices {
 
       // if there is no previous request_action message then
       // we just preserve whatever text there is and append msg
-      data['additionalInfo'] += ` \n\n ${msg}`
+      data['additionalInfo'] += `\n\n${msg}`
```
severinbeauvais, line 78: There is other code that adds messages to the additional info, with double carriage returns, but without extra spaces. So I removed them here to be consistent.

src/store/actions.ts
```diff
@@ -219,8 +220,8 @@ export const cancelEditExistingRequest: ActionIF = ({ commit }) => {
   commit('mutateEditMode', false)
 }
 
-export const editExistingRequest: ActionIF = ({ commit, getters }) => {
-  commit('mutateEditMode', true)
+// called to commit data into "newRequestModel" object
+const commitExistingData = ({ commit, getters }) => {
```
severinbeauvais, line 224: I created this to reuse the code in editExistingRequest() and resubmit().

src/store/actions.ts
```diff
@@ -369,22 +375,18 @@ export const getNrStateData = ({ getters }) => {
 }
 
 // FUTURE: not an action - move it to another module?
-export const getNrTypeData = ({ getters }, type: string) => {
-  if (getters.getAssumedName) type = 'assumed'
-  let data: any
+export const getNrTypeData = ({ getters }, type: NrType): any => {
```
severinbeauvais, line 378: Cleanup again.

src/store/actions.ts
```diff
@@ -436,6 +444,69 @@ export const submit: any = async ({ commit, getters, dispatch }): Promise<any> =
   }
 }
 
+/**
+ * Re-submits an expired NR (without changing the current NR data - in case of failure).
```
severinbeauvais, line 448: It was really frustrating to have multiple store objects for a new NR or the current NR (for display)... To avoid messing up that code, this method works with local data and overrides things as needed so the POST data object looks like a new NR request.

src/store/actions.ts
```diff
@@ -436,6 +444,69 @@ export const submit: any = async ({ commit, getters, dispatch }): Promise<any> =
...
+  if (request) {
+    commit('setNrResponse', request)
+    return true
+  }
+
+  return false
+}
+
+/**
+ * Sometimes name objects have a null designation (because the designation is part of the name).
```
severinbeauvais, line 499: This is probably an unintended consequence of a previous change, @cameron-freshworks , should this be fixed?

src/store/actions.ts
```diff
@@ -644,10 +715,12 @@ export const setStaffPayment: ActionIF = ({ commit }, staffPayment: StaffPayment
 export const setFolioNumber: ActionIF = ({ commit }, folioNumber: string): void => {
   commit('mutateFolioNumber', folioNumber)
 }
-/** Name Check actions
+
+/**
+ * Name Check actions
  * FUTURE: move these into a factory if converting to composition api
  */
-export const getMatchesExact = async (
```
severinbeauvais, line 650: There are only used locally (and this isn't a class so I couldn't make them private), so I just stopped exporting these.

src/store/mutations.ts
```diff
@@ -394,8 +394,8 @@ export const setNrResponse = (state: StateIF, nr: NameRequestI) => {
   }
 }
 
-export const updateReservationNames = (state: StateIF, nrName: [] = []) => {
-  nrName.forEach(({ choice, name = '', designation = '' }) => {
+export const updateReservationNames = (state: StateIF, nrNames: [] = []) => {
+  nrNames.forEach(({ choice, name = '', designation = '' }) => {
```
severinbeauvais, line 398: I like when array variables are plural, and then you can iterate over their singular elements :)

src/components/dialogs/resubmit.vue
```diff
@@ -204,11 +223,12 @@ export default class ResubmitDialog extends Mixins(
...
-      priorityRequest: this.getPriorityRequest // same value as originally
+      priorityRequest: this.getPriorityRequest
     } as CreatePaymentParams, onSuccess)
```
andrepestana-aot, line 232: Let's see with Linda if we can do a `PaymentAction.CREATE` here instead. I think it would the best approach.

andrepestana-aot (commented):
LGTM. Let's wait to check if it will RESUBMIT or CREATE?

src/components/dialogs/resubmit.vue
```diff
@@ -160,6 +166,16 @@ export default class ResubmitDialog extends Mixins(
     }
   }
 
+  @Watch('isPriorityRequest')
+  async onPriorityRequestChange (): Promise<boolean> {
+    const params: FetchFeesParams = {
+      filingType: FilingTypes.NM620,
+      jurisdiction: Jurisdictions.BC,
```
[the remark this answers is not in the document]
severinbeauvais: I don't know, but clearly this allows for that possibility... However, BCROS is currently serving BC, so I don't think this will change.

src/components/dialogs/resubmit.vue
severinbeauvais: Can we stay with both CREATE and RESUBMIT actions? The API can always combine them when it calls the Pay API.

src/components/dialogs/resubmit.vue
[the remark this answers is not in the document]
cameron-freshworks: Just to add onto this, there is no other potential `filingTypes` either? Is the the sole code for Name Request Applications?

src/components/existing-request/existing-request-display.vue
cameron-freshworks: Cool!

src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -71,22 +71,18 @@ export default class ReserveSubmitButton extends Vue {
...
-      case 'draft':
-        data = this.getDraftNameReservation
-        break
-      case 'conditional':
-        data = this.getConditionalNameReservation
-        break
-      case 'reserved':
-        data = this.getReservedNameReservation
-        break
+      case NrType.ASSUMED:
```
cameron-freshworks, line 77: This may* be obsolete. I know it is an option to edit an NR to `assumed` via name examination ui but i'm not sure Name Request supports it anymore. I could be wrong tho!

cameron-freshworks: No action required, just thinking!

src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -140,7 +136,7 @@ export default class ReserveSubmitButton extends Vue {
 
       // @ts-ignore - typescript knows setup can only === 'assumed' at this point and gives error
       case 'consent':
-        data = this.getData('conditional')
+        data = this.getData(NrType.CONDITIONAL)
```
cameron-freshworks, line 139: 👍

src/components/payment/fee-summary.vue
```diff
@@ -20,22 +20,26 @@
...
-                <v-row v-if="lineItem.serviceFees" no-gutters>
-                  <v-col cols="8">
-                    <div>Service fee</div>
-                  </v-col>
-                  <v-col cols="4">
-                    <div class="float-right">${{lineItem.serviceFees.toFixed(2)}}</div>
-                  </v-col>
-                </v-row>
+                <v-expand-transition>
+                  <v-row v-if="lineItem.priorityFees" no-gutters>
```
cameron-freshworks, line 24: Are these properties in a loop where `lineItem` will always be defined?

src/enums/request-code.ts
```diff
@@ -1,4 +1,4 @@
-/** Enum for Request (action) codes. Not to be confused with NR Action codes. */
```
cameron-freshworks, line 1: I think this was added because `request` alone is vague as there is request type and request action etc..
No biggie, just some thoughts.

src/store/actions.ts
cameron-freshworks: I'm not too sure, i don't recall implementing this, I most likely copy pasted this over.
Not too sure of the implications of this.

cameron-freshworks (approved):
Looks good to me, just had a few questions/comments 👍

src/store/actions.ts
severinbeauvais: OK, I won't worry about it.

src/components/dialogs/resubmit.vue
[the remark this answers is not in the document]
severinbeauvais: I asked Linda about the Filing Type, and she confirmed that resubmit is to use NM620.

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/8221#issuecomment-889531008

src/components/new-request/submit-request/reserve-submit.vue
severinbeauvais: Yes, I think some of this is obsolete. In this method, only CONDITIONAL and RESERVED are ever called. In the similar method (getNrTypeData) in actions.ts, only DRAFT is ever called.
This could be cleaned up later, unless you really want me to clean it up now :)

src/components/payment/fee-summary.vue
severinbeauvais: Yes! See line 9.

src/enums/request-code.ts
severinbeauvais: The comment? Yes, I added it previously because I found the naming confusing. Is this OK or should I go back to being explicit? (I think it's much more clear now with the enums anyway.)

src/components/dialogs/resubmit.vue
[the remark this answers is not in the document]
severinbeauvais: > Just to add onto this, there is no other potential `filingTypes` either? Is the the sole code for Name Request Applications?

Tbh, I brought this up in a previous ticket (for Renew I think) and nobody knew much about these Filing Type codes, so I kinda feel like it's not that important. I would think the Pay guys might care but they didn't say much about it either.

src/components/new-request/submit-request/reserve-submit.vue
cameron-freshworks: No no, no biggie! 
Just encountered the same thing previously when looking at the Edits.

src/enums/request-code.ts
cameron-freshworks: For sure! I'm fine with this, was just sharing what i encountered when working in Namex UI :)

severinbeauvais: Is there something (easy) that I can do to make this sync up with Namex better?

src/components/dialogs/resubmit.vue
andrepestana-aot: ok. Does it create a "NR L..." too, right? Same process?

src/enums/request-code.ts
cameron-freshworks: This make sense for this UI. 
It's on Namex side where it gets confusing since they use request type and request action.
These do align with request action but i don't think we need to be explicit here. 
Either way, its allll good :)

src/components/dialogs/resubmit.vue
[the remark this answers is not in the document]
andrepestana-aot: ok.

src/components/dialogs/resubmit.vue
severinbeauvais: Yes, it creates a NRL (which doesn't live for very long as payment is forthcoming). Same process.

andrepestana-aot: ok 👍

andrepestana-aot (approved):

PR #432, opened 2021-07-30
8429 Expired text

severinbeauvais:
*Issue #:* /bcgov/entity#8429

*Description of changes:*
- display status text for legacy state EXPIRED
- updated app version


src/components/existing-request/existing-request-display.vue
```diff
@@ -500,10 +503,7 @@ export default class ExistingRequestDisplay extends Mixins(
 
   /** True if the current state should display an alert icon. */
   private get isAlertState (): boolean {
-    return ['Cancelled', 'Cancelled, Refund Requested', 'Expired'].includes(this.requestStatusText) ||
-    this.isNotPaid
-    // FUTURE: use enums when EXPIRED state is implemented (ticket #5669)
```
severinbeauvais, line 505: There will be no future, as EXPIRED state is deprecated.

I had hoped that this check could be against state instead of request status text, but "expired" can be a calculated state, so that won't work. The existing code is the lesser of two evils IMHO.

cameron-freshworks (approved):
Looks good!

andrepestana-aot (approved):
LGTM!

PR #433, opened 2021-08-04
8221 - Adjustments for RESUBMIT

andrepestana-aot:
*Issue #:* /bcgov/entity#8221

*Description of changes:*

- Cleared partyIds to make the BE create new applicant records; 
- Changed logic for addComment that was failing for resubmit.

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -485,6 +485,11 @@ export const resubmit: any = async ({ commit, getters }): Promise<boolean> => {
   // override request action code
   nrTypeData['request_action_cd'] = RequestCode.RESUBMIT
 
+  // override applicants partyIds
+  for (let [key, value] of Object.entries(nrTypeData['applicants'])) {
+    value['partyId'] = ''
+  }
+
   // post new NR (without adding comment)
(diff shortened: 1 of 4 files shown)


src/store/actions.ts
```diff
@@ -485,6 +485,11 @@ export const resubmit: any = async ({ commit, getters }): Promise<boolean> => {
   // override request action code
   nrTypeData['request_action_cd'] = RequestCode.RESUBMIT
 
+  // override applicants partyIds
+  for (let [key, value] of Object.entries(nrTypeData['applicants'])) {
+    value['partyId'] = ''
```
cameron-freshworks, line 490: What are the id's used for?

cameron-freshworks (approved):
Looks good to me

severinbeauvais (approved):
Changes are as we discussed ... looks good.

src/store/actions.ts
andrepestana-aot: They come as part of the request with the ids for the old NR. In a submission we are creating a new NR and it needs a new applicant record in the DB. With the ids populated it doesn't create new applicants.

PR #436, opened 2021-08-11
 Quick fix to prevent Applicant Last Name only contains spaces (8214)

eve-git:
*Issue #:* /bcgov/entity#8214

*Description of changes:*
1. First Name, Middle Name and Last Name: trim spaces and combine multiple continuous spaces to one
2. First Name is optional
3. Last Name keeps required


src/components/common/applicant-info-1.vue
```diff
@@ -42,16 +43,17 @@
                 :name="Math.random()"
                 autocomplete="chrome-off"
                 label="Middle Name (Optional)"
+                v-model="middleName"
```
severinbeauvais, line 46: Updating the value was previously done by the `@input` event. Can you remove that event now (line 37) ?

severinbeauvais: Actually, wait...

You have separate events for input changes and for blur. You also have, I think, 3 mechanisms for updating the properties (handleNameBlur, updateApplicant, and the v-model). This needs to be simplified.

src/components/common/applicant-info-1.vue
```diff
@@ -550,6 +558,20 @@ export default class ApplicantInfo1 extends Mixins(ActionMixin) {
     return this.getNrData?.xproJurisdiction
   }
 
+  handleNameBlur (key, event) {
```
severinbeauvais, line 561: This updates the local (class) properties but not the store properties.

The store properties are updated by `updateApplicant`.

And, you also bind the properties to the v-model prop (which is equivalent to a `value` prop and an `input` event) - see https://vuejs.org/v2/guide/forms.html#Basic-Usage

This is too complicated.

severinbeauvais (changes requested):
Code style and quality is good, but the mechanism implemented is too complicated. Please simplify it.

src/components/common/applicant-info-1.vue
eve-git: @input also calls clearValidation. Should I keep clearValidation? also, the @input be removed from firstName and lastName as well?

severinbeauvais: At the moment, `@input` is what actually updates the values in the store, but you could do that differently.

Yes, it's the same problem with all 3 inputs.

src/components/common/applicant-info-1.vue
eve-git: updateApplicant does not update the trimmed name in UI. I think it is better to let user see the formated/trimed input string after the input finished.

src/components/common/applicant-info-1.vue
severinbeauvais: I'll respond in a different conversation. Closing this one.

eve-git: I've updated code by calling setApplicantDetails instead of using v-model. Please review again.

src/components/common/applicant-info-1.vue
eve-git: I've updated code by calling setApplicantDetails instead of using v-model. Please review again.

severinbeauvais: Your changes look much better!

I think you still have a conflict or duplication with `@blur` and `@input`, which both call setApplicantDetails().

severinbeauvais: I un-resolved this conversation. Since I started this conversation, let me close it. I use it as a reminder to look at the changes surrounding this conversation. (If you close it then I might lose track of outstanding issues.)

severinbeauvais: > I think you still have a conflict or duplication with @blur and @input, which both call setApplicantDetails().

eve-git: removeExcessSpace can only be called when input finished (@blur). @blur has to call setApplicationDetails to update the trimmed input text in UI.
Because the function called by @input is a common function which shared by all the text input in this page, so far it is good. I don't want to change it.
Your suggestions?

severinbeauvais: What do you think of this (for the 3 name fields only)?
```
     ...
     @blur="handleNameBlur('firstName', $event)"
     @input="this.clearValidation()"
     ...
```

argush3: is just using `@onblur` problematic?  i.e. are there scenarios where just using `@onblur` will produce undesirable outcomes?

eve-git: > What do you think of this (for the 3 name fields only)?
> 
> ```
>      ...
>      @blur="handleNameBlur('firstName', $event)"
>      @input="this.clearValidation()"
>      ...
> ```

I tested after changing @input function as your suggested. However, not all the test cases are successful: the input value wasn't updated after trimming. I am not sure why.

severinbeauvais: > Currently, clearValidation() is called @input. It may not be a big problem call it in @blur, but I don't want to change existing functionalities.

I don't know. Try it and test it, I guess.

severinbeauvais: There something going on in the store actions/mutate code that requires both the blur and input events to be handled in order to be reactive. I don't understand it, and this was supposed to be a quick fix... so, let's go with this code after all.

To mitigate another developer wasting time trying to figure this out, please add a comment (eg, to handleNameBlur) to describe what you found here.

Also, I'm not sure why GH isn't flagging this as a conflict, but the app version in the package files has been updated (by a high priority fix). Please rebase and update to the next version number.

After this, I will approve and merge this PR.

severinbeauvais: Good comments - thanks, resolving this conversation.

eve-git: The version number updated and the comments added.

severinbeauvais (approved):
Looks good, thanks.

For some reason GH is still showing some conflicts with the app version. When you fix this, I am OK with this PR being merged.

eve-git:
@severinbeauvais I could not save the fix of the conflict until I click the merge button. Can you do it? There are the version conflict. Please use my change. Thanks

severinbeauvais:
> @severinbeauvais I could not save the fix of the conflict until I click the merge button. Can you do it? There are the version conflict. Please use my change. Thanks

You should be able to fix the conflicts yourself...
1. click on "Resolve conflicts" button
2. edit the files so they look correct (delete the <<< and >>> and keep only your new code)
3. click "Mark as resolved" button
4. I think this will then save a merge commit, and then you should be good to go

eve-git:
"Mark as resolved" does not save the change results until click "Commit Merge" button.

PR #439, opened 2021-08-18
Added tooltip for Refund Requested

andrepestana-aot:
*Issue #:* /bcgov/entity#8563

*Description of changes:*

- Added tooltip for Refund Request


src/components/existing-request/existing-request-display.vue
```diff
@@ -749,4 +901,36 @@ export default class ExistingRequestDisplay extends Mixins(
...
+.v-tooltip__content:after {
+  content: "" !important;
+  position: absolute !important;
+  top: 100% !important;
+  right: 50% !important;
+  border-right: 10px solid transparent !important;
+  border-left: 10px solid transparent !important;
+  border-top: 8px solid RGB(73, 80, 87) !important;
+  transform: translateY(50%);
+}
```
cameron-freshworks, line 920: I believe the tool tips were styled in the overrides already. Does this design call for something different than we've already done?

Check it out though, there is tool tips for top, bottom and right i think,

src/enums/payment-method.ts
```diff
@@ -0,0 +1,13 @@
+export enum PaymentMethod {
+    DIRECT_PAY = 'DIRECT_PAY', // Credit Card
+    INTERNAL = 'INTERNAL', // Staff payment: Routing Slip and No Fee
+    DRAWDOWN = 'DRAWDOWN', // BCOL
```
cameron-freshworks, line 4: Should this be `DRAW_DOWN` maybe? Feels like 2 words but i could be wrong :)

src/components/existing-request/existing-request-display.vue
```diff
@@ -228,6 +260,14 @@ export default class ExistingRequestDisplay extends Mixins(
   /** The pending payment, if any. See mounted(). */
   private pendingPayment = null
 
+  /** Params used to format refund message (tooltip and modal). */
+  private refundParams = {
```
cameron-freshworks, line 264: This object could probably benefit from an interface.

src/components/existing-request/existing-request-display.vue
```diff
@@ -734,6 +775,117 @@ export default class ExistingRequestDisplay extends Mixins(
       this.$root.$emit('showSpinner', false)
     }
   }
+
+  private get isThereMoreThanOnePaymentMethod () {
```
cameron-freshworks, line 779: This method and the ones below could probably benefit from return types and methods comments/doc block comments.

src/components/existing-request/existing-request-display.vue
```diff
@@ -734,6 +775,117 @@ export default class ExistingRequestDisplay extends Mixins(
...
+              'Your Name Request has been cancelled and a refund request is being processed.<br/><br/>' +
+              'A credit will be applied to your BC Registries account.<br/>There may be a one day delay before the ' +
+              'credit will show on your transactions / statetements.'
+            this.refundParams.showStaffContact = false
+          } else {
+            // May happen when a PAD is not processed yet.
+            // It usually takes a day to be processed.
+            this.refundParams.refundLabel = 'Refund Not Processed'
+            this.refundParams.refundMessage =
+              'Your Name Request has been cancelled.<br/><br/>' +
```
cameron-freshworks, line 830: Why the double <br>?

src/components/existing-request/existing-request-display.vue
```diff
@@ -44,8 +44,29 @@
                 <v-col cols="12" class="request-status">
                   <span>Request Status:</span>
                   &nbsp;
-                  <span :class="isNotPaid ? 'app-red' : isPaymentProcessing ? 'app-green' : ''">
-                    {{ requestStatusText }}</span>
+                  <span
+                    :class="isRefundRequested ? '' :
+                            isNotPaid ? 'app-red' :
+                            isPaymentProcessing ? 'app-green' : ''">
```
cameron-freshworks, line 50: This chained ternary is a little hard to follow. 
Have you looked into class binding the colors conditionally?

src/enums/payment-method.ts
andrepestana-aot: Thanks for checking but it is DRAWDOWN. At least it is the way it comes from pay-api.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: image
 There's an empty line.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: yes, actually I changed isPaymentProcessing logic and revert this part back the original. Wil come in the next commit.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: 👍

src/components/existing-request/existing-request-display.vue
andrepestana-aot: 👍

src/components/existing-request/existing-request-display.vue
andrepestana-aot: They actually asked me to use a tooltip with 270px as it is the default but it doesn't look good:
image

the `:after` is to make the arrow be pointed down.

cameron-freshworks: I understand the context, this style already exists (or close too) in `overrides.scss` 
You would need to add the content class to your tool tip, you might not need to duplicate the styling in the component itself. 

See class `top-tooltip` in overrides. Using this on the tool tip itself is by adding it as a content class.
`<v-tooltip top content-class="top-tooltip">`

If you need to override the width that can be set on the component itself as well.

src/components/existing-request/existing-request-display.vue
cameron-freshworks: Easy!

src/components/existing-request/existing-request-display.vue
cameron-freshworks: Sounds good!

src/components/existing-request/existing-request-display.vue
andrepestana-aot: ok. Got it. Thanks!

src/components/existing-request/existing-request-display.vue
```diff
@@ -44,8 +44,31 @@
                 <v-col cols="12" class="request-status">
                   <span>Request Status:</span>
                   &nbsp;
-                  <span :class="isNotPaid ? 'app-red' : isPaymentProcessing ? 'app-green' : ''">
-                    {{ requestStatusText }}</span>
+                  <span
+                    :class="isNotPaid ? 'app-red' : isPaymentProcessing ? 'app-green' : ''">
```
severinbeauvais, line 48: If the alterate class binding syntax doesn't make this any neater, consider using a getter in the script instead.

src/components/existing-request/existing-request-display.vue
```diff
@@ -228,6 +262,14 @@ export default class ExistingRequestDisplay extends Mixins(
   /** The pending payment, if any. See mounted(). */
   private pendingPayment = null
 
+  /** Params used to format refund message (tooltip and modal). */
+  private refundParams: RefundParamsIF = {
+    'refundMessage': '',
+    'refundLabel': '',
+    'showStaffContact': false,
+    'showAlertIcon': false
```
severinbeauvais, line 270: You don't need the quotes around the property names.

src/components/existing-request/existing-request-display.vue
```diff
@@ -734,6 +778,136 @@ export default class ExistingRequestDisplay extends Mixins(
       this.$root.$emit('showSpinner', false)
     }
   }
+
+  /** Check if there is more than one payment method used in the payments. */
+  private get isThereMoreThanOnePaymentMethod (): boolean {
```
severinbeauvais, line 783: This is enough new code (and specific enough) that I wonder if it should be its own (sub) component? I see this component is almost 1000 lines already!

severinbeauvais (approved):

cameron-freshworks (approved):

src/components/existing-request/existing-request-display.vue
cameron-freshworks: Awesome!

src/components/existing-request/existing-request-display.vue
andrepestana-aot: I created a getter to define what class to use here. Thanks.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: 👍 done.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: do you think I should move it to payment mixin?

severinbeauvais: Without looking it deeply, this sounds like a good idea.

Does the payment mixin have access to `this.payments` (which is used in the methods)?

I say do it if it's easy and cleaner than this :)

andrepestana-aot: ok. Moved some getters to payment-mixin.

PR #440, opened 2021-08-18
8592

eve-git:
*Issue #:* /bcgov/entity#8592

*Description of changes:*
Remove trailing spaces before & after requested names

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "2.3.10",
+  "version": "2.3.12",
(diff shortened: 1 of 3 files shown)


severinbeauvais (approved):
Looks good. Same solution as name sanitization, right? Could you also add a comment here (why you need both blur and input event handlers)?

eve-git:
1. Added comments as advised
2. Updating searchValue in handleBlur() instead of updating value because 'v-model="searchValue"' has been defined for the field. Just for consistence.

PR #441, opened 2021-08-18
Allow letters and dashes ( - ) in xpro home juri corp num field

eve-git:
*Issue #:* /bcgov/entity#5994

*Description of changes:*
Change the text input rule to allow letters and dashes (-) in xpro home juri corp num field

### DIFF
--- a/src/components/new-request/name-input.vue
+++ b/src/components/new-request/name-input.vue
@@ -64,7 +64,7 @@ export default class NameInput extends Vue {
   /** The array of validation rules for the MRAS corp num. */
   private get mrasRules (): Function[] {
     return [
-      v => (/^\d+$/.test(v) || 'A corporate number is required'),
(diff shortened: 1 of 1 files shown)


andrepestana-aot (approved):
LGTM

severinbeauvais (approved):
Approved.

We try to increase the app version number so it's easy to see what is deployed (ie, up to which point in the commit history).

Here, this change is minor and easy to test, so if you're OK without updating the version number, I am too.

andrepestana-aot:
> Approved.
> 
> We try to increase the app version number so it's easy to see what is deployed (ie, up to which point in the commit history).
> 
> Here, this change is minor and easy to test, so if you're OK without updating the version number, I am too.

Yes, I forgot about it too. For this one is ok, I think...

PR #443, opened 2021-08-20
7018 - Modal for the refund request response

andrepestana-aot:
*Issue #:* /bcgov/entity#7018

*Description of changes:*

- Added modal to show a response for the refund request; 
- moved some methods from `existing-request-display.vue` to `payment-mixin.ts`.

### DIFF
--- a/src/components/dialogs/refund.vue
+++ b/src/components/dialogs/refund.vue
@@ -1,44 +1,83 @@
 <template>
-  <v-dialog width="40rem" :value="isVisible" persistent v-if="payments">
-    <v-card>
+  <span>
+    <v-dialog width="40rem" :value="isConfirmModalVisible" persistent v-if="payments">
+      <v-card>
 
-      <v-card-title class="d-flex justify-space-between">
-        <div>Cancel and Refund</div>
-        <v-btn icon large class="dialog-close" @click="hideModal()">
-          <v-icon>mdi-close</v-icon>
-        </v-btn>
-      </v-card-title>
+        <v-card-title class="d-flex justify-space-between">
+          <div>Cancel and Refund</div>
+          <v-btn icon large class="dialog-close" @click="hideConfirmModal()">
+            <v-icon>mdi-close</v-icon>
+          </v-btn>
+        </v-card-title>
 
-      <v-card-text class="text-body-1">
-        This will cancel your Name Request and refund the Name Request fee, Priority Request
-        fee (if any), and Service fees.
-      </v-card-text>
+        <v-card-text class="text-body-1">
+          This will cancel your Name Request and refund the Name Request fee, Priority Request
+          fee (if any), and Service fees.
+        </v-card-text>
 
-      <v-card-text>
-        <v-alert v-if="fetchError" color="error" icon="mdi-alert" outlined class="my-0" v-html="fetchError" />
-        <refund-summary v-else :payments="payments" />
-      </v-card-text>
+        <v-card-text>
+          <v-alert v-if="fetchError" color="error" icon="mdi-alert" outlined class="my-0" v-html="fetchError" />
+          <refund-summary v-else :payments="payments" />
+        </v-card-text>
 
-      <v-card-text class="text-body-2">
-        The refund will be applied to your original payment method and the requested name will not be
-        examined for use. <template v-if="emailAddress">An email confirming the cancellation and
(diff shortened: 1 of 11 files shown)


src/components/dialogs/refund.vue
```diff
@@ -1,51 +1,91 @@
...
+          <contact-info
+            id="tooltip-contact-info"
+            direction="col" />
+        </v-card-text>
+
+        <v-card-actions class="justify-center">
+          <v-btn
+            class="px-6 button-normal"
+            id="keep-nr-btn"
+            @click="closeResponseModal()">OK</v-btn>
```
severinbeauvais, line 75: Why is this one "close" but the one at the top is "hide"?

src/components/existing-request/existing-request-display.vue
```diff
@@ -58,7 +58,10 @@
                             </v-icon>
                           </span>
                         </template>
-                        <div v-html="refundParams.refundMessage"></div>
+                        <div v-html="refundParams.refundMessageText1" />
+                        <br/>
```
severinbeauvais, line 62: Should this line break depend on whether there is a text2?

src/mixins/payment-mixin.ts
```diff
@@ -28,6 +28,16 @@ export class PaymentMixin extends Mixins(ActionMixin) {
   @Getter getCurrentJsDate!: Date
   @Getter getStaffPayment!: StaffPaymentIF
   @Getter getFolioNumber!: string
+  @Getter getNr!: Partial<NameRequestI>
+
+  /** Params used to format refund message (tooltip and modal). */
+  refundParams: RefundParamsIF = {
```
severinbeauvais, line 34: This data was OK when it was in the component, but now I think it would be better in the store. This is the first time we're storing data in a mixin and it feels wrong. Thoughts?

andrepestana-aot: I'm not 'storing' anything. I don't see any need to store it in vuex... 🤷‍♂️ Do you?

src/components/dialogs/refund.vue
andrepestana-aot: I can use the verb 'hide' here.

src/mixins/payment-mixin.ts
severinbeauvais: You are storing an object with properties that may contain values, and the Existing Request component template binds to this object's properties to conditionally display text, right?

src/components/dialogs/refund.vue
severinbeauvais: Can you use close for the other one, just to be consistent?

andrepestana-aot: 👍

severinbeauvais: I misunderstood your previous comment.

Use either verb you want (as long as they're consistent) :)

src/mixins/payment-mixin.ts
andrepestana-aot: I can store it in vuex if you prefer. I only see a real need to store in vuex when we need to share the state with other components that is not the case but ok, I will add it to vuex.

severinbeauvais: It is shared: this mixin sets it, and the Existing Request Display component gets it.

Thanks.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: No, there should always be an empty line before contact-info.

severinbeauvais: I don't mean the BR on line 66. I mean the one on line 62.

src/mixins/payment-mixin.ts
andrepestana-aot: Yeah, ok. There's a case it can be shared instead of recalculating it. It makes sense. Thanks. Done.

src/components/existing-request/existing-request-display.vue
andrepestana-aot: ok. It didn't look bad with both br elements but you're right.

andrepestana-aot: Thanks! Done.

severinbeauvais (approved):

PR #445, opened 2021-08-24
8653 -  Bug for payments not processed by pay-api (PAD)

andrepestana-aot:
*Issue #:* /bcgov/entity#8563

*Description of changes:*

- Ignored payments if empty

### DIFF
--- a/src/mixins/payment-mixin.ts
+++ b/src/mixins/payment-mixin.ts
@@ -98,13 +98,16 @@ export class PaymentMixin extends Mixins(ActionMixin) {
 
   /** Check if there is more than one payment method used in the payments. */
   get isThereMoreThanOnePaymentMethod (): boolean {
-    const paymentMethods = this.payments.map(payment => payment.sbcPayment.paymentMethod)
-    return paymentMethods.some(method => method !== paymentMethods[0])
+    if (Array.isArray(this.payments) && this.payments.length > 1) {
+      const paymentMethods = this.payments.map(payment => payment.sbcPayment.paymentMethod)
+      return paymentMethods.some(method => method !== paymentMethods[0])
+    }
+    return false
   }
 
   /** Check if the user has been waived of all fees for the NR. */
   get isNoFeePayment (): boolean {
-    if (this.payments.length) { // Payments might not have been processed yet, e.g., PAD
+    if (Array.isArray(this.payments) && this.payments.length) { // Payments might not have been processed yet, e.g., PAD
       if (this.payments.length > 1) {
         return this.payments.reduce((paymentA, paymentB) => paymentA.sbcPayment.paid + paymentB.sbcPayment.paid) === 0
       } else {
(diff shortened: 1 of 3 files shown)


cameron-freshworks (approved):
Looks good! 
Is this related to the 'Retry Payments' bug?
Noticed Retry Payments was having issues in Dev

src/mixins/payment-mixin.ts
```diff
@@ -104,10 +104,12 @@ export class PaymentMixin extends Mixins(ActionMixin) {
 
   /** Check if the user has been waived of all fees for the NR. */
   get isNoFeePayment (): boolean {
-    if (this.payments.length > 1) {
-      return this.payments.reduce((paymentA, paymentB) => paymentA.sbcPayment.paid + paymentB.sbcPayment.paid) === 0
-    } else if (this.payments.length === 1) {
-      return this.payments[0].sbcPayment.paid === 0
+    if (this.payments.length) { // Payments might not have been processed yet, e.g., PAD
```
severinbeauvais, line 107: Are you thinking that `this.payments` is null, or an empty array?

Maybe the check should have been `length >= 1` ?

severinbeauvais: What was the bug exactly?

Maybe the old code would have been OK with `this.payments?.length` on old line 107 and old line 109 ?

andrepestana-aot: Yes, it could be ok with `this.payments?.sbcPayment.paid`. 

For PAD payments will be processed after 1 day. So pay-api doesn't even have a payment. It always send an empty array though.

andrepestana-aot:
> Looks good!
> Is this related to the 'Retry Payments' bug?
> Noticed Retry Payments was having issues in Dev

No, I don't think it's related. These methods are new and only used by refund.

src/mixins/payment-mixin.ts
andrepestana-aot: Do you prefer `this.payments?.sbcPayment.paid` ?

severinbeauvais: Well, first I'm skeptical of `if (this.payments.length)` unless you're sure "payments" is always an array.

Second, I'm not sure what the bug is and therefore what this change fixes.

Third, I'd prefer to see checks for specific properties instead of mixing conditional chaining with values... In your example, `this.payments?.sbcPayment.paid` can be true, false, or undefined. I think it would be better if this code didn't even run if `this.payments` is null/undefined.

andrepestana-aot: > Well, first I'm skeptical of `if (this.payments.length)` unless you're sure "payments" is always an array.
This always come from pay-api as an array but I can add a check for it.

> Second, I'm not sure what the bug is and therefore what this change fixes.
The bug happens for PAD because a payment is not processed in sbc-pay at the time the user creates an NR. Premium users can create an NR and have their payments processed later and that is why we don't have a payment in the array.

> Third, I'd prefer to see checks for specific properties instead of mixing conditional chaining with values... In your example, `this.payments?.sbcPayment.paid` can be true, false, or undefined. I think it would be better if this code didn't even run if `this.payments` is null/undefined.
ok.

andrepestana-aot: ok. I add `Array.isArray()` that is a null/undefined-safe to certify payments is an Array.

severinbeauvais (approved):
Thanks. The code looks safe and is understandable 👍

PR #449, opened 2021-09-03
8788 Reset jurisdiction + other fixes

severinbeauvais:
*Issue #:* /bcgov/entity#8788

*Description of changes:*
- removed some list data out of global scope and imported locally where needed
- updated edit error text to match create error text
- added missing clearErrors() when location is changed
- set isValid false if there are any errors
- added call to new action in validateButton()
- reset additional search values when location has changed
- deleted unused types (models)
- fixed invalid commit causing MRAS search to fail after first time
- added new action startEditName() to check for errors
- deleted duplicate getter
- made copy of Locations so original is not mutated and missing values
- explicitly include misc empty properties so PATCH deletes them
- misc cleanup
- app version = 2.3.18
- ~updated Sentry package to 6.12.0~ _breaking change; needs to be done later_


src/store/getters.ts
```diff
@@ -623,7 +623,7 @@ export const getLocationOptions = (state: StateIF): Array<any> => {
   if (getRequestActionCd(state) === RequestCode.MVE) {
     return Locations.filter(location => location.value === Location.BC)
   }
-  return Locations // return all locations
+  return Locations.filter(location => true)
```
severinbeauvais, line 626: This fixes another issue where the jurisdiction menu (see ticket) is not set. I determined that the reason for this is that the old code used the original `Locations` object, and it was being overwritten (possibly by the v-select). The new code creates a copy of the object (using code consistent within this method. I know there are other ways to do this but this seemed the more clear).

src/components/common/names-capture.vue
```diff
@@ -883,9 +891,14 @@ export default class NamesCapture extends Mixins(CommonMixin) {
 
   @Watch('location')
   handleLocation (newVal: Location, oldVal: Location) {
+    // reset search values when location has changed
     if (newVal !== oldVal) {
-      this.setNRData({ key: 'xproJurisdiction', value: '' })
+      this.setNRData({ key: 'xproJurisdiction', value: 'RESET' })
+      this.setNRData({ key: 'name', value: 'RESET' })
+      this.setNRData({ key: 'corpSearch', value: 'RESET' })
+      this.setNRData({ key: 'homeJurisNum', value: 'RESET' })
```
severinbeauvais, line 899: For debugging only - ignore!

src/components/dialogs/mras-search-info.vue
```diff
@@ -52,7 +52,6 @@ export default class MrasSearchInfoDialog extends Vue {
   @Getter getMrasSearchInfoModalVisible!: boolean
 
   // Global actions
-  @Action setCorpSearch!: ActionBindingIF
```
severinbeauvais, line 55: Not used in this component.

src/components/existing-request/existing-request-display.vue
```diff
@@ -209,7 +209,6 @@ import NamexServices from '@/services/namex.services'
 import ContactInfo from '@/components/common/contact-info.vue'
 
 // Interfaces
-import { NameRequestI, RefundParamsIF } from '@/interfaces'
```
severinbeauvais, line 212: Not used in this component.

src/components/new-request/search.vue
```diff
@@ -124,7 +124,7 @@
       <v-col cols="4" v-if="getIsXproMras">
         <v-select :error-messages="getErrors.includes('jurisdiction') ? 'Please select a jurisdiction' : ''"
                   :items="jurisdictionOptions"
-                  label="Select business's home jurisdiction"
+                  label="Select business' home jurisdiction"
```
severinbeauvais, line 127: Fixed typo.

src/components/new-request/submit-request/reserve-submit.vue
```diff
@@ -23,7 +23,7 @@ import NamexServices from '@/services/namex.services'
 import { Location, NrType, RequestCode } from '@/enums'
 
 @Component({})
-export default class ReserveSubmitButton extends Vue {
+export default class ReserveSubmit extends Vue {
```
severinbeauvais, line 26: Fixed typo. (This component was always imported as "ReserveSubmit" so it didn't matter, but now it's correct and you can import the default name correctly.)

src/interfaces/models.ts
```diff
@@ -8,14 +8,11 @@ import {
   RequestCode
 } from '@/enums'
 
-export type NameReqT = DraftReqI | ConditionalReqI | ReservedReqI
-export type NrDataResponseT = NameRequestI | null
-export type NrDataT = string | null
```
severinbeauvais, line 13: These were no used anywhere.

src/interfaces/models.ts
```diff
@@ -8,14 +8,11 @@ import {
...
 export type SubmissionTypeT = 'examination' | 'consent' | 'normal'
 
 export interface AnalysisJSONI {
   header?: string
-  issues: IssueI[] | null
```
severinbeauvais, line 18: Anything can be null -- it doesn't need to be typed.

src/store/actions.ts
```diff
@@ -319,7 +319,7 @@ export const checkMRAS = ({ getters }, corpNum: string) => {
 export const fetchMRASProfile = async ({ commit, getters }): Promise<any> => {
   if (getters.getCorpSearch) {
     try {
-      let url = `mras-profile/${getters.getRequestJurisdictionCd}/${getters.getCorpSearch}`
+      let url = `mras-profile/${getters.getJurisdictionCd}/${getters.getCorpSearch}`
```
severinbeauvais, line 322: `getRequestJurisdictionCd` was a duplicate getter, so I use the other getter here.

src/store/actions.ts
```diff
@@ -333,7 +333,7 @@ export const fetchMRASProfile = async ({ commit, getters }): Promise<any> => {
         const msg = await NamexServices.handleApiError(err, 'Could not fetch mras profile')
         console.error('fetchMRASProfile() =', msg) // eslint-disable-line no-console
       }
-      commit('mutateName')
```
severinbeauvais, line 336: This caused `name` to be undefined, which caused some de-reference error (eg, "cannot call replace on undefined").

src/store/actions.ts
```diff
@@ -1009,6 +1009,10 @@ export const setDoNameCheck: ActionIF = ({ commit }, check: boolean): void => {
   commit('mutateDoNameCheck', check)
 }
 
+export const startEditName: ActionIF = ({ commit, getters }) => {
+  if (!getters.getEntityTypeCd) commit('setErrors', 'entity_type_cd')
```
severinbeauvais, line 1013: We can add other checks here, as needed.

src/store/getters.ts
```diff
@@ -623,7 +619,7 @@ export const getLocationOptions = (state: StateIF): Array<any> => {
   if (getRequestActionCd(state) === RequestCode.MVE) {
     return Locations.filter(location => location.value === Location.BC)
   }
-  return Locations // return all locations
+  return Locations.filter(() => true)
```
severinbeauvais, line 622: This fixes a separate issue where the jurisdiction menu is not set (see ticket). I determined that the reason for this is that the old code used the original Locations object, and it was being overwritten (possibly by the v-select). The new code creates a copy of the object (using code consistent within this method. I know there are other ways to do this but this seemed the more clear).

src/store/getters.ts
```diff
@@ -699,11 +695,16 @@ export const getCorpNumForReservation = (state: StateIF): any => {
...
-    if (getNrData(state)[key]) {
+    // only set truthy keys
+    // except jurisdiction (which needs to be "" so API can reset it)
+    if (getNrData(state)[key] || key === 'xproJurisdiction') {
       nrData[key] = getNrData(state)[key]
     }
   }
-  let data = {
+
+  console.log('*** getEditNameReservation(), nrData =', nrData)
```
severinbeauvais, line 705: For debugging only - ignore!

andrepestana-aot (approved):
Looks good.

cameron-freshworks (approved):
Looks good! Good clean up 👍

src/components/common/names-capture.vue
```diff
@@ -883,9 +887,14 @@ export default class NamesCapture extends Mixins(CommonMixin) {
 
   @Watch('location')
   handleLocation (newVal: Location, oldVal: Location) {
+    // reset search values when location has changed
     if (newVal !== oldVal) {
-      this.setNRData({ key: 'xproJurisdiction', value: '' })
+      this.setName('')
+      this.setJurisdiction('')
+      this.setCorpSearch('')
+      this.setNoCorpNum(false)
```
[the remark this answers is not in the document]
severinbeauvais: Yes, this is the component for editing an existing NR.

As far as I can tell, yes, the fields are correctly updated. This particular method (watcher) is to reset fields when the user selected a new location (jurisdiction).

src/store/getters.ts
severinbeauvais: Deleted!

src/components/common/names-capture.vue
severinbeauvais: Deleted!

severinbeauvais:
@kialj876 Please have a quick look and merge if OK!

kialj876 (approved):

PR #453, opened 2021-09-10
8681 Show spinner when app starts

severinbeauvais:
*Issue #:* /bcgov/entity#8681

*Description of changes:*
- always show spinner when app starts
- hide spinner once everything is rendered/loaded
- app version = 2.3.21

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -158,12 +158,16 @@ export default class App extends Mixins(DateMixin) {
   }
 
   async created (): Promise<void> {
-    // update Current Js Date now and every 1 minute thereafter
-    await this.updateCurrentJsDate()
-    this.updateCurrentJsDateId = window.setInterval(this.updateCurrentJsDate, 60000)
+    // start with "page not ready" - show spinner
+    // (spinner will be hidden in landing.vue)
+    this.showSpinner = true
 
     // listen for spinner show/hide events
     this.$root.$on('showSpinner', (flag = false) => { this.showSpinner = flag })
+
+    // update Current Js Date now and every 1 minute thereafter
+    await this.updateCurrentJsDate()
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -158,13 +158,17 @@ export default class App extends Mixins(DateMixin) {
...
   async created (): Promise<void> {
-    // update Current Js Date now and every 1 minute thereafter
-    await this.updateCurrentJsDate()
-    this.updateCurrentJsDateId = window.setInterval(this.updateCurrentJsDate, 60000)
+    // start with "page not ready" - show spinner
+    // (spinner will be hidden in landing.vue)
+    this.showSpinner = true
 
     // listen for spinner show/hide events
     this.$root.$on('showSpinner', (flag = false) => { this.showSpinner = flag })
```
severinbeauvais, line 166: Note: The event handler has to be in place before the async call on line 169, otherwise landing.vue gets mounted and sends the event to hide it before the handler exists!

src/views/landing.vue
```diff
@@ -75,11 +75,13 @@ export default class Landing extends Vue {
     const { id } = this
     // if an id was specified then get and load the subject NR
     if (id) {
-      this.$root.$emit('showSpinner', true)
```
severinbeauvais, line 78: This is the old code which showed the spinner while loading the NR, but there is a delay before this happens (ie, while Vue loads all the pages) so the user saw the initial page for a while before this code executed.
Now the spinner is shown as soon as the app loads (ie, before Vue loads all the pages).

severinbeauvais:
This is a draft PR until Linda approves this solution.

cameron-freshworks (approved):
Code looks good!

eve-git (commented):
nice!

PR #460, opened 2021-09-30
#8772 NR UI: need to prevent submission before corp num validation

eve-git:
*Issue #:* /bcgov/entity#8772

*Description of changes:*

- removed the icons that attached to corpNum
- validating "Required" and "at least 4 characters" by corpNum input rules
- validating corpNum existence only when click "Review and Confirm" button
- after clicking "Review and Confirm", waiting for the completion of corpNum validation 

### DIFF
--- a/src/components/common/applicant-info-3.vue
+++ b/src/components/common/applicant-info-3.vue
@@ -32,6 +32,9 @@
                         :rules="phoneRules"
                         @blur="messages = {}"
                         @input="updateApplicant('phoneNumber', $event)"
+                        :error-messages="corpNumError"
+                        @focus="corpNumError = ''"
+                        :loading="loading"
                         filled
                         hide-details="auto"
                         label="Phone Number" />
@@ -147,19 +150,11 @@
                 <v-text-field :messages="messages['corpNum']"
                               :rules="corpNumRules"
                               :error-messages="corpNumError"
-                              validate-on-blur
-                              @blur="corpNumError = ''"
+                              @focus="corpNumError = ''"
                               :loading="loading"
                               filled
-                              v-on:update:error="setError"
-                              :hide-details="hideCorpNum"
                               :label="corpNumFieldLabel"
                               v-model="corpNum">
-                  <template v-slot:append>
-                    <v-icon :class="corpNumError ? 'red--text' : 'green--text'"
-                            v-if="hideCorpNum === 'auto'">
-                      {{ error || loading || corpNumDirty ? 'mdi-close' : 'mdi-check' }}</v-icon>
-                  </template>
                 </v-text-field>
               </div>
             </template>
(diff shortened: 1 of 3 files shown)


src/components/common/applicant-info-3.vue
```diff
@@ -357,40 +347,29 @@ export default class ApplicantInfo3 extends Vue {
...
   }
 
   async validateCorpNum (num): Promise<boolean> {
-    if (!num) {
-      return
-    }
-    if (num.length < 4) {
-      this.corpNumError = 'Too short. Please confirm number.'
+    if (!num || num.length < 4) {
       return
```
severinbeauvais, line 361: Should this return False?

src/components/common/applicant-info-3.vue
```diff
@@ -434,6 +410,9 @@ export default class ApplicantInfo3 extends Vue {
   async nextAction () {
     this.setIsLoadingSubmission(true)
     this.validate()
+    if (this.getShowCorpNum === CorpNumRequests.COLIN) {
+      await this.validateCorpNum(this.getCorpNum)
```
severinbeauvais, line 414: This is good but:
1. while waiting for this to validate, I can still edit the other fields on the page -- this needs to be prevented; you can maybe use the full-page spinner for this (search for "showSpinner" events)
2. it would be nice if the error text is cleared when validation begins, and then is set again after it completes (otherwise I can see the old error while it's still validating, which is incorrect)

src/components/common/applicant-info-3.vue
```diff
@@ -147,19 +147,11 @@
...
-                              validate-on-blur
-                              @blur="corpNumError = ''"
+                              @focus="corpNumError = ''"
                               :loading="loading"
                               filled
-                              v-on:update:error="setError"
-                              :hide-details="hideCorpNum"
                               :label="corpNumFieldLabel"
                               v-model="corpNum">
-                  <template v-slot:append>
```
severinbeauvais, line 158: What about the similar code on applicant-info-2.vue ?

eve-git: I wonder if I need to wait for another ticket or do it together? I am not sure about the test case in applicant-info-2.vue.

src/components/common/applicant-info-3.vue
eve-git: it is a good point to add full-page spinner.

src/components/common/applicant-info-3.vue
eve-git: Yes, return false is better.

src/components/common/applicant-info-3.vue
severinbeauvais: Since you ticket has the UX Assurance label, please check with Tracey or Yui on item 2. If they're OK with it then I'm OK with it.

andrepestana-aot:
Looks good. I just agree with Severin comments. Will wait for the changes.

src/components/common/applicant-info-3.vue
eve-git: Yui agreed with removing error on focusing.

severinbeauvais (approved):
Looks good.

By the way, do not resolve comments opened by reviewers. It is up to the reviews to close their comments (once they are satisfied with the resolution). Do not assume this.

severinbeauvais:
Merging per decision at standup on Oct 4 (so Riyaz can test it ASAP).

PR #462, opened 2021-10-11
More mobile forms and Dialogs

cameron-eyds:
*Issue #:* /bcgov/entity#8667

*Description of changes:*
* Improved Dialogs and some more names forms.
* Disabled touch sliding of tab items.
* adjusted btn alignment in some dialogs for mobile and desktop.

### DIFF
--- a/src/components/dialogs/confirm-name-request.vue
+++ b/src/components/dialogs/confirm-name-request.vue
@@ -54,23 +54,33 @@
             </v-card-text>
 
             <v-card-actions class="pt-8">
-              <v-btn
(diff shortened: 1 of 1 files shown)


cameron-freshworks:
It is a getter! This was done in a previous commit, this is just a small piece of the pie. 
To you your second point though, i could update to use the `$vuetify.breakpoint.thresholds.xs` which is the same as saying `< 600` which i've done. Either or for sure!

To the third point, the app will be bumped 100%, 3.0.0 sounds good. will do this before merging back to main :)

PR #479, opened 2021-11-22
8667 names mobile ux

cameron-freshworks:
*Issue #:* /bcgov/entity#8667

*Description of changes:*
* UX updates that could be done without major work, from this document:
https://docs.google.com/document/d/1lRIhQ4K98h8yU7N1S8GkDgxHZM7VHfB6r1ZNcu8LxLA/edit

### DIFF
--- a/src/components/common/app-title-cols.vue
+++ b/src/components/common/app-title-cols.vue
@@ -1,6 +1,6 @@
 <template>
   <v-row no-gutters>
-    <v-col cols="12" md="6" lg="6" :class="[{'pt-10': !isMobile}, 'pl-2']">
+    <v-col cols="12" md="6" lg="6" class="mt-4" :class="[{'pt-10': !isMobile}, 'pl-2']">
       <span class="h2 colour-white">Name Request
       </span>
     </v-col>
--- a/src/components/common/app-title-cols.vue
+++ b/src/components/common/app-title-cols.vue
@@ -1,13 +1,13 @@
 <template>
-  <div>
-    <v-col cols="12" class="mb-4">
-      <Stats v-if="!isMobile" class="pr-10"/>
-    </v-col>
-    <v-col cols="12" class="mb-n2">
+  <v-row no-gutters>
+    <v-col cols="12" md="6" lg="6" :class="[{'pt-10': !isMobile}, 'pl-2']">
       <span class="h2 colour-white">Name Request
       </span>
     </v-col>
-  </div>
+    <v-col cols="12" md="6" lg="6">
+      <Stats />
+    </v-col>
+  </v-row>
 </template>
 
 <script lang="ts">
--- a/src/components/common/app-title-cols.vue
+++ b/src/components/common/app-title-cols.vue
@@ -1,6 +1,6 @@
 <template>
   <v-row no-gutters>
-    <v-col cols="12" md="6" lg="6" class="mt-4" :class="[{'pt-10': !isMobile}, 'pl-2']">
+    <v-col cols="12" md="6" lg="6" class="mt-4 pl-2" :class="{'pt-10': !isMobile}">
       <span class="h2 colour-white">Name Request
       </span>
     </v-col>
--- a/src/components/dialogs/confirm-name-request.vue
+++ b/src/components/dialogs/confirm-name-request.vue
@@ -34,11 +34,17 @@
           </v-tab-item>
 
           <v-tab-item>
-            <v-card-title class="d-flex justify-space-between">
-              <div>Confirm Name Request</div>
-              <v-btn icon large class="dialog-close float-right" @click="hideModal()">
-                <v-icon>mdi-close</v-icon>
-              </v-btn>
+            <v-card-title>
+              <v-row no-gutters>
+                <v-col cols="8">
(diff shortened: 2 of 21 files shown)


src/components/common/app-title-cols.vue
```diff
@@ -1,13 +1,13 @@
 <template>
-  <div>
-    <v-col cols="12" class="mb-4">
-      <Stats v-if="!isMobile" class="pr-10"/>
-    </v-col>
-    <v-col cols="12" class="mb-n2">
+  <v-row no-gutters>
+    <v-col cols="12" md="6" lg="6" class="mt-4" :class="[{'pt-10': !isMobile}, 'pl-2']">
```
severinbeauvais, line 3: Might as well just move "pl-2" to class property.

src/components/dialogs/confirm-name-request.vue
```diff
@@ -34,11 +34,17 @@
...
+            <v-card-title>
+              <v-row no-gutters>
+                <v-col cols="8">
+                  <span>Confirm Name Request</span>
+                </v-col>
+                <v-col>
+                  <v-btn icon large class="dialog-close" @click="hideModal()">
+                    <v-icon>mdi-close</v-icon>
+                  </v-btn>
+                </v-col>
```
severinbeauvais, line 46: I'm not sure that the new code is equivalent to the old code.
The old code floated the button all the way to the right.
The new code puts the button in the last 4 columns... how does this look on a wide screen?

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -34,6 +34,41 @@
         </v-card-text>
       </template>
 
+      <!-- Mobile Tables -->
```
severinbeauvais, line 37: I see almost no difference between mobile and desktop tables... just that the mobile v-simple-table is nested in a row+col <<< won't this work on desktop as well?

src/components/existing-request/existing-request-search.vue
```diff
@@ -122,6 +121,19 @@
...
+      <v-col  v-if="isMobile" cols="12" md="3" lg="3" class="existing-request-info copy-small">
+        <div class="mx-8 my-10">
+          <ul>
+            <li>Check review time and approval status</li>
+            <li>Upgrade Name Request to Priority ($100.00)</li>
+            <li>Renew your Name Request for an additional 56 days before it expires ($30.00)</li>
+            <li>Cancel your Name Request</li>
+            <li>Edit details, download receipts, and more</li>
+          </ul>
+        </div>
```
severinbeauvais, line 134: You could totally remove duplication here by changing it to a sub-component.

src/components/existing-request/existing-request-search.vue
```diff
@@ -274,4 +286,8 @@ export default class ExistingRequestSearch extends Vue {
   padding: 0 30px !important;
   font-size: 1rem !important;
 }
+
+.mobile-btn {
+  width: 17rem !important;
```
severinbeauvais, line 291: OK, I like this page-specific special class for mobile buttons 👍

src/components/lower-info-area/name-build-sub-components/example-name.vue
```diff
@@ -1,14 +1,14 @@
 <template>
   <v-container id="example-name">
-    <v-row>
-      <v-col cols="5" class="mx-3">
+    <v-row >
```
severinbeauvais, line 3: typo

eve-git (approved):
LGTM

src/components/common/app-title-cols.vue
cameron-freshworks: For sure, will do.

src/components/dialogs/pick-entity-or-conversion.vue
cameron-freshworks: Yea, the only difference is the grid system. 
I couldn't get it to preserve it's original layout on Desktop no, it stacks them as it would mobile. 
I can look into it some more but i didn't want to rebuild the table and how/where it loops. I will look again though.

cameron-freshworks: It might fit if we loop the columns instead of the table. Will try

src/components/existing-request/existing-request-search.vue
cameron-freshworks: Good idea

src/components/dialogs/pick-entity-or-conversion.vue
cameron-freshworks: Updated this to the above. It works pretty much the same. When it breaks into two rows on an ipad, the upper and lower columns don't line up perfectly because there is different column widths but it's definitely good enough.

src/components/existing-request/existing-request-search.vue
cameron-freshworks: Updated

severinbeauvais (approved):

src/components/common/app-title-cols.vue
cameron-freshworks: Oops, missed this. Updated 👍

PR #485, opened 2021-12-14
5739 - Remove Federal option from applicant info province dropdown

ozamani9:
*Issue #:* /bcgov/entity#5739

*Description of changes:*

Remove "Federal" option from applicant info province dropdown


eve-git (approved):
Looks good!

PR #499, opened 2022-01-27
#8901 name choice should not have designation added

eve-git:
Name choice should not have designation added

*Issue #:* /bcgov/entity#8901

*Description of changes:*
1. If a designation has NOT been setup before editing the name, the designation is NOT in the name when no designation input available.
2. If the designation has been setup before editing the name, the designation goes with the name when no designation input available.
3. Removed a piece of the code which never reachable.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.0.7",
+  "version": "3.0.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.0.6",
+  "version": "3.0.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.0.7",
+  "version": "3.0.8",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.0.6",
+  "version": "3.0.7",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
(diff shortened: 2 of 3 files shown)


severinbeauvais:
What are your test cases for this change? Please list them in the ticket.

severinbeauvais (commented):
Looks OK but I'd like to verify the test cases pass before approving this change,

eve-git:
I just removed designation that added after the first create. In the test case below, the value of name variable is 'TEST ABC INC.'. Please check if the results are expected.
test 1.
1. Create a NR of BC Limited company with 'INC' designation
2. Edit after payment

severinbeauvais:
This is not what I asked for. QA does not look at PRs for test notes, so they will not know what you have tested and what they should verify. Please add your test cases to the ticket.

I saw your code change, and I think there may be multiple paths that go through there. You need to test as many of those paths as possible to verify that all cases work as expected. Off the top of my head I can think of the following scenarios:
- first create
- other create? or next + back? what about Start Over and Exit?
- edit flow (after paying for it and before it's approved)
- one choice requested
- mutiple choices requested
- BC jurisdiction
- other jurisdictions

Add anything else you can think of to this list.

eve-git:
I did do several test cases for unit test. I had questions for the results, so that I attached the screenshot in this ticket. I wonder if it is what expected?

The code changes are nothing special and only related to the display of the names. The steps of reproduce the error are in the ticket. In the case, should the tester create test cases?

severinbeauvais:
> I did do several test cases for unit test. I had questions for the results, so that I attached the screenshot in this ticket. I wonder if it is what expected?
> 
> The code changes are nothing special and only related to the display of the names. The steps of reproduce the error are in the ticket. In the case, should the tester create test cases?

THIS is not the ticket. THIS is the Pull Request.

Yes, the tester is free to create their own test cases. I am looking for evidence that you have run a variety of test cases to verify the fix and that nothing else is impacted.

eve-git:
OK, I will attach my question to the ticket. Thanks.

severinbeauvais (approved):
I approve. Thanks for the tests!

severinbeauvais (commented):
The rebase apparently cleared out your version update - please update version again.

severinbeauvais (approved):

eve-git:
@cameron-freshworks Can you please review the changes? Thanks!

cameron-freshworks (approved):
You bet, looks good!

PR #520, opened 2022-04-26
Added try catch block prevent type error happen.

pwei1018:
*Issue #:* /bcgov/entity/11984

*Description of changes:*
In some broswer, the get hotjar user function return type error. Added a try catch block to ignore the error throw up.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "3.0.13",
+  "version": "3.0.14",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "3.0.13",
+      "version": "3.0.14",
       "dependencies": {
         "@babel/compat-data": "^7.12.13",
         "@bcrs-shared-components/enums": "1.0.13",
@@ -4706,7 +4706,7 @@
       "version": "3.1.1",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-3.1.1.tgz",
       "integrity": "sha512-mM8522psRCqzV+6LhomX5wgp25YVibjh8Wj23I5RPkPppSVSjyKD2A2mBJmWGa+KN7f2D6LNh9jkBCeyLktzjg==",
-      "dev": true,
+      "devOptional": true,
       "dependencies": {
         "normalize-path": "^3.0.0",
         "picomatch": "^2.0.4"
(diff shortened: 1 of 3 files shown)


severinbeauvais (commented):
1. Isn't Omid working on this already?
2. Definitely update the app version in package.json (and run npm i) so that we can track versions through Dev/Test/Prod environments.

pwei1018:
> 1. Isn't Omid working on this already?
Don't think so.
> 2. Definitely update the app version in package.json (and run npm i) so that we can track versions through Dev/Test/Prod environments.

It is part to 3.0.13 task. Do we still need to update the version?

https://github.com/bcgov/namerequest/commit/c8181241d9eb39cb3b7ba637df134d7f77bd65e0

ozamani9gh (approved):
looks good

severinbeauvais:
> It is part to 3.0.13 task. Do we still need to update the version?

Yes, otherwise how will you know the difference between the broken 3.0.13 and this new code (eg, in Dev)?

pwei1018:
> > It is part to 3.0.13 task. Do we still need to update the version?
> 
> Yes, otherwise how will you know the difference between the broken 3.0.13 and this new code (eg, in Dev)?

Done.

package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "name-request",
-  "version": "3.0.13",
+  "version": "3.0.14",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "3.0.13",
+      "version": "3.0.14",
```
severinbeauvais, line 9: I'm not expecting this change (or the dev/devOptional changes below).

Did you update this manually or did you do `npm i`? What version of node are you using?

severinbeauvais: You should be using node v12 or v14 for now.

pwei1018: I was using npm i.

pwei1018: I updated the package.json version from 3.0.13 to 3.0.14 and `npm i`.

severinbeauvais: Oh, you're right! Looks like you did the right thing.

I'm still puzzled by the dev -> devOptional changes below. Those should have happened when Omid did his `npm i`.

Well, let's go with your code as-is and see what happens :)

severinbeauvais (approved):
Good to go.

PR #522, opened 2022-05-11
#8313 display the exit dialog for incomplete payment

eve-git:
*Issue #:* /bcgov/entity#8313

*Description of changes:*
display the exit dialog for incomplete payment


src/components/dialogs/exit.vue
```diff
@@ -1,6 +1,29 @@
 <template>
   <v-dialog v-model="showModal" max-width="45rem">
-    <v-card>
+    <!-- exit incomplete payment -->
```
severinbeauvais, line 3: This code looks OK but I think it would be better to create a second dialog file here.

Separate dialog files would keep the logic out of here and into the caller. That would simplify this file and would probably simplify the caller as well: it would just be a separate conditional, ie. "if A then show D1 else if B then show D2 else ...".

The more simple the code in this project, the better for all of us :)

severinbeauvais (commented):
The code looks good but I have a small request (see above).

Also, please update the app version so we can track this change through our environments (Dev/Test/Prod). Thanks.

eve-git:
> The code looks good but I have a small request (see above).
> 
> Also, please update the app version so we can track this change through our environments (Dev/Test/Prod). Thanks.

severinbeauvais (approved):
Looks great! Thanks for the updates.

PR #538, opened 2022-08-10
12549 added Registry Your Business functionality

severinbeauvais:
*Issue #:* bcgov/entity#12549

*Description of changes:*
- added Dashboard URL config key
- deleted obsolete configuration keys
- updated Affiliation Error Dialog
- added Show Register Button condition
- added Register Your Business event handler
- updated NR Approved Gray Box
- added firm entity types (and sorted the enum)
- added Business Request optional object
- added isFirm()
- updated createAffiliation()
- added Dashboard URL key fetch
- updated createNrAffiliation()
- updated createBusiness()
- added fetchBusiness()
- added feature flag
- naviagte to BC Registry instead of using Incorporate Login Modal
- fixed App logged in processing
- removed NR_DATA after affiliation
- misc cleanup
- app version = 3.1.0


public/config/configuration.json
```diff
@@ -4,20 +4,15 @@
   "AUTH_WEB_URL": "https://dev.bcregistry.ca/business/auth/",
   "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
   "BUSINESSES_URL": "https://dev.bcregistry.ca/business/auth/",
-  "DEV_EXISTING_NR_TIMEOUT_MS": "0",
-  "DEV_NR_COMPLETION_TIMEOUT_MS": "0",
+  "DASHBOARD_URL": "https://dev.bcregistry.ca/business/",
```
severinbeauvais, line 7: This is in `web-url` vault, `business` application, which is already being loaded (see ~/devops/vaults.json).

public/config/configuration.json
```diff
@@ -4,20 +4,15 @@
   "AUTH_WEB_URL": "https://dev.bcregistry.ca/business/auth/",
   "AUTH_TOKEN_URL": "https://dev.oidc.gov.bc.ca/auth/realms/fcf0kpqr/protocol/openid-connect/token",
   "BUSINESSES_URL": "https://dev.bcregistry.ca/business/auth/",
-  "DEV_EXISTING_NR_TIMEOUT_MS": "0",
```
severinbeauvais, line 7: These are no longer referenced in the code.

src/components/dialogs/affiliation-error.vue
```diff
@@ -1,19 +1,19 @@
...
+        <v-icon large color="error" class="my-4">mdi-alert</v-icon>
+        <span>Error Adding Name Request</span>
       </v-card-title>
-      <v-card-text>
-        <slot name="text">The specified name request has already been affiliated.</slot>
+
+      <v-card-text class="text-center">
+        The specified name request may already be affiliated. Go to
+        <a :href="businessRegistryUrl">your Business Registry</a>
+        to use the name request.
```
severinbeauvais, line 12: This text needs to be approved.

src/components/existing-request/existing-request-display.vue
```diff
@@ -20,7 +20,8 @@
           </v-col>
         </v-row>
       </div>
-      <names-gray-box
+
+      <NamesGrayBox
```
severinbeauvais, line 24: This format (PascalCase instead of kebab-case) allows you to click F12 on this component and it will go to its implementation.

src/components/existing-request/existing-request-display.vue
```diff
@@ -407,46 +401,51 @@ export default class ExistingRequestDisplay extends Mixins(
...
     return (
       this.isBenefitCompany(this.nr) &&
       this.actions.includes(NrAction.INCORPORATE)
     )
   }
 
+  /** True if the Register button should be shown. */
+  get showRegisterButton (): boolean {
+    return this.isFirm(this.nr)
+  }
```
severinbeauvais, line 417: This is trivial atm but it might become more complex later, and this also is a nice abstraction from what the component needs to know (do I show the button?) to what the business logic is (is this a firm?).

src/components/existing-request/existing-request-display.vue
```diff
@@ -641,10 +640,18 @@ export default class ExistingRequestDisplay extends Mixins(
...
     this.setConditionsModalVisible(true)
   }
 
+  /** Called to register the business. */
+  protected async registerYourBusiness (): Promise<void> {
+    // safety check
+    if (!this.isNrApprovedOrConditional) return
+
+    return this.affiliateOrLogin()
+  }
```
severinbeauvais, line 653: Again, trivial atm but it might become more complex later.

Also this lets me add in a safety check (so I don't have to assume that the button can only be clicked when the NR is approved or conditional).

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -1,68 +1,80 @@
...
     } catch (err) {
       console.error('createAffiliation() =', err) // eslint-disable-line no-console
 
       // hide spinner
       this.$root.$emit('showSpinner', false)
 
-      // clear NR Data and show error dialog
+      // clear NR data and show error dialog
       sessionStorage.removeItem('NR_DATA')
       this.setAffiliationErrorModalVisible(true)
```
severinbeauvais, line 79: This modal is displayed even when there is a createBusiness() error, so it displays an invalid message.

I didn't fix this here but I didn't break it more, either :)

Maybe this can be fixed in the future.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -1,68 +1,80 @@
...
+
+      //
+      // try to affiliate NR to current account
+      //
+      const affiliationResponse = await AuthServices.createNrAffiliation(accountId, nr)
+      if (
+        affiliationResponse.status === BAD_REQUEST &&
+        affiliationResponse.data.code === 'NR_CONSUMED'
+      ) {
+        throw new Error('The name request is already affiliated')
```
severinbeauvais, line 37: Possible future improvement: if the NR is already affiliated, try to create the business; if the business already exists, go to its dashboard.

Implementing this would require API changes, since it will need to provide the Business Identifier if already affiliated (otherwise there is no way to get it).

Or, we could navigate to the user's Business Registry page, and they could open the business from there. (future)

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -1,68 +1,80 @@
...
+      // NB: fall back is user's default account
+      const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT')).id || 0
+
+      //
+      // try to affiliate NR to current account
+      //
+      const affiliationResponse = await AuthServices.createNrAffiliation(accountId, nr)
+      if (
+        affiliationResponse.status === BAD_REQUEST &&
+        affiliationResponse.data.code === 'NR_CONSUMED'
```
severinbeauvais, line 35: Actually, it's NOT consumed. It would be nice to know whether the NR is already affiliated, or consumed, or expired, or whatever. Then we could give the user some options (or instructions) here.

This is future.

src/services/auth.services.ts
```diff
@@ -8,21 +8,30 @@ export default class AuthServices {
...
+  static async createNrAffiliation (accountId: number, nr: any): Promise<any> {
+    const url = `${this.authApiUrl}/orgs/${accountId}/affiliations?newBusiness=true`
+    const requestBody: CreateNRAffiliationRequestBody = {
+      businessIdentifier: nr.nrNum,
+      phone: nr.applicants?.phoneNumber || '',
+      email: nr.applicants?.emailAddress || ''
+    }
+
+    return axios.post(url, requestBody)
+      .catch(error => error?.response)
```
severinbeauvais, line 21: Now this method never returns an exception, and the caller can check the response status.

src/services/business.services.ts
```diff
@@ -9,23 +8,14 @@ export default class BusinessServices {
...
-  static async createBusiness (filingBody: BusinessRequest): Promise<any> {
-    try {
-      const response = await axios.post(
-        `${BusinessServices.legalApiUrl}/businesses?draft=true`, filingBody
-      )
-      if (response?.status >= 200 && response?.status < 300) {
-        return response
-      }
-    } catch {
-      // delete the created affiliation if the update failed for avoiding orphan records
```
severinbeauvais, line 22: It is not this method's job to clean up for something that happened in the parent. Hence, this is moved to the caller.

severinbeauvais:
Hi reviewers! Could you have a look at this and give me your feedback and expertise?

I am reusing some existing affiliation/navigation functionality -- but I am not sure if that functionality is currently used? And if not, why not?

src/components/existing-request/existing-request-display.vue
```diff
@@ -170,21 +171,24 @@
...
-          <nr-approved-gray-box
+          <NrApprovedGrayBox
             class="mt-5"
             v-if="showNrApprovedGrayBox"
-            :nrNum="nr.nrNum"
+            :nrNum="nr && nr.nrNum"
             :approvedName="approvedName && approvedName.name"
             :emailAddress="nr && nr.applicants && nr.applicants.emailAddress"
+            :showRegisterButton="showRegisterButton"
+            :disabled="disableUnfurnished"
```
severinbeauvais, line 187: Disabled same as other buttons on line 164.

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Actually, something needs to be done here now (*): if I am logged in and create a NR then it is automatically affiliated to my account, so this call fails.

What should happen is that the UI moves to the next step (try to create business), and if the business is already created then navigate to open it.

(*) I am looking into what needs to be updated to do this. In the meantime, this does not block this PR -- it still works OK, ie:

image

cameron-freshworks: Sorry this is pulling on recollection from ages ago..
I believe you have it right, when there is an account present during the NR creation, i believe an event is published from Namex which Auth picks up on and subsequently auto-affiliates. (Or it could be a request directly to Auth, i can't recall)

I don't know the full context of your work but if you need to add another step here and if* the request isn't directly from Name Request UI to Auth, it might require an update in Namex/Auth. 

Sorry, not a 'totally' useful comment just trying to add some context, if it helps 😁

cameron-freshworks: 🤔 Just at a quick glance, looks like there is a direct link to the auth services, so maybe not so tricky after all.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -1,68 +1,84 @@
...
+      // NB: fall back is user's default account
+      const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT')).id || 0
+
+      //
+      // try to affiliate NR to current account
+      //
+      const createAffiliationResponse = await AuthServices.createNrAffiliation(accountId, nr)
+      if (
+        createAffiliationResponse.status === BAD_REQUEST &&
+        createAffiliationResponse.data.code === 'NR_CONSUMED'
```
severinbeauvais, line 35: Actually, it's not necessarily consumed. It would be nice to know whether the NR is already affiliated, or consumed, or expired, or whatever. Then we could give the user some options (or instructions) here.

---

Actually, something needs to be done now (*): if I am logged in and create a NR then it is automatically affiliated to my account, so this call fails.

What should happen is that the UI moves to the next step (try to create business), and if the business is already created then navigate to open it.

(*) I am looking into what needs to be updated to do this. In the meantime, this does not block this PR -- it still works OK, ie:

image

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Code is updated. This conversation is obsolete.

src/components/dialogs/affiliation-error.vue
severinbeauvais: Janis is OK with this text and link for now. She will update this text depending on the possible error scenarios.

src/components/dialogs/affiliation-error.vue
```diff
@@ -1,19 +1,19 @@
 <template>
-  <v-dialog v-model="showModal" max-width="60%">
+  <v-dialog v-model="getAffiliationErrorModalVisible" max-width="60%" persistent>
```
severinbeauvais, line 2: Persistent so that user must click OK to close this dialog. Otherwise, the visibility state is not reset and this modal does not re-display on subsequent errors.

src/plugins/featureFlags.ts
```diff
@@ -5,9 +5,10 @@ declare const window: any
 
 /**
  * Default flag values when LD is not available.
- * Uses "business-edit" project (per LD client id in config).
+ * Uses "namerequest" project (per LD client id in config).
  */
 const defaultFlagSet: LDFlagSet = {
+  'enable-sp-gp-dba': false, // by default, don't show firm Register button
```
severinbeauvais, line 11: I added this FF in LaunchDarkly. It is the same flag as in sbc-auth to allow only SP/GP registrations from a NR (else goes to OneStop).

Currently, it is True in Dev and Test, and False in Prod.

When SP/GP is released to the public, this flag will have to be enabled in Prod (for both sbc-auth and namerequest LD projects).

src/views/auth/Signin.vue
```diff
@@ -53,11 +53,9 @@ export default class Signin extends Mixins(NrAffiliationMixin, UpdateUserMixin)
 
     await this.updateUser()
 
-    // If there is stored NR data to process, create the affiliation.
-    // If this succeeds, the user will be redirected to their Manage Businesses page.
-    // If this fails, just fall through...
+    // If there is stored NR data to process then affiliate it.
     const nr = JSON.parse(sessionStorage.getItem('NR_DATA'))
-    if (nr) await this.createAffiliation(nr).catch()
```
severinbeauvais, line 60: `createAffiliation` doesn't throw errors, so catch is not needed.

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Code is updated. This conversation is obsolete.

src/App.vue
```diff
@@ -1,14 +1,6 @@
 <template>
   <v-app id="app">
     <div id="main-column">
-      <SbcAuthenticationOptionsDialog
-        attach="#app"
-        :inAuth="false"
-        :showModal="getIncorporateLoginModalVisible"
-        :redirectUrl="nameRequestUrl"
-        @close="setIncorporateLoginModalVisible(false)"
-      />
```
severinbeauvais, line 10: Per Janis, this is obsolete. We now redirect to BC Registry site to login (or create a new account).

src/App.vue
```diff
@@ -194,15 +180,16 @@ export default class App extends Mixins(DateMixin) {
...
 
-    // in case user is already logged in,
-    // get and store keycloak roles
-    try {
-      const keycloakRoles = getKeycloakRoles()
-      this.setKeycloakRoles(keycloakRoles)
-      console.info('Got roles!') // eslint-disable-line no-console
-    } catch (error) {
-      // just log the error message
-      console.log(`Did not get roles (${error.message})`) // eslint-disable-line no-console
```
severinbeauvais, line 205: Moved to new mixin.

src/App.vue
```diff
@@ -194,15 +180,16 @@ export default class App extends Mixins(DateMixin) {
...
+    // in case user is already logged in, load Keycloak roles and update LaunchDarkly
+    this.loadKeycloakRoles()
+    await this.updateUser()
```
severinbeauvais, line 185: LD update call was previously missing.

src/components/new-request/stats.vue
```diff
@@ -56,7 +56,7 @@
 </template>
 
 <script lang="ts">
-import { Vue, Component, Watch } from 'vue-property-decorator'
```
severinbeauvais, line 59: `Watch` wasn't used in this component so I deleted its import.

(I happened upon this file while verifying other things.)

eve-git (approved):

src/components/new-request/stats.vue
ozamani9gh: awesome , thanks bud

ozamani9gh (approved):
Looks Great Sev.  Great work on this

cameron-freshworks (approved):
Looks great, nice work!

src/mixins/load-keycloak-roles-mixin.ts
```diff
@@ -0,0 +1,21 @@
+import { Component, Vue } from 'vue-property-decorator'
```
pwei1018, line 1: Do you think this function can move to common components in the future?

pwei1018 (approved):

src/mixins/load-keycloak-roles-mixin.ts
severinbeauvais: Yes. I just followed the existing code style (eg, NR Affiliation Mixin). Personally I don't mind this as it doesn't pollute the namespace of every component with a method they won't use.

Can this be done in the future maybe?

cameron-freshworks: `vue-property-decorator` isn't a package required for Vue 3 (or Vue 2 that integrates with composition api),
but to fully move away from it would require a rewrite of A LOT of class based components 😅 

Just some thoughts!

severinbeauvais: I think Patrick was referring to this new mixin that I created that contains a single method.

cameron-freshworks: Ah of course 😆  Sorry!

PR #546, opened 2022-08-30
13323 Updated affiliation error dialog

severinbeauvais:
*Issue #:* bcgov/entity#13323

*Description of changes:*
- refactored Affiliation Error dialog to support different titles and text, and to add contact info
- changed Affiliation Modal state from boolean to number
- updated createAffiliation() to support different error dialogs (WIP)
- misc cleanup
- app version = 3.1.8


src/components/dialogs/affiliation-error.vue
```diff
@@ -1,19 +1,36 @@
 <template>
-  <v-dialog v-model="getAffiliationErrorModalVisible" max-width="60%" persistent>
-    <v-card class="notify-dialog">
-      <v-card-title class="flex-column">
-        <v-icon large color="error" class="my-4">mdi-alert</v-icon>
-        <span>Error Affiliating Name Request</span>
+  <v-dialog :value="getAffiliationErrorModalValue" max-width="45rem" persistent>
```
severinbeauvais, line 2: I updated the max width so the dialog looks like this on a wide screen:

image

And like this on a narrow screen:

image

src/components/dialogs/affiliation-error.vue
```diff
@@ -1,19 +1,36 @@
 <template>
-  <v-dialog v-model="getAffiliationErrorModalVisible" max-width="60%" persistent>
-    <v-card class="notify-dialog">
```
severinbeauvais, line 3: This class was never used. This dialog also now has an id if needed.

src/components/dialogs/affiliation-error.vue
```diff
@@ -23,17 +40,32 @@
...
+  components: { ContactInfo }
+})
 export default class AffiliationErrorDialog extends Vue {
-  @Getter getAffiliationErrorModalVisible!: boolean
-  @Action setAffiliationErrorModalVisible!: ActionBindingIF
+  @Getter getAffiliationErrorModalValue!: number
+  @Action setAffiliationErrorModalValue!: ActionBindingIF
+
+  get isUnableToStart (): boolean {
+    return (this.getAffiliationErrorModalValue === 1)
```
severinbeauvais, line 53: I didn't think it was worth an enum for this. Thoughts?

0 = hide modal
1 = unable to start registration
2 = associated with another account

Maybe create an enum if we add another value in the future?

src/components/dialogs/error.vue
```diff
@@ -15,7 +15,7 @@
       </v-card-text>
 
       <v-card-text class="copy-normal pt-6">
-        <contact-info class="mt-2" direction="col" />
+        <ContactInfo direction="col" />
```
severinbeauvais, line 18: VS Code colours this component specially and allows clicking F12 to go to its declaration. Example:

image

src/components/dialogs/exit-incomplete-payment.vue
```diff
@@ -1,21 +1,26 @@
 <template>
   <v-dialog v-model="showModal" max-width="45rem">
     <v-card>
-      <v-card-title class="d-flex justify-space-between mt-n3">
```
severinbeauvais, line 4: Since I used this component (and others) as a template for the Affiliation Error dialog, I fixed up the layout in this component for consistency.

src/interfaces/new-request-interface.ts
```diff
@@ -34,7 +34,7 @@ export interface NewRequestIF {
   extendedRequestType: SelectOptionsI
   getNameReservationFailed: boolean
   helpMeChooseModalVisible: boolean
-  affiliationErrorModalVisible: boolean
+  affiliationErrorModalValue: number // 0, 1 or 2
```
severinbeauvais, line 37: I didn't think it was worth an enum for this. Thoughts?

0 = hide modal
1 = unable to start registration
2 = associated with another account

Maybe create an enum if we add another value in the future?

src/services/auth.services.ts
```diff
@@ -33,7 +33,10 @@ export default class AuthServices {
     const url = `${this.authApiUrl}/orgs/${accountId}/affiliations`
 
     return axios.get(url)
-      .catch(error => error?.response)
+      .then(response => {
+        if (response?.data) return response.data
+        throw new Error('Invalid affiliations')
+      })
```
severinbeauvais, line 39: This is now consistent with the other GETs in this class.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -33,6 +34,10 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
       if (createAffiliationResponse.status === CREATED) {
         // create the business
         const businessId = await this.createBusiness(accountId, nr, true)
+          .catch(error => {
+            this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
+            throw error
+          })
```
severinbeauvais, line 40: Needed because `createBusiness()` can throw an exception.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -42,21 +47,26 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
...
         createAffiliationResponse.status === BAD_REQUEST &&
-        createAffiliationResponse.data.code === 'NR_CONSUMED'
+        createAffiliationResponse.data?.code === 'NR_CONSUMED'
       ) {
         // fetch existing affiliations
         const fetchAffiliationsResponse = await AuthServices.fetchAffiliations(accountId)
+          .catch(error => {
+            this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
+            throw error
+          })
```
severinbeauvais, line 57: Needed because `fetchAffiliations()` can throw an exception.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -42,21 +47,26 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
...
         // is this a name request affiliation?
-        const entities = fetchAffiliationsResponse.data?.entities || []
         const nameRequestEntity = entities.find(entity => entity.businessIdentifier === nr.nrNum)
         if (nameRequestEntity) {
           // create the business
           const businessId = await this.createBusiness(accountId, nr, false)
+            .catch(error => {
+              this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
+              throw error
+            })
```
severinbeauvais, line 69: Needed because `createBusiness()` can throw an exception.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -69,28 +79,35 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
           // use existing business id
           const businessId = temporaryBusinessEntity.businessIdentifier
 
+          if (!businessId) {
+            this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
+            throw Error('Invalid existing affiliation business identifier')
+          }
```
severinbeauvais, line 85: Safety check.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -103,7 +120,7 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
         await AuthServices.removeNrAffiliation(accountId, nr.nrNum).catch(() => null)
       }
 
-      throw new Error('Unable to create new business')
```
severinbeauvais, line 106: `Error()` can be called with or without `new`, so I went with the cleaner approach.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

src/components/dialogs/affiliation-error.vue
```diff
@@ -1,19 +1,36 @@
...
       </v-card-title>
 
-      <v-card-text class="text-center">
-        The specified name request may already be affiliated. Go to
-        <a :href="businessRegistryUrl">your Business Registry</a>
-        to use the name request.
+      <v-card-text class="copy-normal pt-8">
+        <p v-if="isUnableToStartRegistration" class="mb-0">
+          We encountered an error starting your registration. Try again later, or go to
+          <a :href="businessRegistryUrl">My Business Registry</a> and manually add this Name
```
cameron-freshworks, line 17: Normally i agree the same window makes sense but in Name Request, would a new tab make sense? 
Does the user need to reference any content here that might help after navigating to the registries home?

cameron-freshworks (approved):
Looks good!

src/components/dialogs/affiliation-error.vue
severinbeauvais: That's a really good question.

In this "hand-holding" flow (to create a business from a NR), it makes sense to stay in the same window, but what you've pointed out an "error handling" scenario... In this case I think it makes sense to open My Business Registry in a new tab.

Fixed!

ozamani9gh (approved):

PR #547, opened 2022-08-30
13323 Fixed error cases

severinbeauvais:
*Issue #:* bcgov/entity#13323

*Description of changes:*

While testing (in Dev) and re-reading the ticket another time, I caught this mixup in displaying the correct error.

- fixed error cases
- app version = 3.1.9

### DIFF
--- a/src/mixins/nr-affiliation-mixin.ts
+++ b/src/mixins/nr-affiliation-mixin.ts
@@ -89,11 +89,11 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
           return
         }
 
-        this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
+        this.setAffiliationErrorModalValue(NrAffiliationErrors.ASSOCIATED_OTHER_ACCOUNT)
         throw Error('Unable to find existing affiliation')
       }
 
-      this.setAffiliationErrorModalValue(NrAffiliationErrors.ASSOCIATED_OTHER_ACCOUNT)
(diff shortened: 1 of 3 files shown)


src/mixins/nr-affiliation-mixin.ts
```diff
@@ -89,11 +89,11 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
           return
         }
 
-        this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
+        this.setAffiliationErrorModalValue(NrAffiliationErrors.ASSOCIATED_OTHER_ACCOUNT)
```
severinbeauvais, line 92: If we get to this point then we know the NR is already affiliated, but when we fetched the affiliations for the current account, we didn't find a matching name request affiliation or temporary business affiliation -- so the most likely scenario is that the NR is affiliated to a different account.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -89,11 +89,11 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
...
         throw Error('Unable to find existing affiliation')
       }
 
-      this.setAffiliationErrorModalValue(NrAffiliationErrors.ASSOCIATED_OTHER_ACCOUNT)
+      this.setAffiliationErrorModalValue(NrAffiliationErrors.UNABLE_TO_START_REGISTRATION)
```
severinbeauvais, line 96: If we get to this point then we were unable to create a new affiliation and the NR is not already affiliated.

This is a "technical problem" so use this version of the affiliation error dialog.

cameron-freshworks (approved):
Looks good!

PR #560, opened 2022-11-08
13643 use shared WebChat component

eve-git:
*Issue #:* /bcgov/entity#13643

*Description of changes:*
1. replaced chat-popup with bcrs-shared-components/web-chat 2.0.0, which is the nearest version before Vue upgrade in bcrs-shared-components
2. replaced the style folder name /assets/scss with /assets/styles, to make it consistent with what in bcrs-shared-components


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.2.4",
+  "version": "3.2.5",
```
severinbeauvais, line 3: I recommend you go to 3.3.0 to denote that something significant happened since 3.2.x.

package.json
```diff
@@ -17,6 +17,7 @@
     "@bcrs-shared-components/enums": "1.0.13",
     "@bcrs-shared-components/interfaces": "1.0.22",
     "@bcrs-shared-components/staff-payment": "1.0.29",
+    "@bcrs-shared-components/web-chat": "2.0.0",
```
severinbeauvais, line 20: I saw your comment about "last version before Vue upgrade" : 👍

src/App.vue
```diff
@@ -155,6 +164,31 @@ export default class App extends Mixins(
...
+  }
+
+  get openTooltipMessage (): String {
+    return 'Click here to chat live with Helpdesk staff about Name Requests.'
+  }
+
+  get closedTooltipMessage (): String {
+    return 'We are closed. The Service BC Contact Centre is open Monday through Friday' +
+           '7:30am - 5:00pm Pacific Time excluding BC statutory holidays.'
+  }
```
severinbeauvais, line 190: Nice use of getters to provide messages 👍 

You don't need the other getters. You can simply do like this: https://github.com/bcgov/business-create-ui/blob/25827cb57a6bae5089ff60dd79fb553cbf6e6a2f/src/App.vue#L76

And this: https://github.com/bcgov/business-create-ui/blob/25827cb57a6bae5089ff60dd79fb553cbf6e6a2f/src/App.vue#L337

severinbeauvais: I think you don't need the closed tooltip message -- the default message in the shared component looks the same to me.

If you added the message here to make sure it's separate from the default message (in case someone changes the default) then that's OK with me.

src/components/dialogs/staff-payment-error.vue
```diff
@@ -71,7 +71,7 @@ export default class StaffPaymentErrorDialog extends Vue {
 </script>
 
 <style lang="scss" scoped>
-@import "../../assets/scss/theme.scss";
+@import "../../assets/styles/theme.scss";
```
severinbeauvais, line 74: This should be changed to the "@" format like the other files.

severinbeauvais: By the way (for your information only), here is where the "@" shortcut is declared:
https://github.com/bcgov/business-create-ui/blob/25827cb57a6bae5089ff60dd79fb553cbf6e6a2f/tsconfig.json#L23

severinbeauvais (commented):
~You can also delete the tests/unit/chat-popup.spec.ts file.~ _done, thanks_

severinbeauvais:
I wonder why the CI checks didn't run?

src/App.vue
eve-git: I removed the closed tooltip message. You are right, it is not necessary.

src/components/dialogs/resubmit.vue
```diff
@@ -311,7 +311,7 @@ export default class ResubmitDialog extends Mixins(
 </script>
 
 <style lang="scss" scoped>
-@import "@/assets/scss/theme";
+@import "@/assets/styles/theme";
```
severinbeauvais, line 314: I just noticed this now!

This is an invalid import. It didn't show as an error because none of the SCSS statements here use any variabled from theme.scss.

Please fix this: either fix or delete this import (since it's not doing anything). Check the other files, too.

src/components/existing-request/existing-request-display.vue
```diff
@@ -792,7 +792,7 @@ export default class ExistingRequestDisplay extends Mixins(
 </script>
 
 <style lang="scss" scoped>
-@import "@/assets/scss/theme";
+@import "@/assets/styles/theme";
 
 .nr-data .col {
   color: $text;
```
severinbeauvais, line 798: I don't know why this isn't showing as an error (`$text` is not declared). Does this work even with the weird import?

src/components/dialogs/resubmit.vue
severinbeauvais: Even tested this (in request-details.vue) and it seems this import works after all. (I just learned something.)

Therefore this is not an error. No need to change the imports.

src/components/existing-request/existing-request-display.vue
severinbeauvais: Per comment above, this works, so this conversation is invalid and I'm going to resolve it.

severinbeauvais (approved):
LGTM.

I don't know who the other reviewer is...? Did you mean rarmitag?

eve-git:
> LGTM.
> 
> I don't know who the other reviewer is...? Did you mean rarmitag?

Yes, I asked Richard to review. I can see his name in the page.

PR #569, opened 2022-12-06
Add the conditions to show Business Register button

eve-git:
*Issue #:* /bcgov/entity#14366 and /bcgov/entity#14465

*Description of changes:*
Add the conditions to show Business Register button in existing NR page:
1. the NR should not be used for name change
2. the NR request status 
    a.  is Approved OR
    b.  is Conditional and consent status is
             - null (not required)
             - R (received)
             - N (waived)

### DIFF
--- a/src/components/existing-request/existing-request-display.vue
+++ b/src/components/existing-request/existing-request-display.vue
@@ -427,10 +427,10 @@ export default class ExistingRequestDisplay extends Mixins(
     return this.isFirm(this.nr) &&
            this.nr.request_action_cd !== RequestCode.CHG &&
            (NrState.APPROVED === this.nr.state ||
-            !this.isConcentRequired)
+            !this.isConsentRequired)
   }
 
-  get isConcentRequired (): boolean {
+  get isConsentRequired (): boolean {
     return NrState.CONDITIONAL === this.nr.state &&
             this.nr.consentFlag &&
             this.nr.consentFlag === 'Y'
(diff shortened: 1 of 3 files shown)


stevenc987 (approved):

src/components/existing-request/existing-request-display.vue
```diff
@@ -424,7 +424,16 @@ export default class ExistingRequestDisplay extends Mixins(
 
   /** True if the Register button should be shown. */
   get showRegisterButton (): boolean {
-    return this.isFirm(this.nr)
+    return this.isFirm(this.nr) &&
+           this.nr.request_action_cd !== RequestCode.CHG &&
+           (NrState.APPROVED === this.nr.state ||
+            !this.isConcentRequired)
+  }
```
severinbeauvais, line 431: Instead of negative logic ("not CHG"), I think it would be better to positively compare with the allowable request action code (NEW, I think).

ozamani9gh (approved):
Looks Good Eve

src/components/existing-request/existing-request-display.vue
eve-git: There are total 10 request_action_cds. I think it is better to keep Not CHG.

severinbeauvais: "Not CHG" means this might be true for 8 invalid request action codes, but we only care about the one valid code (NEW).

severinbeauvais (approved):

PR #575, opened 2022-12-19
show Payment Incomplete status in UI only when nr.state='DRAFT'

eve-git:
*Issue #:* /bcgov/entity#14774

*Description of changes:*
Add condition nr.state='DRAFT' when Payment Incomplete state to be show

### DIFF
--- a/src/components/existing-request/existing-request-display.vue
+++ b/src/components/existing-request/existing-request-display.vue
@@ -532,7 +532,8 @@ export default class ExistingRequestDisplay extends Mixins(
   }
 
   get isNotPaid () {
-    return (this.pendingPayment?.sbcPayment?.statusCode === PaymentStatus.CREATED)
+    return (this.pendingPayment?.sbcPayment?.statusCode === PaymentStatus.CREATED) &&
+            NrState.DRAFT === this.nr.state
   }
 
   get isPaymentProcessing () {
--- a/src/components/existing-request/existing-request-display.vue
+++ b/src/components/existing-request/existing-request-display.vue
@@ -533,7 +533,7 @@ export default class ExistingRequestDisplay extends Mixins(
 
   get isNotPaid () {
     return (this.pendingPayment?.sbcPayment?.statusCode === PaymentStatus.CREATED) &&
-            NrState.DRAFT === this.nr.state
+            ([NrState.DRAFT, NrState.PENDING_PAYMENT].includes(this.nr.state))
   }
 
   get isPaymentProcessing () {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "3.3.4",
+  "version": "3.3.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "3.3.4",
+      "version": "3.3.5",
       "dependencies": {
         "@babel/compat-data": "^7.12.13",
         "@bcrs-shared-components/enums": "1.0.13",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.3.4",
+  "version": "3.3.5",
   "private": true,
   "appName": "Name Request UI",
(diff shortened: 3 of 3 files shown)


src/components/existing-request/existing-request-display.vue
```diff
@@ -532,7 +532,8 @@ export default class ExistingRequestDisplay extends Mixins(
   }
 
   get isNotPaid () {
-    return (this.pendingPayment?.sbcPayment?.statusCode === PaymentStatus.CREATED)
+    return (this.pendingPayment?.sbcPayment?.statusCode === PaymentStatus.CREATED) &&
+            NrState.DRAFT === this.nr.state
```
severinbeauvais, line 536: In which cases can the payment be CREATED?

What will be displayed if the payment is created but the NR is not DRAFT?

ozamani9gh (approved):
simple change, i am fine with it.

src/components/existing-request/existing-request-display.vue
eve-git: Add condition nr.state = 'PENDING-PAYMENT' to isNotPaid to show Payment Incomplete incase of incomplete payment of nr.

severinbeauvais:
Did you mean for `amichard` to be a reviewer, or did you mean `rarmitag`?

src/components/existing-request/existing-request-display.vue
eve-git: A user has payment only when first request nr and upgrade nr, so:
when requesting NR with incomplete payment, the status = 'Payment Incomplete'
when upgrading NR with incomplete payment, the status = 'Payment Incomplete'

for other NR state, such as 'Approved', it means nr has been processed by a staff. There is no sense to show 'Payment Incomplete'( in case of upgrade). Do I miss anything?

severinbeauvais: Thanks for looking into it.

What about an expired NR that a user resubmits and the payment fails?

severinbeauvais: My concern is that `isNotPaid()` is one link in an if-then-else chain, so if it now returns False, is the correct thing displayed?

eve-git: NrState.REFUND_REQUESTED: Cancelled
PaymentStatus.APPROVED, PaymentStatus.COMPLETED: Processing Payment
NrState.CONSUMED: Conditional Approval or Approved
NrState.APPROVED: Approved
NrState.CANCELLED: Cancelled
NrState.CONDITIONAL: Conditional Approval
NrState.DRAFT: 'Pending Staff Review'  ---- with complete payment
NrState.EXPIRED: Expired
NrState.HOLD:  In Progress
NrState.INPROGRESS: In Progress
NrState.REJECTED:  Rejected

severinbeauvais: I believe you have added an extra condition to `isNotPaid()` because something was not working correctly. In other words, there are some scenarios where the payment can be CREATED but we don't want to show Payment Incomplete. What are those scenarios, and what is displayed for them now?

eve-git: if incomplete payment for a request, we don't give the user NR number and only 'repay' button available. "Payment Incomplete" status shows in the case. 
But for a Incomplete payment of Upgrade, 'Payment Incomplete' shows even after the NR approved or expired. I don't think the user wants to see "Payment Incomplete" status when 'Upgrade' button is not available.

severinbeauvais: So what will display for a incomplete payment of Upgrade (or Resubmit)?

eve-git: for the incomplete payment of Upgrade, the same as before, 'Incomplete Payment' is the status
for the incomplete payment of Resubmit, the same as before, 'Expired' is the status. A new NR number will be created with PENDING_PAYMENT status when user tries to pay for the resubmit. However, the user won't know the new NR. After a successful payment of resubmit, the user sees the new NR in UI.

severinbeauvais: OK, that sounds good.

severinbeauvais (approved):

PR #585, opened 2023-02-27
15183 Imported Genesys Web Chat

JazzarKarim:

*Description of changes:*
- Imported the new Genesys Web Chat Component.
name request genesys


src/App.vue
```diff
@@ -9,6 +9,13 @@
                  :webChatStatusUrl = "window['webChatStatusUrl']"
       />
 
+      <!-- Display the Genesys WebMessage -->
+      <GenesysWebMessage
+        genesysURL="https://apps.cac1.pure.cloud/genesys-bootstrap/genesys.min.js"
+        environmentKey="cac1"
+        deploymentKey="5c012aaa-1f17-4d0b-8da2-bea9c1133d5d"
+      />
```
severinbeauvais, line 17: As in Create UI, don't hard-code these -- use values from configuration.json.

severinbeauvais (approved):
I think this is OK... but remember to update the app version.

severinbeauvais (commented):
You're also missing the new component import :)

JazzarKarim:
> I think this is OK... but remember to update the app version.

I did update it! From 3.3.8 to 3.3.9. Or do you mean something else?

JazzarKarim:
> You're also missing the new component import :)

Which component are you talking about Sev? 😟

severinbeauvais:
> I did update it! From 3.3.8 to 3.3.9. Or do you mean something else?

> Which component are you talking about Sev? 😟 

Oh, I see... First, I missed the package files. Second, you cheated and didn't run `npm i`, so the shared component wasn't installed (as seen in the check errors above). You should always run `npm i` and never update the package lock manually.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "3.3.7",
```
severinbeauvais, line 3: From 7 to 9?

JazzarKarim:
> > I did update it! From 3.3.8 to 3.3.9. Or do you mean something else?
> 
> > Which component are you talking about Sev? 😟
> 
> Oh, I see... First, I missed the package files. Second, you cheated and didn't run `npm i`, so the shared component wasn't installed (as seen in the check errors above). You should always run `npm i` and never update the package lock manually.

I did to that actually first but I reverted the change. I looked at previous commits and I thought that this is how you update after doing something since that's what's everybody was doing. So sorry about that. Good to know.

package-lock.json
JazzarKarim: There was a mistake from the previous commit. Omid forgot to update the package-lock.json.

severinbeauvais: Tsk tsk.

OK, this is fine. Thanks.

PR #586, opened 2023-02-28
15183 Added content security policy to genesys web chat

JazzarKarim:

*Description of changes:*
- Added the Content Security Policy to unblock the execution of the Maximus Web Chat Script in Dev.
genesys in dev


severinbeauvais (approved):
LGTM

severinbeauvais:
> I don't think the package and package.lock make sense to update here.

If not then how would we know whether Dev or Test or Prod have the needed configuration, other than testing if the new webchat works?

seeker25:
By watching the hash on the image, they should always be checked before deploying. 

Package.json is for the web app, you're making a nginx configuration change.. they're two different concerns.

severinbeauvais:
> Package.json is for the web app, you're making a nginx configuration change (infrastructure related).. they're two different concerns.

They are different things but the webserver config and the app that is served are bundled together.

The app version is just easy to see, in any environment, from which we can deduce whether some features/commits are present or not.

PR #587, opened 2023-02-28
removed content security policy (breaking change)

JazzarKarim:

*Description of changes:*
- Reverted the change since it's breaking dev.


severinbeauvais (approved):
You've got my blessing if you want to go this route, but if you're able to work this out with Travis without using a PR then I'm OK with that too.

PR #589, opened 2023-03-01
15183 Added feature flags for old and new webchats

JazzarKarim:

*Description of changes:*
- Added feature flags for the old webchat and the new Genesys web message component.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -273,20 +273,12 @@ export default class App extends Mixins(
 
   /** Whether the old webchat should be enabled. */
   get enableOldWebchat (): boolean {
-    if (getFeatureFlag('enable-web-chat')) {
-      return !!getFeatureFlag('enable-web-chat')
-    } else {
-      return false // Failsafe value
-    }
+    return getFeatureFlag('enable-web-chat')
   }
 
   /** Whether the Genesys web message should be enabled. */
   get enableGenesysWebMessage (): boolean {
-    if (getFeatureFlag('enable-genesys-web-message')) {
-      return !!getFeatureFlag('enable-genesys-web-message')
-    } else {
-      return true // Failsafe value
-    }
+    return getFeatureFlag('enable-genesys-web-message')
   }
 }
 </script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -273,12 +273,12 @@ export default class App extends Mixins(
 
   /** Whether the old webchat should be enabled. */
   get enableOldWebchat (): boolean {
-    return getFeatureFlag('enable-web-chat')
+    return !!getFeatureFlag('enable-web-chat')
   }
 
   /** Whether the Genesys web message should be enabled. */
   get enableGenesysWebMessage (): boolean {
-    return getFeatureFlag('enable-genesys-web-message')
+    return !!getFeatureFlag('enable-genesys-web-message')
   }
 }
 </script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,8 @@
 <template>
   <v-app id="app">
     <div id="main-column">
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -268,6 +270,24 @@ export default class App extends Mixins(
...
     await this.toggleConfirmNrModal(false)
   }
+
+  /** Whether the old webchat should be enabled. */
+  get enableOldWebchat (): boolean {
+    if (getFeatureFlag('enable-web-chat')) {
+      return !!getFeatureFlag('enable-web-chat')
+    } else {
+      return false // Failsafe value
+    }
```
severinbeauvais, line 280: 1. Please simplify this. Also below.

2. Add the failsafe values to https://github.com/bcgov/namerequest/blob/main/src/plugins/featureFlags.ts

src/App.vue
```diff
@@ -268,6 +270,24 @@ export default class App extends Mixins(
...
+  }
+
+  /** Whether the Genesys web message should be enabled. */
+  get enableGenesysWebMessage (): boolean {
+    if (getFeatureFlag('enable-genesys-web-message')) {
+      return !!getFeatureFlag('enable-genesys-web-message')
+    } else {
+      return true // Failsafe value
```
severinbeauvais, line 288: Should also fail as False.

severinbeauvais (commented):
Logic looks good. Remember the app version :)

src/App.vue
severinbeauvais: "No special features if we can't get the FFs from LaunchDarkly -- only base functionality."

JazzarKarim: Yes, makes a lot of sense. Noted.

src/App.vue
```diff
@@ -273,20 +273,12 @@ export default class App extends Mixins(
 
   /** Whether the old webchat should be enabled. */
   get enableOldWebchat (): boolean {
-    if (getFeatureFlag('enable-web-chat')) {
-      return !!getFeatureFlag('enable-web-chat')
-    } else {
-      return false // Failsafe value
-    }
+    return getFeatureFlag('enable-web-chat')
```
severinbeauvais, line 276: `getFeatureFlag()` returns a type of any. If you use `!!` then you get a boolean instead.

src/plugins/featureFlags.ts
```diff
@@ -12,7 +12,9 @@ const defaultFlagSet: LDFlagSet = {
   'disable-analysis': true, // by default, analysis is disabled
   'hardcoded_regular_wait_time': 0, // by default, use actual wait time
   'hardcoded_priority_wait_time': 0, // by default, use actual wait time
-  'banner-text': '' // by default, there is no banner text
+  'banner-text': '', // by default, there is no banner text
+  'enable-web-chat': false, // by default, old webchat is disabled
+  'enable-genesys-web-message': false // by default, genesys web message is disabled
```
severinbeauvais, line 17: 👍

severinbeauvais (approved):
LGTM (after any->boolean fix)

src/App.vue
JazzarKarim: You're right! I was following like this here: https://github.com/bcgov/namerequest/blob/e279a02b010ffaafe8092b69267a070dddb3c8fa/src/components/common/applicant-info-2.vue#L280
I'll fix it.

severinbeauvais: Hah, that other code should be fixed :)

(I think it evaluates the same, but typing is helpful to prevent errors early on.)

JazzarKarim: Got it. Shall I just add the `!!` since I'm already on the subject of feature flags or leave it for another ticket?

severinbeauvais: Sure, fix it.

PR #590, opened 2023-03-02
sentry errors related to keycloak calls

bolyachevets:
*Issue: /bcgov/entity/issues/6970

*Description of changes:
The main loop in the vue app tries to sync the token every time changes are detected, i.e. before we are redirected to /signin and after we are redirected to /signout the path is /namerequest and we call initializeToken.  Majority of still relevant sentry errors take place during logout (i.e. at https://www.bcregistry.ca/namerequest/signout -  see full list of errors here https://registries.sentry.io/issues/2098686855/events/?project=5433955&referrer=issue-stream&statsPeriod=14d). The errors take place right after the session is cleared and PREVENT_STORAGE_SYNC flag is set during the logout. Checking for PREVENT_STORAGE_SYNC the flag should prevent us from fetching userInfo when we don’t have the token, as auth-api returns 401 error if auth token is not provided in /users/@me calls (authorization_header_missing errors can be observed in the developer tools console)

### DIFF
--- a/src/services/auth.services.ts
+++ b/src/services/auth.services.ts
@@ -54,12 +54,19 @@ export default class AuthServices {
    */
   static async fetchUserInfo (): Promise<any> {
     const url = `${this.authApiUrl}/users/@me`
-
-    return axios.get(url)
-      .then(response => {
-        if (response?.data) return response.data
-        throw new Error('Invalid user info')
-      })
+    if (!sessionStorage.getItem('PREVENT_STORAGE_SYNC')) {
+      const token = sessionStorage.getItem('KEYCLOAK_TOKEN')
+      if (token) {
+        const headers = {
+          Authorization: `Bearer ${token}`
+        }
+        return axios.get(url, { headers: headers })
+          .then(response => {
+            if (response?.data) return response.data
+            throw new Error('Invalid user info')
+          })
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/services/auth.services.ts
```diff
@@ -54,12 +54,19 @@ export default class AuthServices {
...
+        const headers = {
+          Authorization: `Bearer ${token}`
+        }
+        return axios.get(url, { headers: headers })
+          .then(response => {
+            if (response?.data) return response.data
+            throw new Error('Invalid user info')
+          })
+      }
+    }
```
severinbeauvais, line 69: This function doesn't return anything if the condition on line 57 is False.

It might be cleaner to change line 57 to something like this:
`if (!sessionStorage.getItem('PREVENT_STORAGE_SYNC')) return null`

Note1: Please add a comment above line 57 explaining something about PREVENT_STORAGE_SYNC and why we're checking it here.

Note2: In my sample above, I return null, so you might want to change the check in update-user-mixin.ts to `if (!userInfo)`.

severinbeauvais (commented):
In order to track your fix through Dev/Test/Prod, please update the app version in package.json and run `npm i`. Then check in both package files.

severinbeauvais (approved):
LGTM

pwei1018 (approved):

PR #597, opened 2023-04-27
Retrieve NR detail through account information.

pwei1018:
*Issue #:* /bcgov/entity### https://github.com/bcgov/entity/issues/16003

*Description of changes:*
1. Add a new function to retrieve NR detail through account information (token, nr id and account id).
2. Modify landing page to switch using new function.

### DIFF
--- a/src/services/namex.services.ts
+++ b/src/services/namex.services.ts
@@ -273,6 +273,35 @@ export default class NamexServices {
     }
   }
 
+  static async getNameRequestByToken (handleError: boolean, nrId:string, accountId:string): Promise<NameRequestI> {
+    try {
+      const token = sessionStorage.getItem('KEYCLOAK_TOKEN')
+      if (token) {
+        const headers = {
+          Authorization: `Bearer ${token}`,
+          'Accept': 'application/pdf'
+        }
+
+        // call namerequest get endpoint to retrieve NR detail, also verify affiliation
+        const response = await this.axios.get(`${this.namexUrl()}/namerequests/${nrId}?org_id=${accountId}`, {
+          headers
+        })
+        if (response?.status === OK && response?.data) return response.data
+        throw new Error(`Invalid response = ${response}`)
+      } else {
+        // if without token then check the session storage
+        return await this.getNameRequest(handleError)
+      }
+    } catch (err) {
+      if (handleError) {
+        const msg = await this.handleApiError(err, 'Could not get name request')
+        console.error('getNameRequest() =', msg) // eslint-disable-line no-console
+        await errorModule.setAppError({ id: 'get-name-request-error', error: msg } as ErrorI)
+      }
+      return null
+    }
+  }
+
   static async getNameRequest (handleError: boolean): Promise<NameRequestI> {
     try {
       const { CancelToken } = Axios
--- a/src/views/landing.vue
+++ b/src/views/landing.vue
@@ -73,7 +73,7 @@ export default class Landing extends Vue {
 
   async mounted () {
     const { id } = this
-    const accountId = this.$route.query.accountid.toString()
+    const accountId = this.$route.query.accountid?.toString()
 
     // if an id and accountid was specified then get and load the subject NR
     if (id && accountId) {
--- a/src/views/landing.vue
+++ b/src/views/landing.vue
@@ -73,7 +73,7 @@ export default class Landing extends Vue {
 
   async mounted () {
     const { id } = this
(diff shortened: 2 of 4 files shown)


pwei1018:
Temporary Url for review: https://namerequest-dev--pr-597-0wzpnfsj.web.app

src/views/landing.vue
```diff
@@ -73,9 +73,11 @@ export default class Landing extends Vue {
 
   async mounted () {
     const { id } = this
-    // if an id was specified then get and load the subject NR
-    if (id) {
-      const nrData = await NamexServices.getNameRequest(true)
+    const accountId = this.$route?.query?.accountid?.toString()
```
severinbeauvais, line 76: When is `$route` ever not an object?

src/views/landing.vue
```diff
@@ -73,9 +73,11 @@ export default class Landing extends Vue {
 
   async mounted () {
     const { id } = this
-    // if an id was specified then get and load the subject NR
-    if (id) {
-      const nrData = await NamexServices.getNameRequest(true)
```
severinbeauvais, line 78: Does this mean we can delete `getNameRequest` from here and from the API now?

pwei1018: No, for the anonymous user still need to use getNameRequest function to retrieve NR detail information.

src/services/namex.services.ts
```diff
@@ -273,6 +273,35 @@ export default class NamexServices {
...
+
+        // call namerequest get endpoint to retrieve NR detail, also verify affiliation
+        const response = await this.axios.get(`${this.namexUrl()}/namerequests/${nrId}?org_id=${accountId}`, {
+          headers
+        })
+        if (response?.status === OK && response?.data) return response.data
+        throw new Error(`Invalid response = ${response}`)
+      } else {
+        // if without token then check the session storage
+        return await this.getNameRequest(handleError)
```
severinbeauvais, line 293: I don't understand how this "checks the session storage". It looks like it does the same GET as above but without the org_id (and it does something with a cancel token).

If you figure out exactly how this works, please add comments.

src/views/landing.vue
pwei1018: Don't know.... I just copied it from other spot.

src/views/landing.vue
severinbeauvais: So why add the call with the org_id, if it's not needed?

src/views/landing.vue
severinbeauvais: I believe it should never be null (even in the other spot), except if a unit test runs without setting it first.

Conditional chaining is great except it can hide some errors. For example, you might have a fallback action if the chain is undefined, but maybe you should be doing a different action if the first part of the chain is valid. That's why I don't like to have question marks on all parts of a chain.

src/services/namex.services.ts
pwei1018: In business dashboard (auth-web), it will set the NR number, email etc information into session storage and the namerequest will search NR details by for these information. However, when we migrate to new domain, this way doesn't work. The token (cancel token) is never been used in the namerequests get endpoint.

image

src/views/landing.vue
pwei1018: For login user, we need the org_id to check the affiliation instead of passing around the session storage values between multiple sites.

ozamani9gh (approved):
It looks good to me Patrick

src/views/landing.vue
severinbeauvais: If we can fetch a NR without the org_id (for anonymous users), why pass the org_id here?

src/services/namex.services.ts
pwei1018: not right now. The namex api need to be upgrade completely.

severinbeauvais: OK

src/views/landing.vue
pwei1018: How about now?

severinbeauvais: What changed?

My question is, why pass the org_id if you don't have to?

pwei1018: The landing page route:

image

if the page with nr id in the route, we expect to see the org id as well. So that we can verify the affiliation with token, org id and nr number together.

severinbeauvais: OK.

severinbeauvais (approved):

PR #603, opened 2023-05-26
15817 base url replaced

Madhavkapoor28:
*Issue #:* /bcgov/entity#15817 

*Description of changes:*

The usage of BASE_URL was confusing. It was stored as an environment variable, as well as some applications set the BASE_URL to session storage , which complicates things. The goal was to remove this variable while making sure all features are still functioning  

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "4.1.4",
+  "version": "4.1.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "4.1.4",
+      "version": "4.1.5",
       "dependencies": {
         "@babel/compat-data": "^7.12.13",
         "@bcrs-shared-components/enums": "1.0.13",
--- a/src/components/existing-request/existing-request-display.vue
+++ b/src/components/existing-request/existing-request-display.vue
@@ -206,7 +206,7 @@
           />
 
           <!-- incorporate button -->
-          <div class="mt-5 text-center" v-if="!showIncorporateButton">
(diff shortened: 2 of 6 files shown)


rarmitag (commented):
Where you define the customBaseURL, shouldn't that be read from an environment variable like process.env.BASE_URL  and not hard coded?
@pwei1018 - how do you think he should set the value?

package-lock.json
ozamani9gh, line 1: When we update version number we change package.json   not package-lock.json.    please update version there.

after you do that you can do an npm install or npm i    and it will update the lock file as well.

src/router/router.ts
ozamani9gh, line 1: we should not hardcode this value, customBaseURL (line 16)

we need to pull any urls from ENV or some other global area

src/components/existing-request/existing-request-display.vue
ozamani9gh, line 1: not sure why line 708 you removed the process.env.BASE_URL and did not use appBaseURL instead.    if you dont need it, can problem remove the import statement on line 235

ozamani9gh (changes requested):

src/components/existing-request/existing-request-display.vue
Madhavkapoor28: removed the import statement on line 235, was not required.

src/router/router.ts
Madhavkapoor28: Done. Getting URL from env

Madhavkapoor28:
@ozamani9gh @rarmitag I've implemented the changes requested by both of you. Please review once again and let me know if any further changes are required.

ozamani9gh (approved):
Looks much better Madhav.... Great Job.

rarmitag (approved):

PR #616, opened 2023-06-29
16634 Implemented new actions menu + cleanup

severinbeauvais:
*Issue #:* bcgov/entity#16634

*Description of changes:*
    - app version = 5.0.2
    - changed RequestCode to shared NrRequestActionCodes enum
    - fixed some whitespace and alignment
    - fixed some property/method visibilities
    - cleanup up Pick Request Type dialog (removed blurbs, reformatted layout, filtered items)
    - implemented new v-select with list items for nested menus
    - implemented NameType enum for extraprovincial / assumed names
    - updated RequestActions data structure to support nested groups (and new text, etc)
    - fixed mobile spacing above nav buttons
    - fixed mobile spacing above existing NR buttons
    - misc cleanup
    - deleted obsolete code


src/components/new-request/search.vue
```diff
@@ -394,14 +419,6 @@ export default class NewSearch extends Mixins(CommonMixin) {
     return this.$conversionTypes.find(conversion => conversion.value === this.getConversionType)?.text
   }
 
-  get inputCompClass () {
-    let errorTypes = ['entity_type_cd', 'request_action_cd', 'location']
-    if (errorTypes.some(type => this.getErrors.includes(type))) {
-      return 'mt-n5'
-    }
-    return 'mt-n2'
-  }
```
severinbeauvais, line 403: This was used with the old version of Vuetify, but it no longer needed. This is now replaced by new line 164, which works in all cases.

src/components/new-request/search.vue
```diff
@@ -1,43 +1,46 @@
 <template>
   <v-container fluid id="new-request-container" class="copy-normal pa-10">
     <v-row no-gutters>
-      <v-col cols="12" class="pt-0 font-weight-bold h6"><span>I need a name to:</span></v-col>
+      <v-col cols="12" class="pt-0 font-weight-bold h6">
+        I need a name to: {{ request_action_cd }} / {{ location }} / {{ entity_type_cd }}
```
severinbeauvais, line 5: These 3 codes are for debugging only. I'll remove them before closing off my ticket.

src/interfaces/models.ts
```diff
@@ -201,11 +201,13 @@ export interface RequestActionMappingI {
 }
 
 export interface RequestActionsI {
-  text: string
-  value: RequestCode
-  blurbs?: string
+  group?: number
+  isHeader?: boolean
```
severinbeauvais, line 205: Added group number.
Added isHeader.
Added subtext.
Deleted blurbs.

src/list-data/request-actions.ts
```diff
@@ -1,63 +1,78 @@
...
+  },
+  {
+    group: 0,
+    text: 'Start a new BC-based business',
+    subtext: 'Incorporation or Registration',
     shortDesc: 'New Request',
-    value: RequestCode.NEW,
-    blurbs: `Create a new business in British Columbia or register a business you formed in another province or
-              territory, country or federal jurisdiction so that you may also conduct business here in BC.`
+    value: NrRequestActionCodes.NEW_BUSINESS
```
severinbeauvais, line 16: The new structure is actually fairly similar to the old one -- it provides a Request Action Code that is used in the rest of the flows (eg, NEW, MVE, CHG, AML, CNV).

src/list-data/request-actions.ts
```diff
@@ -1,63 +1,78 @@
...
+    text: 'For businesses that already exist outside BC:'
   },
-  // COMMENTED OUT FOR FUTURE IMPLEMENTATION
-  // {
-  //   text: 'Assume a New Name in BC',
-  //   shortDesc: 'Assumed Name Request',
-  //   value: RequestCode.ASSUMED,
-  //   blurb: `You have an existing business in another province. You are closing your business there and moving your
-  //           business to BC, however, the name of your business is already in use in BC`
-  // },
```
severinbeauvais, line 25: This is an obsolete primary action, but it is dynamically added as a Request Type Option if the Request Type Code is AS/AL/XASO/XCASO/UA.

This is existing code but I'll try to untangle it a bit more as part of this ticket (possibly in a future PR). (See also Name Type enum above.)

src/list-data/request-actions.ts
```diff
@@ -1,63 +1,78 @@
...
   // {
-  //   text: 'Restore with a New Name',
+  //   group: 2,
+  //   text: 'Restore a business that has been dissolved or cancelled',
+  //   subtext: 'Restoration or Reinstatement',
   //   shortDesc: 'Restore-New Request',
-  //   value: RequestCode.REN,
-  //   blurb: `You have a business that has been dissolved or cancelled. You want to start up again with a new name.
-  //           You will need your incorporation or firm number assigned to you by Registries.`
+  //   value: NrRequestActionCodes.RENEW
```
severinbeauvais, line 76: No change -- this is still for future implementation, maybe.

src/components/new-request/search.vue
```diff
@@ -56,16 +59,16 @@
                         filled
                         label="Select a Jurisdiction"
                         v-model="location">
-                <template slot="item" slot-scope="data">
```
severinbeauvais, line 59: This showed an error in VS Code, "Property X does not exist on type Y."

The new format works correctly.

src/enums/name-type.ts
eve-git, line 1: Can the enum name NameType more specific since it is for extraprovincial name request?

severinbeauvais: Good idea. How about `XproNameType` ?

severinbeauvais: By the way, I'm putting together the pieces (like this enum and what it means) from looking through code to see how the previous hard-coded strings were used. I think I'm right about this but if you know otherwise then please let me know.

eve-git: The code looks good to me.

src/interfaces/models.ts
```diff
@@ -217,7 +220,7 @@ export interface RequestNameI {
   consent_words: '' | string[]
   conflict1: string
   conflict1_num: string
-  request_type_cd?: string
+  request_type_cd?: NameType
```
eve-git, line 223: I doubt if the request_type_cd can only be ex-pro name type here?

severinbeauvais: You're probably right but it's not possible to confirm this because of the type-less assignments, eg,
```
  data['request_type_cd'] = <something>
```

I have reverted these declarations back to strings in the next commit.

eve-git (approved):
Looks good to me

src/components/dialogs/pick-request-type.vue
```diff
@@ -1,47 +1,34 @@
...
+        <h4>What would you like to do?</h4>
+        <v-btn icon large class="dialog-close" @click="showModal = false">
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
+      </v-card-title>
+
+      <v-card-text>
+        <v-row no-gutters>
+          <v-col cols="6" class="d-block">
+            <ul v-for="(item, i) in tableData.col1" :key="`col1-${i}`">
```
JazzarKarim, line 17: Question regarding the key here. If you don't have it, you get a "Elements in iteration expect to have 'v-bind:key' directives" error right?

src/components/new-request/search.vue
```diff
@@ -1,43 +1,46 @@
...
+          label="Select an Action"
+          :error-messages="getErrors.includes('request_action_cd') ? 'Please select an action' : ''"
+          :items="requestActions"
+          :menu-props="{ bottom: true, offsetY: true, maxHeight: 423 }"
+          @change="clearErrors()"
+        >
+          <template v-slot:item="{ item }">
+            <v-list-item-title
+              v-if="item.isHeader"
+              class="group-header d-flex justify-space-between align-center"
```
JazzarKarim, line 23: What do you think of creating a class for this in the style block Sev? Would it be nicer?

src/components/dialogs/pick-request-type.vue
severinbeauvais: Something like that. All v-for items need a key to distinguish them and for reactivity to work (eg, list changes).

JazzarKarim: OK, thanks. Resolving.

src/components/new-request/search.vue
severinbeauvais: I think it would be worse, as I would have to code in all the CSS that the Vuetify helper classes provide here.

JazzarKarim (approved):
LGTM! Lots of learning for me here 😄 

Btw, it's not possible to do a `gcbrun` on Name Request right?

severinbeauvais:
> LGTM! Lots of learning for me here 😄
> 
> Btw, it's not possible to do a `gcbrun` on Name Request right?

Looks like it didn't do anything.

I'll follow up with Patrick next week to enable it, or be able to deploy this feature branch some other way.

severinbeauvais:
> LGTM! Lots of learning for me here 😄
> 
> Btw, it's not possible to do a `gcbrun` on Name Request right?

Looks like it didn't do anything.

I'll follow up with @pwei1018 next week to enable it, or be able to deploy this feature branch some other way.

PR #622, opened 2023-07-06
16634 mapped EntityType to CorpTypeCd + cleanup

severinbeauvais:
*Issue #:* bcgov/entity#16634

*Description of changes:*

Commit 1:
- updated browserslist config
- app version = 5.0.3
- added corp-type-module dependency
- mapped EntityType to CorpTypeCd

Commit 2:
- renamed xxx.services -> xxx-services
- changed some strings to EntityType
- worked around EntityType type issue
- sorted items in entityTypeCdToText()


severinbeauvais:
Sorry about the large list of reviewers, but this is a small (but possibly significant) change that I think you should know about, since you are touching, or have recently touched, this repo. Thx.

.browserslistrc
```diff
@@ -1,3 +1,3 @@
 > 1%
 last 2 versions
-not ie <= 10
+not dead
```
severinbeauvais, line 3: Same as other UI repos.

package-lock.json
```diff
@@ -1,14 +1,15 @@
 {
   "name": "name-request",
-  "version": "5.0.2",
+  "version": "5.0.3",
   "lockfileVersion": 2,
```
severinbeauvais, line 4: Note that someone, somewhere along the line, did a `npm i` while on Node < v16, which changed the lockfile to version 1 in main branch. This is a feature branch and it's OK here. At some point, main branch should be fixed, or maybe this can wait until this feature branch is merged into main in about a month.

src/enums/entity-type.ts
severinbeauvais, line 1: I made these changes to be explicit about the entity type codes used in Namerequest vs. LEAR.

If you look at [corp-type-module.ts](https://github.com/bcgov/bcrs-shared-components/blob/main/src/modules/corp-type-module/corp-type-module.ts) then you will see that LEAR knows about some NR-only codes, but below you can see there are others. As far as I know, these haven't caused problems between NR and LEAR, but some day they might. And now it's easier to see the difference between the two.

severinbeauvais:
For some reason, TS and eslint are happy with this code, but the build process isn't. I'm trying to resolve this. Don't merge yet.

image

src/enums/entity-type.ts
```diff
@@ -1,29 +1,31 @@
...
-  CR = 'CR',
-  DBA = 'DBA',
-  FI = 'FI',
-  FR = 'FR',
-  GP = 'GP',
-  LL = 'LL',
-  LP = 'LP',
-  PA = 'PA',
-  PAR = 'PAR',
+  BC = CorpTypeCd.BC_COMPANY,
```
JazzarKarim, line 6: This makes things much easier and cleaner for proceeding with the way of navigating work 😁

I think we'll have to change the value of the `supported-incorporation-registration-entities` FF we created. Also, I'm going to make the necessary changes in my two tickets to accompany this adjustment after you merge Sev 👍

severinbeauvais: OK. Let me know about that existing FF, and whether we need another one for something different.

I haven't figured out the build issue yet...

severinbeauvais:
> For some reason, TS and eslint are happy with this code, but the build process isn't. I'm trying to resolve this. Don't merge yet.
> 
> image

It seems the build process doesn't want to mix enum-strings with pure-strings, so I did this and it's happy with it:
```
// Namerequest-only entity types
enum Other {
  SO = 'SO',
  XUL = 'XUL',
  RLC = 'RLC',
  XLP = 'XLP',
  XLL = 'XLL',
  XSO = 'XSO',
  INFO = 'INFO'
}

/** Enum for Entity Type codes. */
export enum EntityType {
  // BC Entity Types:
  BC = CorpTypeCd.BC_COMPANY,
  ...
  SO = Other.SO,
```

In other words, the EntityType enum takes values from other enums, instead of a mix of enum/string.

severinbeauvais:
I have tested this locally and it works fine.

src/mixins/common-mixin.ts
```diff
@@ -87,6 +89,7 @@ export class CommonMixin extends Vue {
   isFirm (nr: any): boolean {
     return (
       nr?.legalType === EntityType.SP ||
+      nr?.legalType === EntityType.DBA ||
```
severinbeauvais, line 92: Not sure if this was accidentally missing but it can't hurt to have it. 🤞

JazzarKarim (approved):
LGTM 👍

chenhongjing (approved):
LGTM 🐱

Madhavkapoor28:
LGTM!

PR #623, opened 2023-07-07
More Corp Type Code updates

severinbeauvais:
*Issue #:* None

*Description of changes:*
- app version = 5.0.4
- imported latest shared corp-type-module
- imported latest shared interfaces
- fixed some type comparison issues
- deleted unneeded Other enum
- added CorpTypeCd mappings to EntityType
- fixed other property type errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.0.3",
+  "version": "5.0.4",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
@@ -14,10 +14,10 @@
   },
   "dependencies": {
     "@babel/compat-data": "^7.12.13",
-    "@bcrs-shared-components/corp-type-module": "1.0.12",
-    "@bcrs-shared-components/enums": "1.0.40",
+    "@bcrs-shared-components/corp-type-module": "1.0.13",
+    "@bcrs-shared-components/enums": "1.0.41",
     "@bcrs-shared-components/genesys-web-message": "1.0.0",
-    "@bcrs-shared-components/interfaces": "1.0.22",
+    "@bcrs-shared-components/interfaces": "1.0.65",
     "@bcrs-shared-components/staff-payment": "1.0.29",
     "@bcrs-shared-components/web-chat": "2.0.0",
     "@mdi/font": "^4.5.95",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -14,10 +14,10 @@
   },
   "dependencies": {
     "@babel/compat-data": "^7.12.13",
-    "@bcrs-shared-components/corp-type-module": "1.0.12",
-    "@bcrs-shared-components/enums": "1.0.40",
+    "@bcrs-shared-components/corp-type-module": "1.0.13",
+    "@bcrs-shared-components/enums": "1.0.41",
     "@bcrs-shared-components/genesys-web-message": "1.0.0",
-    "@bcrs-shared-components/interfaces": "1.0.22",
+    "@bcrs-shared-components/interfaces": "1.0.65",
```
severinbeauvais, line 20: Updating enums and interfaces _should_ be safe since we almost always add to them instead of remove things. Plus, this is a feature branch, so even if this breaks something then at least it's a step ahead and we can fix it.

src/components/new-request/search.vue
```diff
@@ -400,7 +400,9 @@ export default class NewSearch extends Mixins(CommonMixin) {
       return
     }
     if (type && this.getIsConversion) {
-      let { entity_type_cd } = this.$conversionTypes.find(conv => conv.value === type)
+      // convert NrRequestTypeCodes -> EntityType
+      const value = type as unknown as NrRequestTypeCodes
+      let { entity_type_cd } = this.$conversionTypes.find(conv => conv.value === value)
```
severinbeauvais, line 405: This is the local setter for `entity_type_cd` (used for v-model props). However, depending on the menu, this value may be from the EntityType enum or from the NrRequestTypeCodes enum. This fixes the comparison error, allowing the code to find the conversion type with this value, extract the entity type code from it, and then set it in the store.

src/store/getters.ts
```diff
@@ -337,7 +337,7 @@ export const getConversionTypeOptions = (state: StateIF): ConversionTypesI[] =>
     options = options.concat(getConversionTypeAddToSelect(state))
     n = 4
   }
-  options = options.concat({ text: 'View all Alterations', value: 'INFO', rank: n })
+  options = options.concat({ text: 'View all Alterations', value: 'INFO' as any, rank: n })
```
severinbeauvais, line 340: The type of `value`, an enum, doesn't have the INFO value, so we can do this.

It's better to declare an enum value for INFO as in EntityType enum, but I'm not going to re-write all of Namerequest UI 😛

JazzarKarim (approved):
LGTM 👍

chenhongjing (approved):
LGTM 🐱

PR #624, opened 2023-07-10
16634 Fixed clickable and hover areas + fixed items with same value + fixed blurbs

severinbeauvais:
*Issue #:* bcgov/entity#16634

*Description of changes:*

Commit 1:
    - app version = 5.0.5
    - updated v-select item components to fix clickable areas and hover areas
    - removed unneeded text colour class

Commit 2:
    - used combined group+value for v-select to distinguish between items with the same value

Commit 3:
    - fixed location blurb (tooltip)
    - fixed entity type blurb (tooltip)

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -14,6 +14,7 @@
           label="Select an Action"
           :error-messages="getErrors.includes('request_action_cd') ? 'Please select an action' : ''"
           :items="requestActions"
+          item-value="[group,value]"
           :menu-props="{ bottom: true, offsetY: true, maxHeight: 423 }"
           @change="clearErrors()"
         >
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -18,24 +18,27 @@
           @change="clearErrors()"
         >
           <template v-slot:item="{ item }">
-            <v-list-item-title
+            <v-list-item-content
               v-if="item.isHeader"
-              class="group-header d-flex justify-space-between align-center"
+              class="group-header px-4 py-5"
               @click.stop="toggleActionGroup(item.group)"
             >
-              <div class="app-blue mr-4">{{ item.text }}</div>
-              <v-icon color="primary">
-                {{ item.group === activeActionGroup ? 'mdi-chevron-up' : 'mdi-chevron-down' }}
-              </v-icon>
-            </v-list-item-title>
-            <v-list-item-title
+              <div class="d-flex justify-space-between align-center">
+                <p class="mb-0 mr-4" :class="{'app-blue': item.group === activeActionGroup}">{{ item.text }}</p>
+                <v-icon color="primary">
+                  {{ item.group === activeActionGroup ? 'mdi-chevron-up' : 'mdi-chevron-down' }}
+                </v-icon>
+              </div>
+            </v-list-item-content>
+
+            <v-list-item-content
               v-else
-              class="group-item ml-2 colour-text"
+              class="group-item pl-8 pr-4 py-4"
               @click="request_action_cd = item.value"
             >
               <div class="font-weight-bold">{{ item.text }}</div>
               <div>{{ item.subtext }}</div>
-            </v-list-item-title>
+            </v-list-item-content>
           </template>
         </v-select>
       </v-col>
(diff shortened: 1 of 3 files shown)


src/components/new-request/search.vue
```diff
@@ -18,24 +18,27 @@
           @change="clearErrors()"
         >
           <template v-slot:item="{ item }">
-            <v-list-item-title
+            <v-list-item-content
```
severinbeauvais, line 21: This "hybrid" menu is a v-select with custom v-list items in it. As this group header is clickable (not like a title), this seems to be a better sub-component for it.

Ref: https://v2.vuetifyjs.com/en/api/v-list-item-content/
(also look at the other v-list-item-xxx items if interested)

src/components/new-request/search.vue
```diff
@@ -18,24 +18,27 @@
...
               v-if="item.isHeader"
-              class="group-header d-flex justify-space-between align-center"
+              class="group-header px-4 py-5"
```
severinbeauvais, line 23: I moved the flex container into a div below instead of overloading the previous "title" component.

Also, here is where the item's whitespace is declared. Previously, the parent v-select automatically applied some padding, which caused a non-clickable area around the group header text (below). Now, this entire group header component is clickable -- as expected.

src/components/new-request/search.vue
```diff
@@ -18,24 +18,27 @@
...
               @click.stop="toggleActionGroup(item.group)"
             >
-              <div class="app-blue mr-4">{{ item.text }}</div>
-              <v-icon color="primary">
-                {{ item.group === activeActionGroup ? 'mdi-chevron-up' : 'mdi-chevron-down' }}
-              </v-icon>
-            </v-list-item-title>
-            <v-list-item-title
+              <div class="d-flex justify-space-between align-center">
+                <p class="mb-0 mr-4" :class="{'app-blue': item.group === activeActionGroup}">{{ item.text }}</p>
```
severinbeauvais, line 27: Conditional class here so the group header stays blue when it's expanded.

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-624-dwvpgfbx.web.app

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-624-dwvpgfbx.web.app

src/components/new-request/search.vue
JazzarKarim: This fixes the scenario where when we select an action, the select jurisdiction dropdown remains disabled right Sev?

JazzarKarim (approved):
LGTM 👍

src/components/new-request/search.vue
severinbeauvais: Yes, you may have noticed that, sometimes, clicking the menu item did not actually select the item. The padding was the culprit: it caused the v-select to handle the click action instead of the item to handle the action.

JazzarKarim: Ohhh, interesting. OK Sev, makes sense.

leodube-aot (approved):
LGTM as well

src/components/new-request/search.vue
```diff
@@ -14,28 +14,32 @@
           label="Select an Action"
           :error-messages="getErrors.includes('request_action_cd') ? 'Please select an action' : ''"
           :items="requestActions"
+          item-value="[group,value]"
```
severinbeauvais, line 17: It worked the first time!

image

Ref: https://v2.vuetifyjs.com/en/api/v-select/#props-item-value

JazzarKarim: Why are you adding this prop exactly Sev?

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-624-dwvpgfbx.web.app

src/components/new-request/search.vue
severinbeauvais: To fix a funny behaviour that Yui found and reported in the ticket: if you select the first "NEW" action in the list (in first group), and then you look at the second group, its "NEW" action also appears selected.

This prop, which works along with the `items` array prop, tells the v-select which "value" to keep track of. So I created this compound value which is now unique, and the v-select now knows the difference between the first (`[0,NEW]`) and the second (`[1,NEW]`) items.

Re:
image

JazzarKarim: Oh, OK. I didn't notice that while using the component in my PR. Thanks for explaining Sev. Noted.

severinbeauvais: I didn't notice either! But that Yui feller is pretty sharp 😉

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-624-dwvpgfbx.web.app

PR #628, opened 2023-07-19
Fixed in response to Yui's comment

JazzarKarim:
*Issue #:* /bcgov/entity#16635

*Description of changes:*
- This fixes the comment Yui mentioned regarding the registration/incorporation flow. Yui's comment: https://github.com/bcgov/entity/issues/16635#issuecomment-1642807197
- The behavior is exactly like in PROD now (which is what we want according to Yui).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "5.0.7",
+  "version": "5.0.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "5.0.7",
+      "version": "5.0.8",
       "dependencies": {
         "@babel/compat-data": "^7.12.13",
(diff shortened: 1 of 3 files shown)


pwei1018:
Temporary Url for review: https://namerequest-dev--pr-628-mefl84wf.web.app

severinbeauvais (approved):
Is that how it's done elsewhere?

See line 62. Thoughts?

JazzarKarim:
> Is that how it's done elsewhere?
> 
> See line 62. Thoughts?

This is not how it's done currently in main branch. I tried to match between the two but there's too much code to dig in (plus a lot of stuff is now different due to the changes you and I have made). So, I had to improvise and come up with my own solution 😎

Also, regarding line 62, that's how I originally was trying to do it. It wasn't working though (when society is selected, entity_type_cd was not being changed).

severinbeauvais:
Oh well, it's fine then.

leodube-aot (approved):
LGTM Karim!

PR #630, opened 2023-07-24
Imported latest codes + added ULCB and ULBE

severinbeauvais:
*Issue #:* None

The reason for this change is to restore a couple of conversion records that were deleted (and also use the latest NR codes enum).

*Description of changes:*
- app version = 5.0.9
- added conversion types ULCB and ULBE from main branch
- fixed conversion desc and text capitalizations

### DIFF
--- a/src/list-data/conversion-types.ts
+++ b/src/list-data/conversion-types.ts
@@ -45,21 +45,21 @@ export const ConversionTypes: ConversionTypesI[] = [
     shortlist: false
   },
   {
-    desc: 'Unlimited Liability Company to a limited Company',
-    text: 'Unlimited Liability Company to a limited Company',
+    desc: 'Unlimited liability company to a limited company',
+    text: 'Unlimited liability company to a limited company',
     entity_type_cd: EntityType.CR,
     blurbs: [
-      'Alter business type from an unlimited Liability Company to a limited Company.'
+      'Alter business type from an unlimited liability company to a limited company.'
     ],
     value: NrRequestTypeCodes.CONVERT_ULCB,
     shortlist: false
   },
   {
-    desc: 'Unlimited Liability Company to a Benefit Company',
-    text: 'Unlimited Liability Company to a Benefit Company',
+    desc: 'Unlimited liability company to a benefit company',
+    text: 'Unlimited liability company to a benefit company',
     entity_type_cd: EntityType.BC,
     blurbs: [
-      'Alter business type from an ULC to a Benefit Company.'
+      'Alter business type from an unlimited liability company to a benefit company.'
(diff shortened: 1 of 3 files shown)


JazzarKarim (approved):

src/list-data/conversion-types.ts
```diff
@@ -43,5 +43,25 @@ export const ConversionTypes: ConversionTypesI[] = [
...
     shortlist: false
+  },
+  {
+    desc: 'Unlimited Liability Company to a limited Company',
+    text: 'Unlimited Liability Company to a limited Company',
+    entity_type_cd: EntityType.CR,
+    blurbs: [
+      'Alter business type from an unlimited Liability Company to a limited Company.'
+    ],
+    value: NrRequestTypeCodes.CONVERT_ULCB,
```
severinbeauvais, line 54: See all known NR Request Type codes here:
https://github.com/bcgov/bcrs-shared-components/blob/main/src/enums/name-request-codes.ts

Note that we don't support all of them yet, which was the source of Mihai's concern this morning. He's looking at the new codes in that table and will create tickets for missing functionality.

severinbeauvais: Background/aside: some of those new codes are used for "Societies" (which doesn't use Namerequest UI or LEAR yet, but may be added soon).

src/list-data/conversion-types.ts
```diff
@@ -43,5 +43,25 @@ export const ConversionTypes: ConversionTypesI[] = [
...
+  },
+  {
+    desc: 'Unlimited Liability Company to a Benefit Company',
+    text: 'Unlimited Liability Company to a Benefit Company',
+    entity_type_cd: EntityType.BC,
+    blurbs: [
+      'Alter business type from an ULC to a Benefit Company.'
+    ],
+    value: NrRequestTypeCodes.CONVERT_ULBE,
+    shortlist: false
```
severinbeauvais, line 65: By the way, does anyone know why these 2 records were deleted? Mihai/Janis think they should be there, which is why I restored them.

JazzarKarim: I was wondering why were they deleted in the first place. No idea Sev.

chenhongjing (approved):

src/list-data/conversion-types.ts
severinbeauvais: Ah, this is why: https://github.com/bcgov/namerequest/pull/621

JazzarKarim: Oh, this is the commit that the branch is behind on in regards to main. OK, makes sense.

PR #635, opened 2023-07-28
17152 Menu UI improvements

severinbeauvais:
*Issue #:* /bcgov/entity#17152

*Description of changes:*
    - app version = 5.0.12
    - now return item object so we don't have to find it
    - added class to disabled items
    - now include but disable items not in active group
    - hide disabled list items via CSS

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -17,8 +17,15 @@
           item-value="[group,value]"
           :menu-props="{ bottom: true, offsetY: true, maxHeight: 423 }"
           @change="setClearErrors(null); onRequestActionChange($event)"
+          return-object
         >
-          <template v-slot:item="{ item }">
+          <!-- FUTURE: use "selection" slot to format the selected business -->
+          <!-- <template #selection="{ item }">
+            <div class="font-weight-bold text-truncate">{{ item.text }}</div>
+            <div class="text-subtitle-1">{{ item.subtext }}</div>
+          </template> -->
+
+          <template #item="{ item }">
             <v-list-item-content
               v-if="item.isHeader"
               class="group-header px-4 py-5"
@@ -32,9 +39,12 @@
               </div>
             </v-list-item-content>
 
+            <!-- render but conditionally hide disabled list items, so that the v-select
+            continues to display the current selection even when a different group is active -->
             <v-list-item-content
               v-else
               class="group-item pl-8 pr-4 py-4"
+              :class="{ 'hide-me': item.disabled }"
             >
               <div class="font-weight-bold">{{ item.text }}</div>
               <div>{{ item.subtext }}</div>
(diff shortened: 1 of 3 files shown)


src/components/new-request/search.vue
```diff
@@ -17,8 +17,15 @@
           item-value="[group,value]"
           :menu-props="{ bottom: true, offsetY: true, maxHeight: 423 }"
           @change="setClearErrors(null); onRequestActionChange($event)"
+          return-object
```
severinbeauvais, line 20: This returns the entire item object instead of just its value.

src/components/new-request/search.vue
```diff
@@ -17,8 +17,15 @@
...
         >
-          <template v-slot:item="{ item }">
+          <!-- FUTURE: use "selection" slot to format the selected business -->
+          <!-- <template #selection="{ item }">
+            <div class="font-weight-bold text-truncate">{{ item.text }}</div>
+            <div class="text-subtitle-1">{{ item.subtext }}</div>
+          </template> -->
```
severinbeauvais, line 26: I played with this a little on Janis' and Ethan's request, and we decided to not pursue this yet, but I left the code here for the next dev.

src/components/new-request/search.vue
```diff
@@ -32,9 +39,12 @@
               </div>
             </v-list-item-content>
 
+            <!-- render but conditionally hide disabled list items, so that the v-select
+            continues to display the current selection even when a different group is active -->
```
severinbeauvais, line 43: If the item is not in the list then the v-select sometimes shows a blank selection (ticket 17044).

src/components/new-request/search.vue
```diff
@@ -32,9 +39,12 @@
...
             <v-list-item-content
               v-else
               class="group-item pl-8 pr-4 py-4"
+              :class="{ 'hide-me': item.disabled }"
```
severinbeauvais, line 47: Can't just use `d-none` because it's actually the parent element I need to hide.

src/components/new-request/search.vue
```diff
@@ -598,8 +609,12 @@ export default class Search extends Mixins(CommonMixin) {
   padding: 20px 8px !important;
 }
 
-// set content colour when hovering over list items
+// hide disabled list items
+::v-deep .v-list-item:has(.v-list-item__content.hide-me) {
```
JazzarKarim, line 613: This is the first time I've seen this. So `hide-me` is the conditional class you specified in the template. Pretty cool! I'll be keeping this in mind for later work 😄

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-635-hnknd6zs.web.app

src/components/new-request/search.vue
severinbeauvais: Yes, `:has()` is new-ish.

"apply the following CSS to a v-list-item that has a content element with the hide-me class"

JazzarKarim: I remember you explained it to me once. Awesome to see it in effect!

JazzarKarim (approved):
Looks great Sev! 👍

src/components/new-request/search.vue
severinbeauvais: I was being a bit cheeky with the classname, but I think I'll leave it 😆

leodube-aot (approved):
LGTM

PR #637, opened 2023-07-31
remove currentAccount from session storage if token has expired

bolyachevets:
*Issue #: https://github.com/bcgov/entity/issues/16903

*Description of changes: 

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -8,6 +8,8 @@ import KeycloakService from 'sbc-common-components/src/services/keycloak.service
 import * as Sentry from '@sentry/browser'
 import * as Integrations from '@sentry/integrations'
 import { AllDesignationsList, ConversionTypes, Designations, RequestActions, UsaStateCodes } from '@/list-data'
+import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
+import ConfigHelper from 'sbc-common-components/src/util/config-helper'
 
 // NB: order matters - do not change
 import 'quill/dist/quill.core.css'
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -104,7 +106,12 @@ async function syncSession () {
   // Auto authenticate user only if they are not trying a login or logout
   if (!isSigningIn() && !isSigningOut()) {
     // Initialize token service which will do a check-sso to initiate session
-    await KeycloakService.initializeToken(null).then(() => {}).catch(err => {
+    await KeycloakService.initializeToken(null).then(() => {
+      const token = ConfigHelper.getFromSession(SessionStorageKeys.KeyCloakToken)
+      if (!token) {
+        ConfigHelper.removeFromSession(SessionStorageKeys.CurrentAccount)
+      }
+    }).catch(err => {
```
severinbeauvais, line 114: I think it's OK but you could be please add a comment describing what this is doing?

"When the token is initialized, ..."

"If there is no token in session storage then this means ..."

severinbeauvais (commented):
Please bump up the app version.

src/main.ts
severinbeauvais: Nice, thanks.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "4.1.9",
+  "version": "4.1.10",
```
severinbeauvais, line 3: Please run `npm i` and then commit the lockfile, too.

severinbeauvais (approved):

PR #618, opened 2023-07-05
16633 Add breadcrumb + update title/subtitle + change tab title widths

leodube-aot:
*Issue #:* /bcgov/entity#16633

*Description of changes:*
- Added Breadcrumb component
- Updated subtitle to match UX Pin
- Changed width of tab title bars

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the namerequest license (Apache 2.0).

PR #641, opened 2023-08-03
17045 Fix back button on Breadcrumb shared component

leodube-aot:
*Issue #:* /bcgov/entity#17045

*Notes:*
The override causing the visual bug for the breadcrumb component comes from [overrides.scss](https://github.com/bcgov/namerequest/blob/feature-way-of-navigating/src/assets/styles/overrides.scss#L101), which overrides the `height` and `background-colour` of `v-btn`s. I am hesitant to remove the `!important` directives here since it is a global override and I am not sure what downstream effects that could have. Rather, I am moving the quick fix I had implemented for this bug in #618 (ticket [16633](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/16633)) to a more appropriate spot. Also, instead of targeting it using the sbc component id, I am using a custom class on the component to prevent any issues if the sbc id changes.

*Description of changes:*
- Move the custom breadcrumb button stylings into App.vue

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -360,10 +360,6 @@ export default class App extends Mixins(
 }
 
 .namerequest-sbc-breadcrumb {
-  .container {
-    max-width: 1360px; // should match auth-web, etc
-  }
-
   .v-btn {
     width: 28px;
     height: 28px !important;
--- a/src/App.vue
+++ b/src/App.vue
@@ -46,6 +46,7 @@
 
       <!-- Breadcrumb -->
       <Breadcrumb
+        class="namerequest-sbc-breadcrumb"
         :breadcrumbs="breadcrumbs"
       />
 
@@ -355,6 +356,18 @@ export default class App extends Mixins(
     font-weight: normal !important;
     letter-spacing: normal !important;
     font-size: .875rem !important;
+  }
+}
+
+.namerequest-sbc-breadcrumb {
+  .container {
+    max-width: 1360px; // should match auth-web, etc
+  }
+
+  .v-btn {
+    width: 28px;
+    height: 28px !important;
+    background-color: white !important;
   }
 }
 
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -19,16 +19,6 @@
   padding-right: 0 !important;
 }
 
-#breadcrumb > .container {
-  max-width: 1360px; // should match auth-web, etc
-}
-
-#breadcrumb-back-btn {
-  width: 28px;
-  height: 28px !important;
-  background-color: white !important;
-}
-
 @media only screen and (max-width: 925px) {
   html, body {
     max-width: 100%;
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -19,6 +19,10 @@
   padding-right: 0 !important;
 }
 
+.container {
+  max-width: 1360px; // should match auth-web, etc
+}
+
 @media only screen and (max-width: 925px) {
   html, body {
     max-width: 100%;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "5.0.15",
+  "version": "5.0.16",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "5.0.15",
+      "version": "5.0.16",
       "dependencies": {
         "@babel/compat-data": "^7.12.13",
         "@bcrs-shared-components/breadcrumb": "2.1.24",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.0.15",
+  "version": "5.0.16",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",


src/App.vue
```diff
@@ -46,6 +46,7 @@
 
       <!-- Breadcrumb -->
       <Breadcrumb
+        class="namerequest-sbc-breadcrumb"
```
leodube-aot, line 49: I would have liked to use an id here but it overrides the sbc assigned id and styles.

leodube-aot:
@severinbeauvais Can you add some reviewers to this please? Thanks in advance

JazzarKarim (approved):
LGTM Leo 👍

src/App.vue
```diff
@@ -358,6 +359,18 @@ export default class App extends Mixins(
   }
 }
 
+.namerequest-sbc-breadcrumb {
+  .container {
+    max-width: 1360px; // should match auth-web, etc
```
severinbeauvais, line 364: Should this be in the shared component?

src/assets/styles/layout.scss
```diff
@@ -19,16 +19,6 @@
   padding-right: 0 !important;
 }
 
-#breadcrumb > .container {
-  max-width: 1360px; // should match auth-web, etc
-}
```
severinbeauvais, line 24: Oh, yeah, this doesn't belong here.

src/assets/styles/layout.scss
```diff
@@ -19,16 +19,6 @@
...
-
-#breadcrumb-back-btn {
-  width: 28px;
-  height: 28px !important;
-  background-color: white !important;
-}
```
severinbeauvais, line 30: Wait - so this CSS broke the styling? But it fixes it if you put it in App.vue?

Could you explain why CSS properties got broken and by what?

src/App.vue
leodube-aot: This is just to align it with the edge of the SBC Auth header. If the breadcrumb component will only ever be used underneath the sbc auth header then it can most likely be moved to the shared component.

src/assets/styles/layout.scss
leodube-aot: The two specific CSS properties that break it are in overrides.scss: [height: 2.5rem !important](https://github.com/bcgov/namerequest/blob/93f82f43983165d8661024fb50ebe4d56ddd9343/src/assets/styles/overrides.scss#L106) and [background-color: $app-blue !important;](https://github.com/bcgov/namerequest/blob/93f82f43983165d8661024fb50ebe4d56ddd9343/src/assets/styles/overrides.scss#L109). The css removed from here was the temporary fix used in [16633](https://github.com/bcgov/namerequest/pull/618) (also see [this discussion](https://github.com/bcgov/namerequest/pull/618/commits/f738f6235b0b4500c9a5a94afe27bae799584d49#r1253348992)). 

I believe the confusion is because when I created this ticket, I was not sure what the problem was but believed it was in the shared component. Since the issue was discovered to be specific to this repo, and the problem was technically already fixed, I think the scope of this ticket has changed to more of a clean-up ticket.

src/App.vue
severinbeauvais: It looks to me like the Breadcrumb component looks OK in Storybook:
https://bcgov.github.io/bcrs-shared-components/?path=/story/component-breadcrumb--home-route

And I think it looks OK in the other UIs. Why is it different in Namerequest UI? Also I'd like to find another solution that doesn't require a "special number" in various places (here, auth-web, etc).

src/assets/styles/layout.scss
severinbeauvais: On the first point, I don't understand why having this CSS in layout.scss causes the problem, but moving it to App.vue fixes it.

One the second point, yes, I agree with you.

src/App.vue
leodube-aot: I dug into this a bit and it seems other UIs do indeed rely on this "special number" to set the width of the breadcrumb. It isn't immediately obvious based on the css, but you can find it in[ business-filing-ui](https://github.com/bcgov/business-filings-ui/blob/bef1b9b0d48ca308447f6205f483998ca96a7e14/src/assets/styles/layout.scss#L35), [business-edit-ui](https://github.com/bcgov/business-edit-ui/blob/ef4a03ba0fc64f7b0297fd3aa46d9aaa39eadd18/src/assets/styles/layout.scss#L35), and [business-create-ui](https://github.com/bcgov/business-create-ui/blob/d0e45f28a223a1199201a7df29e513537d30fe75/src/assets/styles/layout.scss#L37). The reason it isn't obvious is that it is just being applied to the `.container` class, but when you inspect the breadcrumb you can see that this css gets applied to it (see screenshot). I agree that I don't like using an arbitrary number. I look into a different solution for this.

Screenshot (3)

src/assets/styles/layout.scss
leodube-aot: Hey Sev, just to clarify the first point, the css removed here is not the cause of the problem. If you look at namerequest ui on the feature branch you'll see the button looks fine even without the changes from this PR. (For example this temp url from Ketaki's PR when the environment get's fixed: https://namerequest-dev--pr-639-c9rqzr4l.web.app/).

This css is the solution, I just didn't like having it in the layout.css.

src/App.vue
severinbeauvais: Shame on us 😆 

I accept the container width spec in the other UIs. I'd rather not see it nested in the Breadcrumb component.

Thanks for looking into this. It's not worth a ton of effort but I bet it will save some time in the future.

src/assets/styles/layout.scss
severinbeauvais: Oh, right! You fixed this previously. Now you're moving the CSS to where it's used. I like it. Resolving.

severinbeauvais (approved):
I'm assuming you'll set the container width somewhere appropriate, and when you do then you have my blessing to merge this :)

src/App.vue
leodube-aot: Okay sounds good, so we'd like to move it back to layout.scss like in the other projects?

severinbeauvais: Yes, but I'm assuming it's not specific to the Breadcrumbs component, right?

ketaki-deodhar (approved):

ketaki-deodhar:
Please rebase before merging as I see package version conflicts

src/App.vue
leodube-aot: Yeah that's correct, I have moved it back to layout.scss

leodube-aot:
This PR should now be ready to merge (I don't have the permissions to do this).

severinbeauvais (approved):
LGTM

severinbeauvais:
> This PR should now be ready to merge (I don't have the permissions to do this).

You do now. Please try again :)

PR #642, opened 2023-08-03
16918 Set business lookup visible

kzdev420:
*Issue #:* /bcgov/entity#16918

*Description of changes:*

Show BusinessLookup for the CHG, AML, CNV, and REH name requests only. That way it won't show for the other flows.

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -184,16 +184,6 @@
         <template>
           <!-- Search for business identifier or name if NR request action is one of [CHG, AML, CNV, REH] -->
           <BusinessLookup v-if="isBusinessLookup" />
-
-          <template v-if="businessIdentifier">
-            <dl>
-              <dt class="font-weight-bold mr-2">Business Name:</dt>
-              <dd>{{businessName}}</dd>
-
-              <dt class="font-weight-bold mr-2">Incorporation Number:</dt>
-              <dd>{{businessIdentifier}}</dd>
-            </dl>
-          </template>
         </template>
       </v-col>
     </v-row>
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -379,8 +379,6 @@ export default class Search extends Mixins(CommonMixin) {
     NrRequestActionCodes.CONVERSION
   ]
   activeActionGroup = NaN
-  businessName = ''
-  businessIdentifier = ''
 
   private mounted () {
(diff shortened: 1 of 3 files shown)


src/components/new-request/search.vue
```diff
@@ -182,10 +182,8 @@
     <v-row no-gutters>
       <v-col cols="12">
         <template>
-          <!-- Search for business identifier or name -->
-          <BusinessLookup
-            @business="businessName = $event.name; businessIdentifier = $event.identifier"
```
severinbeauvais, line 187: When you removed the event handling (assigning businessName and businessIdentifier), I thought you were also going to delete the display block below (and local variables) ...?

severinbeauvais (approved):
Looks good.

Just a question about cleaning up unused code but I don't need to re-review if you change it.

JazzarKarim (approved):
Aside from cleaning up unused variables anymore, LGTM 👍

src/components/new-request/search.vue
kzdev420: Yes

kzdev420: I removed that. But BusinessLookup is hidden when we select one of the results.
If you are okay, I am going to remove the condition block here for now.

https://github.com/bcgov/namerequest/blob/1df5453e6dc37c47293dc0c6e20913d0bcd7bc27/src/components/new-request/business-lookup.vue#L3

What do you think?

severinbeauvais: I think you should leave the Business Lookup component as functional as possible for future work, so don't disable it. Try to make it demo-able without impacting the flows that we are working on.

kzdev420: okay

src/components/new-request/search.vue
```diff
@@ -184,16 +184,6 @@
         <template>
           <!-- Search for business identifier or name if NR request action is one of [CHG, AML, CNV, REH] -->
           <BusinessLookup v-if="isBusinessLookup" />
-
-          <template v-if="businessIdentifier">
-            <dl>
-              <dt class="font-weight-bold mr-2">Business Name:</dt>
-              <dd>{{businessName}}</dd>
```
severinbeauvais, line 193: Are businessName and businessIdentifier still used by other code, or can they be deleted?

severinbeauvais: Thanks.

PR #647, opened 2023-08-10
16636 Display text as tooltip for request action item if text is too long

severinbeauvais:
*Issue #:* bcgov/entity#16636

*Description of changes:*
- app version = 5.0.19
- restored tooltip on Request Action v-select
- made tooltip conditional on whether element has clipped
- calculated whether to show tooltip
- misc cleanup
- deleted obsolete getter

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.0.18",
+  "version": "5.0.19",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -9,48 +9,62 @@
     <v-row class="mt-6">
       <!-- Request Action -->
       <v-col cols="12" md="6" class="py-0">
-        <v-select
-          filled
-          label="Select an Action"
-          :error-messages="getErrors.includes('request_action_cd') ? 'Please select an action' : ''"
-          :items="requestActions"
-          item-value="[group,value]"
-          :menu-props="{ bottom: true, offsetY: true, maxHeight: 423 }"
-          @change="setClearErrors(null); onRequestActionChange($event)"
-          return-object
+        <v-tooltip
+          top
+          content-class="top-tooltip"
+          transition="fade-transition"
+          :disabled="!showRequestActionTooltip || isMobile"
         >
-          <!-- FUTURE: use "selection" slot to format the selected business -->
-          <!-- <template #selection="{ item }">
-            <div class="font-weight-bold text-truncate">{{ item.text }}</div>
-            <div class="text-subtitle-1">{{ item.subtext }}</div>
-          </template> -->
+          <template v-slot:activator="scope">
+            <div v-on="scope.on">
+              <v-select
+                id="search-type-options-select"
+                class="request-action-select"
+                filled
+                label="Select an Action"
+                :error-messages="getErrors.includes('request_action_cd') ? 'Please select an action' : ''"
+                :items="requestActions"
+                item-value="[group,value]"
+                :menu-props="{ bottom: true, offsetY: true, maxHeight: 423 }"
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.0.16",
```
severinbeauvais, line 3: The current version is actually 5.0.17. I don't know why GH isn't showing that here.

Ref: https://github.com/bcgov/namerequest/blob/feature-way-of-navigating/package.json

src/components/new-request/search.vue
```diff
@@ -9,48 +9,62 @@
     <v-row class="mt-6">
       <!-- Request Action -->
       <v-col cols="12" md="6" class="py-0">
```
severinbeauvais, line 11: The change below looks cleaner if you go to ⚙️ and then hide whitespace changes.

src/components/new-request/search.vue
```diff
@@ -9,48 +9,62 @@
...
-              :class="{ 'hide-me': item.disabled }"
-            >
-              <div class="font-weight-bold">{{ item.text }}</div>
-              <div>{{ item.subtext }}</div>
-            </v-list-item-content>
+          <template v-slot:activator="scope">
+            <div v-on="scope.on">
+              <v-select
+                id="search-type-options-select"
+                class="request-action-select"
```
severinbeauvais, line 22: I restored the id (from the original code) even though it's applied to the HTML input element, which is pretty useless for us right now.

The class is used as a selector to find the inner selection. I tried using a ref here instead -- which is equivalent to selecting this item -- but then crawling its children to find the inner selection was messy/fragile, whereas `querySelector()` is simple and probably more robust.

src/components/new-request/search.vue
```diff
@@ -9,48 +9,62 @@
...
+                  >
+                    <div class="font-weight-bold">{{ item.text }}</div>
+                    <div>{{ item.subtext }}</div>
+                  </v-list-item-content>
+                </template>
+              </v-select>
+            </div>
           </template>
-        </v-select>
+          <span>{{ request && request.text }}</span>
```
severinbeauvais, line 66: In the original code, this was `getRequestText` but that didn't return the correct value anyway, so I deleted it.

Note that this v-tooltip is conditional (line 16).

src/components/new-request/search.vue
```diff
@@ -189,6 +203,7 @@
     </v-row>
 
     <!-- Corporate Number checkbox, only for XPro Canadian locations -->
+    <!-- *** TODO: change this to "isCanadian" -->
```
severinbeauvais, line 206: I'll do this in my next PR (part of this ticket, anyway).

package.json
severinbeauvais: Obsolete issue.

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-647-hix22fnx.web.app

src/components/new-request/search.vue
```diff
@@ -611,6 +627,15 @@ export default class Search extends Mixins(CommonMixin) {
   onRequestActionChange (request: RequestActionsI): void {
     this.request = request
 
+    // calculate whether to show tooltip
+    // (in next tick after DOM update)
+    Vue.nextTick(() => {
+      const el = document.querySelector('.request-action-select .v-select__selection') as any
+      const offsetWidth = el?.offsetWidth as number
+      const scrollWidth = el?.scrollWidth as number
+      this.showRequestActionTooltip = (offsetWidth < scrollWidth)
```
JazzarKarim, line 636: This is a pretty cool solution.

JazzarKarim (approved):
LGTM! 👍

src/components/new-request/search.vue
severinbeauvais: It's also the _only_ solution, because the text clipping is done automatically by the browser.

PR #655, opened 2023-08-18
16636 Update Extrapro NR flow (continued)

severinbeauvais:
*Issue #:* bcgov/entity#16636

*Description of changes:*
    - app version = 5.0.25
    - moved Bullets Colin Link into Search
    - hide details in misc components
    - renamed getIsConversion -> isConversion
    - refactored layout in Search
    - added Federal bullets
    - moved isXXX computed to getters.ts
    - updated misc computed
    - title-cased entity names (per uxpin)
    - updated search test suite


src/components/common/bullets-colin-link.vue
severinbeauvais, line 1: This component, with slots for the previous code, was a good idea at the time, but with the new way of navigating it was becoming "messy" to have knowledge of the flows in the parent (Search) and here, so I moved what was in here back into the parent.

As we near the end of the flow changes, there may be an opportunity(ies) to break out some code into sub-components, but I see nothing obvious right now that wouldn't increase, instead of decrease, complexity.

One thing to watch for is to keep row/col layout in a single file. For example, I think we shouldn't have row declarations in Search and columns in sub-components. Instead, all rows/cols should be in Search, and the sub-components can be the column contents.

src/components/common/nested-select.vue
```diff
@@ -2,6 +2,7 @@
   <v-select
     class="request-action-select"
     filled
+    hide-details="auto"
```
severinbeauvais, line 5: This keeps the component "short" unless some detail messages (errors) need to be shown. This is to prevent excess bottom whitespace, which we sometimes absorb with negative bottom margins -- but this is better.

src/components/new-request/name-check/name-check.vue
```diff
@@ -18,7 +18,7 @@
       <v-col v-if="showDesignationSelect" :class="{'pl-3': !isMobile}" cols="12" lg="3">
         <v-select filled
                   :items="designationOptions"
-                  label="Enter designation"
+                  label="Select a Designation"
```
severinbeauvais, line 21: To be consistent with Search page (as per uxpin).

src/list-data/entity-type-data.ts
```diff
@@ -3,7 +3,7 @@ import { EntityI } from '@/interfaces/models'
 
 export const EntityTypesBcData: EntityI[] = [
   {
-    text: 'Sole proprietorship',
+    text: 'Sole Proprietorship',
```
severinbeauvais, line 6: The uxpin showed some entity types in title case, so I changed them all over.

src/components/new-request/search.vue
```diff
@@ -1,14 +1,12 @@
 <template>
-  <v-container fluid id="search-container" class="copy-normal pa-10">
-    <v-row no-gutters>
-      <v-col cols="12" class="pt-0 font-weight-bold h6">
+  <v-container fluid id="search-container" class="copy-normal pt-10 px-10 pb-12">
```
severinbeauvais, line 2: image

src/components/new-request/search.vue
```diff
@@ -1,14 +1,12 @@
...
+    <v-row>
+      <v-col cols="12" class="font-weight-bold h6">
```
severinbeauvais, line 4: image

src/components/new-request/search.vue
```diff
@@ -1,14 +1,12 @@
...
         Get started by selecting an action: [{{ getRequestActionCd }} / {{ getLocation }} / {{ entity_type_cd }}]
       </v-col>
-    </v-row>
 
-    <v-row class="mt-6">
```
severinbeauvais, line 9: I got rid of the "second row". Most sections here are simply columns, which automatically wrap or stack depending on screen width. Eg,

image

src/components/new-request/search.vue
```diff
@@ -96,92 +95,135 @@
...
 
-    <!-- Corporate Number checkbox, only for XPro Canadian locations -->
-    <v-row v-if="isCanadian" no-gutters>
-      <v-col class="d-flex justify-end">
-        <v-tooltip
-          top min-width="390"
-          content-class="top-tooltip"
-          transition="fade-transition"
-          :disabled="isMobile"
+    <!-- Type of Business, only when an entity type is selected (or federal) -->
```
severinbeauvais, line 105: Moved here from previous sub-component.

src/components/new-request/search.vue
```diff
@@ -315,40 +365,20 @@ export default class Search extends Mixins(CommonMixin) {
...
   get showEntityType (): boolean {
-    if (this.getLocation) return true
+    if (this.getLocation && !this.isFederal) return true
     return false
   }
 
   get showBusinessLookup () {
     if (this.isChangeName) return true
-    if (this.isAlteration) return true
+    if (this.isConversion) return true
```
severinbeauvais, line 381: A conversion is an alteration, but the NR Request Action Code is CNV so I renamed this getter accordingly.

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-655-fw1o16m2.web.app

JazzarKarim (approved):
LGTM! 👍

src/components/common/bullets-colin-link.vue
JazzarKarim: I see. Makes so much sense Sev. Having it with slots previously provided limitations on flexible we could've been with the flows. Cool!

kzdev420 (approved):
LGTM

ketaki-deodhar (approved):
much cleaner! LGTM 👍

PR #656, opened 2023-08-21
16636 Refactored name-input and misc state/actions/getters

severinbeauvais:
*Issue #:* bcgov/entity#16636

*Description of changes:*
    - app version = 5.0.26
    - fixed MRAS Search Info  layout
    - fixed Name Input component validity error
    - cleaned up Name Input (label + message)
    - moved debugging from Search to Landing
    - fixed Search layout (all inputs are now columns + whitespace + bold)
    - cleaned up is-mras-search flags
    - changed v-checkbox from value/input to v-model (to fix reactivity issue)
    - renamed getIsXproMras -> getIsXproFlow
    - added NS and ON to MRAS jurisdiction list
    - added some mutations (for actions cleanup)
    - removed MRAS search on last page (getShowCorpNum and fetchCorpNum)
    - fixed isMrasJurisdiction to not depend on NR data (which is reset in some flows)
    - misc cleanup

### DIFF
--- a/src/components/dialogs/mras-search-info.vue
+++ b/src/components/dialogs/mras-search-info.vue
@@ -1,31 +1,32 @@
 <template>
   <v-dialog v-model="showModal" id="mras-search-info-modal" width="40rem" persistent>
     <v-card class="pa-9">
-      <v-row no-gutters>
-        <v-col cols="11">
-          <v-card-title>
-            Corporate Number Not Found
-          </v-card-title>
-        </v-col>
-        <v-col cols="1">
-          <v-icon md color="primary" @click="clearAndClose()">mdi-close</v-icon>
-        </v-col>
-      </v-row>
+      <v-card-title class="d-flex justify-space-between">
+        Corporate Number Not Found
+        <div class="mt-3">
+          <v-btn icon large class="dialog-close" @click="clearAndClose()">
+            <v-icon>mdi-close</v-icon>
+          </v-btn>
+        </div>
+      </v-card-title>
+
       <v-card-text class="copy-normal pb-3">
         <p>Corporate Number assigned by Home Jurisdiction: <span class="font-weight-bold">{{ corpSearch }}</span><br>
           Home jurisdiction: <span class="font-weight-bold">{{ jurisdictionText }}</span></p>
         <p>{{ resultDesc  }}</p>
         <p>{{ resultAct }}</p>
+
+        <NameInput
+          id="name-input-component"
+          :is-mras-search="!isNameSearch"
+        />
       </v-card-text>
-      <NameInput
-        id="name-input-component"
-        class="mb-n7 pa-0"
-        :is-mras-search="!isNameSearch"
-      />
+
       <v-card-actions class="justify-center">
         <div class="mt-1 mb-1 text-center">
-          <v-btn class="search-btn"
-                 @click="isNameSearch ? handleSubmit() : clearAndClose()">{{ name ? 'Search' : 'Close' }}</v-btn>
+          <v-btn class="search-btn" @click="isNameSearch ? handleSubmit() : clearAndClose()">
+            {{ name ? 'Search' : 'Close' }}
+          </v-btn>
         </div>
       </v-card-actions>
     </v-card>
--- a/src/components/new-request/business-lookup.vue
+++ b/src/components/new-request/business-lookup.vue
@@ -10,12 +10,12 @@
       append-icon="mdi-magnify"
       autocomplete="chrome-off"
       autofocus
-      class="mt-5 mb-n2"
       filled
-      hint="For example: &quot;Joe's Plumbing Inc.&quot;, &quot;BC1234567&quot;, &quot;FM1234567&quot;"
+      hint="Search by name, incorporation or registration number of existing business"
       item-text="name"
       item-value="identifier"
-      label="Business Name or Incorporation/Registration Number"
+      label="Find an existing business"
+      hide-details="auto"
       no-filter
       persistent-hint
       return-object
--- a/src/components/new-request/name-input.vue
+++ b/src/components/new-request/name-input.vue
@@ -1,27 +1,22 @@
 <template>
-  <v-row no-gutters>
-    <v-col cols="12" class="pa-0">
-      <v-text-field
-        id="name-input-text-field"
-        ref="nameInput"
-        :error-messages="message"
-        autocomplete="chrome-off"
-        :filled="!isReadOnly"
-        :rules="(searchValue && isMrasSearch) ? mrasRules : additionalRules"
-        :label="label"
-        :class="{ 'read-only-mode': isReadOnly }"
-        :disabled="isReadOnly"
-        :hint="hint"
-        hide-details="auto"
-        persistent-hint
-        v-model="searchValue"
-        @input="clearErrors()"
-        @blur="handleBlur()"
-        @keydown.enter="handleSubmit($event)"
-      >
-      </v-text-field>
-    </v-col>
-  </v-row>
+  <v-text-field
+    id="name-input-text-field"
+    ref="nameInputRef"
+    :error-messages="message"
+    autocomplete="chrome-off"
+    :filled="!isReadOnly"
+    :rules="(searchValue && isMrasSearch) ? mrasRules : defaultRules"
+    :label="label"
+    :class="{ 'read-only-mode': isReadOnly }"
+    :disabled="isReadOnly"
+    :hint="hint"
+    hide-details="auto"
+    persistent-hint
+    v-model="searchValue"
+    @input="setClearErrors()"
+    @blur="handleBlur()"
+    @keydown.enter="handleSubmit($event)"
+  />
 </template>
 
 <script lang="ts">
@@ -34,37 +29,42 @@ import { MRAS_MAX_LENGTH } from '@/components/new-request/constants'
 
 @Component({})
 export default class NameInput extends Vue {
-  // Refs
-  $refs!: {
-    nameInput: any
-  }
+  /** Whether to perform MRAS search. */
+  @Prop({ default: false }) readonly isMrasSearch!: boolean
+  /** Whether this component is read-only (eg, on name check page). */
+  @Prop({ default: false }) readonly isReadOnly!: boolean
+  /** Hint to show (eg, on name check page). */
+  @Prop({ default: null }) readonly hint!: string
 
-  // Global getters
+  // Store getters
   @Getter getCorpSearch!: string
   @Getter getErrors!: string[]
+  @Getter getHasNoCorpNum!: boolean
+  @Getter getIsXproFlow!: boolean
   @Getter getLocation!: Location
   @Getter getName!: string
   @Getter getRequestActionCd!: NrRequestActionCodes
-  @Getter getIsXproMras!: boolean
+  @Getter isMrasJurisdiction!: boolean
 
-  // Global actions
-  @Action setClearErrors!: ActionBindingIF
+  // Store actions
+  @Action setClearErrors!: () => void
   @Action setCorpSearch!: ActionBindingIF
   @Action setName!: ActionBindingIF
   @Action setMrasSearchInfoModalVisible!: ActionBindingIF
   @Action startAnalyzeName!: ActionBindingIF
 
-  // Props
-  @Prop({ default: false }) readonly isSearchAgain!: boolean
-  @Prop({ default: false }) readonly isMrasSearch!: boolean
-  @Prop({ default: false }) readonly isReadOnly!: boolean
-  @Prop({ default: null }) readonly hint!: string
+  readonly err_msg = 'Cannot exceed ' + MRAS_MAX_LENGTH + ' characters'
 
-  private err_msg = 'Cannot exceed ' + MRAS_MAX_LENGTH + ' characters'
-
-  additionalRules = [
+  readonly defaultRules = [
     v => (!v || v.length <= MRAS_MAX_LENGTH) || this.err_msg
   ]
+
+  nameInputComponent = null
+
+  mounted (): void {
+    // ref is only valid after component is mounted
+    this.nameInputComponent = this.$refs['nameInputRef']
+  }
 
   /** The array of validation rules for the MRAS corp num. */
   get mrasRules (): Function[] {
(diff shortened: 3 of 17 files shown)


pwei1018:
Temporary Url for review: https://namerequest-dev--pr-656-sagrrcnq.web.app

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-656-sagrrcnq.web.app

src/components/dialogs/mras-search-info.vue
```diff
@@ -1,31 +1,32 @@
...
-        </v-col>
-      </v-row>
+      <v-card-title class="d-flex justify-space-between">
+        Corporate Number Not Found
+        <div class="mt-3">
+          <v-btn icon large class="dialog-close" @click="clearAndClose()">
+            <v-icon>mdi-close</v-icon>
+          </v-btn>
+        </div>
+      </v-card-title>
```
severinbeauvais, line 11: Now using flexbox instead of v-row/v-col for layout. Looks like this:

image

pwei1018:
Temporary Url for review: https://namerequest-dev--pr-656-sagrrcnq.web.app

src/components/new-request/business-lookup.vue
```diff
@@ -10,12 +10,12 @@
...
       autofocus
-      class="mt-5 mb-n2"
       filled
-      hint="For example: &quot;Joe's Plumbing Inc.&quot;, &quot;BC1234567&quot;, &quot;FM1234567&quot;"
+      hint="Search by name, incorporation or registration number of existing business"
       item-text="name"
       item-value="identifier"
-      label="Business Name or Incorporation/Registration Number"
+      label="Find an existing business"
+      hide-details="auto"
```
severinbeauvais, line 18: As per uxpin:

image

PS The `hide-details` prop reduces the height if there is no message. I did this elsewhere, too.

src/components/new-request/name-input.vue
```diff
@@ -34,38 +29,43 @@ import { MRAS_MAX_LENGTH } from '@/components/new-request/constants'
...
-  additionalRules = [
+  readonly defaultRules = [
     v => (!v || v.length <= MRAS_MAX_LENGTH) || this.err_msg
   ]
 
+  nameInputComponent = null
+
+  mounted (): void {
+    // ref is only valid after component is mounted
+    this.nameInputComponent = this.$refs['nameInputRef']
```
severinbeauvais, line 66: I needed to do this or the computed `isCorpValid` threw an error before this component was mounted and was not reactive after. I believe this was an existing issue.

src/components/new-request/name-input.vue
```diff
@@ -76,40 +76,39 @@ export default class NameInput extends Vue {
...
+      }
     }
 
     if (this.getErrors.includes('length')) {
       return ['Please enter a longer name']
     }
 
-    if (this.getErrors.includes('name')) {
-      return ['Please enter a name to search for']
-    }
```
severinbeauvais, line 111: This case is handled in new code on line 104.

src/components/new-request/name-input.vue
```diff
@@ -138,19 +133,18 @@ export default class NameInput extends Vue {
...
+        if (this.getIsXproFlow) this.$root.$emit('showSpinner', true)
+        if (this.searchValue) await this.startAnalyzeName(null)
+        if (this.getIsXproFlow) this.$root.$emit('showSpinner', false)
+      }
       return
     }
     return event
   }
 
-  async analyzeName () {
```
severinbeauvais, line 147: There was no advantage to having this as its own function so I moved the code above.

src/interfaces/new-request-interface.ts
severinbeauvais, line 1: I just sorted the properties here.

src/components/dialogs/mras-search-info.vue
JazzarKarim: What's the benefit here Sev? Doesn't it look the same as before?

src/components/new-request/search.vue
```diff
@@ -2,7 +2,7 @@
   <v-container fluid id="search-container" class="copy-normal pt-10 px-10 pb-12">
     <v-row>
       <v-col cols="12" class="font-weight-bold h6">
-        Get started by selecting an action: [{{ getRequestActionCd }} / {{ getLocation }} / {{ entity_type_cd }}]
```
severinbeauvais, line 5: Move to Landing.vue, ie:

image

src/components/new-request/search.vue
```diff
@@ -45,7 +45,7 @@
       </v-col>
 
       <!-- Entity Type -->
-      <v-col v-if="showEntityType" cols="12" md="6">
+      <v-col v-if="showEntityType" cols="12" :md="getIsXproFlow ? 4 : 6">
```
severinbeauvais, line 48: Examples:

image

image

src/components/new-request/search.vue
```diff
@@ -97,133 +97,141 @@
       </v-col>
 
       <!-- Business Lookup -->
-      <v-col v-if="showBusinessLookup" cols="12">
-        <BusinessLookup />
-      </v-col>
-    </v-row>
```
severinbeauvais, line 103: Basic layout cleanup below. Basically, all input fields are now columns inside a single row, and they wrap/stack as needed.

src/components/new-request/search.vue
```diff
@@ -97,133 +97,141 @@
...
+              <li>You may register your extraprovincial business immediately using its existing name
+                at Corporate Online.</li>
+            </ul>
+          </v-col>
+
+          <!-- XPRO/MRAS number/name search/input -->
+          <v-col v-if="!isFederal" cols="12" :md="(getIsXproFlow || showDesignation) ? 8 : 12">
+            <NameInput
+              id="name-input-component"
+              :is-mras-search="(getIsXproFlow && isMrasJurisdiction && !noCorpNum)"
```
severinbeauvais, line 145: This is a fix -- no need to do MRAS search if jurisdiction isn't MRAS.

src/components/new-request/search.vue
```diff
@@ -97,133 +97,141 @@
...
+            v-if="showColinButton"
+            class="px-9"
+            id="go-to-colin-button"
+            :href="colinLink"
+            target="_blank"
           >
-            <template v-slot:activator="{ on }">
-              <v-checkbox
-                :value="getHasNoCorpNum"
-                @input="setNoCorpNum($event)"
```
severinbeauvais, line 172: For some unexplained reason, the "input" event lagged by a tick, so basically it wasn't set when we needed it. Changing this to simply using "v-model" works fine and is cleaner.

src/components/new-request/search.vue
```diff
@@ -451,11 +462,16 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
-      return this.getConversionTypeOptions
-    }
-    return this.getEntityTypeOptions
+  get noCorpNum (): boolean {
+    return this.getHasNoCorpNum
+  }
+
+  set noCorpNum (value: boolean) {
+    this.setNoCorpNum(value)
+  }
```
severinbeauvais, line 471: Get and Set needed for v-model use.

src/components/new-request/search.vue
```diff
@@ -482,9 +498,10 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
 
   get showContinueInButton (): boolean {
     // for now, return True because Continuation In filings are not yet implemented
-    return true
+    return this.isContinuationIn
```
severinbeauvais, line 501: Another small bugfix: this always showed the COLIN button.

src/components/new-request/search.vue
```diff
@@ -578,49 +595,47 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
     this.request = request
 
-    // calculate whether to show tooltip
-    // (in next tick after DOM update)
-    Vue.nextTick(() => {
-      const el = document.querySelector('.request-action-select .v-select__selection') as any
-      const offsetWidth = el?.offsetWidth as number
-      const scrollWidth = el?.scrollWidth as number
-      this.showRequestActionTooltip = (offsetWidth < scrollWidth)
-    })
```
severinbeauvais, line 591: Don't worry. I move this below, and I went with all-async in this method instead of using a callback in the nextTick function.

src/components/new-request/search.vue
```diff
@@ -681,8 +698,8 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
 
 // Line spacing between bullet points and sizing
 .bullet-points {
-  font-size: $px-14;
   line-height: 1.5rem;
+  font-size: $px-15;
```
severinbeauvais, line 702: Uxpin:

image

Actual:

image

src/components/new-request/search.vue
```diff
@@ -97,133 +97,141 @@
...
-          />
-          <v-radio
-            id="numbered-company-radio"
-            label="Numbered Company"
-            :value="CompanyType.NUMBERED_COMPANY"
-          />
-        </v-radio-group>
+      <v-col v-if="showBusinessLookup" cols="12" md="6">
+        <BusinessLookup v-if="!business" @business="business=$event"/>
+        <v-text-field v-else clearable disabled filled hide-details :value="business.name" />
```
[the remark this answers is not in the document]
JazzarKarim: Ah so right now it's only a placeholder (always disabled) right?

src/list-data/entity-type-data.ts
```diff
@@ -31,6 +31,7 @@ export const EntityTypesBcData: EntityI[] = [
   },
   {
     text: 'Limited Company',
+    value: EntityType.CR,
```
severinbeauvais, line 34: In this file, I moved the value property to the top of each record to be consistent and to more easily see what type each record is.

src/store/actions.ts
```diff
@@ -297,11 +297,12 @@ export const setAddressSuggestions: ActionIF = ({ commit }, addressSuggestions:
...
 export const fetchCorpNum = async ({ getters }, corpNum: string): Promise<any> => {
-  // NB: MRAS search should have been done on first page for xpro
-  // NB: COLIN search should have been done on first page for BC
-  if (getters.getShowCorpNum === CorpNumRequests.MRAS) {
-    return checkMRAS({ getters }, corpNum)
-  }
+  // NB: MRAS search was done on first page
+  // if (getters.getShowCorpNum === CorpNumRequests.MRAS) {
+  //   return checkMRAS({ getters }, corpNum)
+  // }
```
severinbeauvais, line 303: Should be deleted but I think there's another flow that does a MRAS search so I want to wait to see if it needs this.

src/store/actions.ts
```diff
@@ -1107,10 +1114,11 @@ export const startAnalyzeName: ActionIF = async ({ commit, getters }) => {
   commit('mutateNameOriginal', name) // Set original name for reset baseline
 
   // xpro get name call
-  if (getters.getIsXproMras) {
-    commit('mutateNRData', { key: 'xproJurisdiction', value: getters.getJurisdictionText })
-    commit('mutateNRData', { key: 'homeJurisNum', value: getters.getCorpSearch })
-    if (!getters.getHasNoCorpNum) {
+  if (getters.getIsXproFlow && !getters.getHasNoCorpNum) {
+    commit('mutateXproJurisdiction', getters.getJurisdictionText)
+    commit('mutateHomeJurisNum', getters.getCorpSearch)
```
severinbeauvais, line 1119: Now callling mutations correctly instead of abstract "mutateNRData".

src/store/getters.ts
```diff
@@ -182,6 +199,14 @@ export const isRestoration = (state: StateIF): boolean => {
   return (getRequestActionCd(state) === NrRequestActionCodes.RESTORE)
 }
 
+/** True if current flow is XPRO (includes AMALGAMATION, NEW_BUSINESS, and possibly others). */
+export const getIsXproFlow = (state: StateIF): boolean => {
+  return (
+    (getRequestActionCd(state) !== NrRequestActionCodes.MOVE) &&
+    [Location.CA, Location.IN].includes(getLocation(state))
+  )
+}
```
severinbeauvais, line 208: Note that "MOVE" from Canada or International doesn't seem to count as a XPRO flow 🤷‍♂️

src/store/getters.ts
```diff
@@ -711,20 +733,21 @@ export const getShowCorpNum = (state: StateIF): CorpNumRequests.COLIN | CorpNumR
...
+  //   ) {
+  //     return CorpNumRequests.MRAS
+  //   }
+
+  //   if (getLocation(state) === Location.BC &&
+  //     [NrRequestActionCodes.MOVE].includes(getRequestActionCd(state))
+  //   ) {
+  //     return CorpNumRequests.MRAS
+  //   }
+  // }
```
severinbeauvais, line 750: As per comment above, I think we no longer need this but I'm not certain.

src/components/new-request/search.vue
```diff
@@ -578,49 +595,47 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   }
 
   /** Called when Request Action menu item is changed. */
-  onRequestActionChange (request: RequestActionsI): void {
+  async onRequestActionChange (request: RequestActionsI): Promise<void> {
```
JazzarKarim, line 598: Does it have to return `Promise<void>`? Wouldn't be sufficient to be just `void` Sev?

src/store/getters.ts
```diff
@@ -1233,6 +1256,6 @@ export const isMrasRequestType = (state: StateIF): boolean => {
 
 /** True if jurisdiction is a MRAS jurisdiction. */
 export const isMrasJurisdiction = (state: StateIF): boolean => {
-  const { xproJurisdiction } = getNrData(state)
+  const xproJurisdiction = getJurisdictionText(state)
```
severinbeauvais, line 1259: The old code depended on `startAnalyzeName` action being called, but:
- we need to identify if this is a MRAS jurisdiction before that
- we really don't need to look at the NR data for this (this is the object that will be posted to the API)
- we can just get the xpro jurisdiction text from another getter which, itself, doesn't depend on anything we don't already have

src/store/mutations.ts
```diff
@@ -250,6 +250,14 @@ export const mutateNRData = (state: StateIF, { key, value }) => {
   state.stateModel.newRequestModel.nrData[key] = value
 }
 
+export const mutateXproJurisdiction = (state: StateIF, val: string) => {
+  state.stateModel.newRequestModel.nrData.xproJurisdiction = val
+}
+
+export const mutateHomeJurisNum = (state: StateIF, val: string) => {
+  state.stateModel.newRequestModel.nrData.homeJurisNum = val
+}
```
severinbeauvais, line 259: I added these to get was from `mutateNRData` as that was really abstract.

src/components/dialogs/mras-search-info.vue
severinbeauvais: Before:

image

In the new code, the X is better lined up. Overall, flexbox for this purpose (and correctly using dialog title, text, actions) makes for cleaner template code.

JazzarKarim: Oh, OK. Got it.

src/components/new-request/search.vue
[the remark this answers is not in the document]
severinbeauvais: Yeah. It'll look better for demos until we actually use the results (business id and entity type) in misc flows.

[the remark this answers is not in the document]
JazzarKarim: Makes sense. Deal.

src/components/new-request/search.vue
severinbeauvais: Nope, because it's async, so technically it returns a promise.

Remember that await is equivalent to ".then().catch()", if that helps you figure out promises.

JazzarKarim (approved):
LGTM! 👍 Lots of discoveries/learning for me here.

src/components/new-request/search.vue
severinbeauvais: We could probably go to 16px for the bullet font size, but UXA didn't say anything last time so let's see what they think here.

chenhongjing (approved):
LGTM 🐱 some parts are also very helpful for other flow updates

ketaki-deodhar (approved):
LGTM!

PR #663, opened 2023-08-28
17489 Incremental business fetch + bit of cleanup

severinbeauvais:
*Issue #:* bcgov/entity#17489

*Description of changes:*
    - changed Legal Api to V2
    - app version = 5.0.30
    - added max-height support to Nested Select component
    - added call to fetch corp num
    - updated validation
    - added error handling
    - increased max height of request action menu to partially show last option (UX feedback)
    - simplified business lookup/search template code
    - added business event handler + misc fallbacks
    - deleted unused business interfaces + added interface for business fetch
    - commented out interceptor debugging message
    - moved searchEntities and searchColin code to namex-services.ts
    - fixed a bunch of function signatures in actions.ts
    - refactored checkCOLIN() for simplicity and better return object
    - fixed a bug in submit()


bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

.env.example
```diff
@@ -15,7 +15,7 @@ VUE_APP_SOCIETIES_ONLINE_HOME_URL="https://dev.bcregistry.ca/societies/"
 VUE_APP_AUTH_API_URL="https://auth-api-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_AUTH_API_VERSION="/api/v1"
 VUE_APP_LEGAL_API_URL="https://legal-api-dev.apps.silver.devops.gov.bc.ca"
-VUE_APP_LEGAL_API_VERSION="/api/v1"
+VUE_APP_LEGAL_API_VERSION="/api/v2"
```
severinbeauvais, line 18: The Legal API "fetch business" call doesn't work on V1, but I need to test the "create business" call on V2.

src/components/new-request/search.vue
```diff
@@ -363,7 +362,7 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   readonly EntityType = EntityType
   activeActionGroup = NaN
   showRequestActionTooltip = false
-  business = null as BusinessLookupResultIF
+  business = null
```
severinbeauvais, line 365: I'll probably create a type (interface) for this, containing the minimum fields.

src/components/new-request/search.vue
```diff
@@ -465,7 +464,7 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     if (type && this.isConversion) {
       // convert NrRequestTypeCodes -> EntityType
       const value = type as unknown as NrRequestTypeCodes
-      let { entity_type_cd } = ConversionTypes.find(conv => conv.value === value)
+      let { entity_type_cd } = ConversionTypes.find(conv => conv.value === value) || { entity_type_cd: null }
```
severinbeauvais, line 467: Added fallback to pre-empt errors when Find returns nothing.

src/components/new-request/search.vue
```diff
@@ -590,7 +595,7 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     this.setDesignation('')
     // clear "Select a Business Type" field when "View all business types" or Society is selected
     if (!this.entity_type_cd || this.entity_type_cd === EntityType.INFO) {
-      this.$refs.selectBusinessTypeRef.reset()
+      this.$refs.selectBusinessTypeRef && this.$refs.selectBusinessTypeRef.reset()
```
severinbeauvais, line 598: Added this to pre-empt errors when selectBusinessTypeRef isn't defined (ie, component isn't rendered at the moment due to v-if).

src/list-data/request-action-mapping.ts
```diff
@@ -76,12 +76,12 @@ export const XproMapping: MappingI = {
...
-  NrRequestActionCodes.AMALGAMATE,
   NrRequestActionCodes.CHANGE_NAME,
   NrRequestActionCodes.CONVERSION,
   NrRequestActionCodes.RESTORE
 ]
 
+// *** TODO: should these be from FF instead?
 export const ColinRequestTypes = [
   EntityType.BC,
   EntityType.CC,
```
severinbeauvais, line 87: This is used in companyRadioBtnApplicable() and I think we should use a FF instead.

src/services/business-services.ts
```diff
@@ -8,13 +8,18 @@ export default class BusinessServices {
...
   }
 
   /** Fetches a business record. */
-  static async fetchBusiness (businessId: number): Promise<any> {
-    return axios.get(`${BusinessServices.legalApiUrl}/businesses/${businessId}`)
+  // *** TODO: this only works with V2 API -- need to update 1Password
+  static async fetchBusiness (identifier: string): Promise<any> {
+    const url = `${BusinessServices.legalApiUrl}/businesses/${identifier}`
+    return axios.get(url)
+      .then(response => response.data?.business)
```
severinbeauvais, line 23: I added this to return the actual business object.

src/services/namex-services.ts
```diff
@@ -26,7 +26,7 @@ axiosNamex.interceptors.request.use(
     config.headers.common['BCREG-NRL'] = sessionStorage.getItem('BCREG-NRL')
     config.headers.common['BCREG-User-Phone'] = sessionStorage.getItem('BCREG-phoneNumber')
     config.headers.common['BCREG-User-Email'] = sessionStorage.getItem('BCREG-emailAddress')
-    console.log('in interceptor, headers: ', config?.headers?.common) // eslint-disable-line
+    // console.log('in interceptor, headers: ', config?.headers?.common) // eslint-disable-line
```
severinbeauvais, line 29: I don't think we've used this for debugging in a loooooong time.

src/services/namex-services.ts
```diff
@@ -122,6 +122,20 @@ export default class NamexServices {
     }
   }
 
+  /** Searches LEAR (Legal API) for business record for specified corp num. */
+  static async searchLear (corpNum: string): Promise<any> {
+    const url = `${appBaseURL}/businesses/${corpNum}`
+    return this.axios.get(url).then(response => response.data?.business)
+  }
```
severinbeauvais, line 129: This is a proxied call (Namerequest UI <--> Namex API <--> Legal API) and it returns the same object as `BusinessServices.fetchBusiness()`.

src/services/namex-services.ts
```diff
@@ -122,6 +122,20 @@ export default class NamexServices {
...
+
+  /** Searches COLIN for business record for specified corp num. */
+  static async searchColin (corpNum: string): Promise<any> {
+    // remove BC prefix as COLIN only supports base number with no prefix for BC's
+    const cleanedCorpNum = corpNum.replace(/^BC+/i, '')
+    const url = `${appBaseURL}/colin/${cleanedCorpNum}`
+    return this.axios.post(url).then(response => response.data)
+  }
```
severinbeauvais, line 137: I moved this code here (was in actions before).

.env.example
severinbeauvais: Tested! `BusinessServices.createBusiness()` works fine on V2 endpoint.

severinbeauvais: I added a "todo" item to my ticket to remember to update 1Password.

src/store/actions.ts
```diff
@@ -80,7 +78,7 @@ export const setActiveComponent: any = ({ commit }, component): void => {
  * @param action the action to confirm
  * @returns True if confirmed, otherwise False
  */
-export const confirmAction: ActionIF = async ({ commit, getters }, action: string): Promise<boolean> => {
```
severinbeauvais, line 83: `ActionIF` is wrong (even though the syntax allows it). The actual return type is `Promise<boolean>`.

Ditto the other changes below. It's just cleanup -- no functional impact.

src/store/actions.ts
```diff
@@ -312,37 +310,44 @@ export const fetchCorpNum = async ({ getters }, corpNum: string): Promise<any> =
...
-          return Promise.reject(error)
-        }
-      }).catch(error => {
-        return Promise.reject(error)
-      })
-    } else {
-      return Promise.reject(error)
-    }
-  })
+async function checkCOLIN ({ getters }, corpNum: string): Promise<any> {
```
severinbeauvais, line 313: I changed this to "await" format as it's cleaner (IMO).

src/store/actions.ts
```diff
@@ -463,7 +468,7 @@ export const submit: any = async ({ commit, getters, dispatch }): Promise<any> =
       if (!getters.isEditMode && [NrState.COND_RESERVED, NrState.RESERVED].includes(getters.getNrState)) {
         request = await NamexServices.getNameRequest(true)
         if (request?.stateCd === NrState.CANCELLED) {
-          await setActiveComponent('Timeout')
+          setActiveComponent({ commit }, 'Timeout')
```
severinbeauvais, line 471: Bugfix -- found when I fixed the signature of setActiveComponent() above!

src/components/new-request/search.vue
```diff
@@ -97,23 +98,22 @@
...
         </template>
-        <div v-else class="d-flex justify-space-between align-center">
-          <v-text-field
-            append-outer-icon="mdi-close"
-            disabled
-            filled
-            hide-details
-            :value="business.name"
-          />
-          <div @click="business=null">
```
severinbeauvais, line 113: Don't need separate clickable icon -- now use readonly text field (instead of disabled) with close icon.

src/components/new-request/search.vue
```diff
@@ -560,6 +560,12 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     return this.getEntityTextFromValue || 'specified business type'
   }
 
+  /** Event handled for business lookup/fetch. */
+  onBusiness (business: BusinessFetchIF): void {
+    this.business = business
+    this.entity_type_cd = this.business?.legalType || null
```
severinbeauvais, line 566: In case user clears business input.

src/components/new-request/search.vue
severinbeauvais: Done.

src/store/actions.ts
```diff
@@ -312,37 +312,47 @@ export const fetchCorpNum = async ({ getters }, corpNum: string): Promise<any> =
...
-  return axios.get(url, {}).catch(error => {
-    if (error.response && error.response.status === 404) {
-      // Remove BC prefix as Colin only supports base number with no prefix for BC's
-      const cleanedCorpNum = corpNum.replace(/^BC+/i, '')
-      url = `${appBaseURL}/colin/${cleanedCorpNum}`
-      return axios.post(url, {}).then(response => {
-        if (response.data.directors === 'Not Available' && response.data.incorporated === 'Not Available') {
-          const error = new Error('Not Found')
-          return Promise.reject(error)
+async function checkCOLIN ({ getters }, corpNum: string): Promise<BusinessFetchIF> {
```
severinbeauvais, line 315: Note that this function is now in "await" format (instead of "then" format), which I think is cleaner (and is also more modern).

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

severinbeauvais:
I'm still working on validation of the entered business id (in business-fetch.vue) but otherwise this is usable right now.

I will also have a future PR (in a few days) to use the new endpoint to fetch COLIN data.

Note that a business that only exists in COLIN currently returns partially-mocked data (hard-coded legal name, type and state). Your options are to just use that data as-is or pick a business that exists in LEAR.

src/interfaces/business.ts
```diff
@@ -1,27 +1,28 @@
-export interface Business {
-  natureOfBusiness: string
-}
-
-export interface NameRequest {
```
severinbeauvais, line 5: This interface and the one above weren't used anywhere else so there was no need to export them.

I could have left them un-exported for use in BusinessRequest below, but I didn't like them appearing to represent a Business (or NameRequest) but not containing all the necessary fields, so I just moved them into BusinessRequest below (which I think is more clear).

src/store/actions.ts
```diff
@@ -312,37 +312,47 @@ export const fetchCorpNum = async ({ getters }, corpNum: string): Promise<any> =
...
+  try {
+    // first try to find business in Entities (Legal API)
+    const data = await NamexServices.searchEntities(corpNum)
+    return {
+      identifier: data.identifier,
+      legalName: data.legalName,
+      legalType: data.legalType,
+      state: data.state
+    }
```
severinbeauvais, line 324: I previously wrote this in a different syntax but I prefer the "simple" syntax above.

Here's the alternate syntax. Do you understand it?

```
const { identifier, legalName, legalType, state } = await NamexServices.searchEntities(corpNum)
return { identifier, legalName, legalType, state }
```

src/components/new-request/business-fetch.vue
```diff
@@ -6,28 +6,26 @@
...
-      :name="Math.random()"
       hide-details="auto"
       hint="Enter registration number of existing business"
       label="Fetch an existing business"
       persistent-hint
-      ref="searchField"
-      :rules="rules"
       v-model="searchField"
+      @click:append="onItemSelected()"
+      @keydown.tab.exact="onItemSelected()"
```
severinbeauvais, line 15: This excludes Shift-Tab (so it searches when tabbing forward but not tabbing backward.

src/components/new-request/business-fetch.vue
```diff
@@ -40,45 +38,54 @@ enum States {
...
+  private validate (): boolean {
+    if (!this.searchField) {
+      this.errorMessages = ['Required field']
+    } else if (!/^(A|BC|C|CP|FM)\d{7}$/.test(this.searchField)) {
+      this.errorMessages = ['Please enter a valid registration number']
+    } else {
+      this.errorMessages = []
+    }
+    return (this.errorMessages.length === 0)
+  }
```
severinbeauvais, line 66: I couldn't get rules-based validation to not validate while the user was typing (even when I wrapped the v-text-field in a v-form with lazy-validation) so I did this instead.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

JazzarKarim (approved):
LGTM. Shouldn't be too bad to merge. There's some overlap in `search.vue` but other than that should be OK.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

src/components/new-request/business-fetch.vue
```diff
@@ -6,28 +6,26 @@
...
-      :loading="state === States.SEARCHING"
-      :name="Math.random()"
       hide-details="auto"
       hint="Enter registration number of existing business"
       label="Fetch an existing business"
       persistent-hint
-      ref="searchField"
-      :rules="rules"
-      v-model="searchField"
+      v-model.trim="searchField"
```
severinbeauvais, line 13: "Trim" removes leading and trailing whitespace.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

src/components/new-request/business-fetch.vue
```diff
@@ -6,28 +6,27 @@
...
+      @click:append="onItemSelected()"
+      @input="errorMessages = []"
```
severinbeauvais, line 15: Clear error if user changes input.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-663-tjsvuma7.web.app

src/list-data/request-action-mapping.ts
severinbeauvais: Well, someone working on Change Name / Restoration / Alteration will need to handle this (ie, for which entity types do we do a business lookup?).

cc: @JazzarKarim @chenhongjing @ketaki-deodhar

severinbeauvais: Or maybe this isn't needed at all to decide whether to look up a business?

However, the store getter that uses this is used to determine whether to show numbered/named options...

ketaki-deodhar (approved):
LGTM!

chenhongjing (approved):
LGTM 😸

src/store/actions.ts
severinbeauvais: No responses on this? 😃

src/list-data/request-action-mapping.ts
severinbeauvais: How about if I tackle this?

JazzarKarim: Sev, here's how I'm doing it right now. I created a new getter called `showRestoreNowButton`. This getter returns a boolean whether we can restore by using the FF.

I'm using this getter next to `showColinButton`. So, `showColinButton || showRestoreNowButton`.

I didn't need to touch the ColinRequestTypes as of yet. 

I was thinking for the other types; Alteration for example, we would do the same.

src/store/actions.ts
JazzarKarim: Yes! Much easier to read now 🙌

src/list-data/request-action-mapping.ts
severinbeauvais: OK but this is different -- it's whether to show the named/numbered radio buttons or not.

ketaki-deodhar: Using FF might be a good idea here.

JazzarKarim: Wait, sorry, you're right.

I added an `|| business` to the template encapsulating the `companyRadioBtnApplicable`. I created a watcher to watch when the business object changes and I'm doing the work there in an ifRestoration block.

However, you're right. This list might have to change based on which flow we're in? Currently it works for restorations since this is what I need but it might be different for other flows.

severinbeauvais: I'm not doing any more work on this. I think the existing code is OK (but #664 makes is clearer).

JazzarKarim: Sounds good 👍

PR #666, opened 2023-08-30
Implemented the Restoration/Reinstatement flow

JazzarKarim:
*Issue #:* /bcgov/entity#16640

*Description of changes:*
- Implemented the new Restoration/Reinstatement flow

### DIFF
--- a/src/components/new-request/business-lookup.vue
+++ b/src/components/new-request/business-lookup.vue
@@ -14,7 +14,7 @@
       hint="Search by name, incorporation or registration number of existing business"
       item-text="name"
       item-value="identifier"
-      label="Find an existing business"
+      :label="lookupLabel"
       hide-details="auto"
       no-filter
       persistent-hint
@@ -34,7 +34,7 @@
 
       <template v-slot:no-data>
         <p class="pl-5 font-weight-bold">
-          No active B.C. business found
+          {{  lookupNoActiveText }}
         </p>
         <p class="pl-5">
           Ensure you have entered the correct business name or number.
@@ -46,7 +46,7 @@
 
 <script lang="ts">
 import Vue from 'vue'
-import { Component, Emit, Watch } from 'vue-property-decorator'
+import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
 import { debounce } from 'lodash'
 import { BusinessLookupResultIF, BusinessSearchIF } from '@/interfaces'
 import BusinessLookupServices from '@/services/business-lookup-services'
@@ -65,6 +65,9 @@ enum States {
  */
 @Component({})
 export default class BusinessLookup extends Vue {
+  // Status of businesses to search for prop
+  @Prop({ default: 'ACTIVE' }) readonly businessStatus!: string
+
   // enum for template
   readonly States = States
 
@@ -77,17 +80,43 @@ export default class BusinessLookup extends Vue {
   /** State of this component. */
   state = States.INITIAL
 
+  /** Business Lookup labels and warnings. */
+  lookupLabel = ''
+  lookupNoActiveText = ''
+
+  /** Set business lookup text on mount. */
+  mounted (): void {
+    this.setLookupText()
+  }
+
   /** Called when searchField property has changed. */
   @Watch('searchField')
   onSearchFieldChanged (): void {
     this.onSearchInputDebounced(this)
   }
 
+  /** Called when business lookup search status prop has changed. */
+  @Watch('businessStatus')
+  onStatuschanged (): void {
+    this.setLookupText()
+  }
+
+  /** Business lookup text based on the business status to search for. */
+  private setLookupText (): void {
+    if (this.businessStatus === 'HISTORICAL') {
+      this.lookupLabel = 'Find a historical business'
+      this.lookupNoActiveText = 'No historical business found'
+    } else {
+      this.lookupLabel = 'Find an existing business'
+      this.lookupNoActiveText = 'No active B.C. business found'
+    }
+  }
+
   private onSearchInputDebounced = debounce(async (that: this) => {
     // safety check
     if (that.searchField && that.searchField.length > 2) {
       that.state = States.SEARCHING
-      const searchStatus = 'ACTIVE' // search for ACTIVE
+      const searchStatus = this.businessStatus // search for status based on prop
       that.searchResults = await BusinessLookupServices.search(that.searchField, searchStatus).catch(() => [])
 
       // display appropriate section
--- a/src/components/new-request/business-lookup.vue
+++ b/src/components/new-request/business-lookup.vue
@@ -34,7 +34,7 @@
 
       <template v-slot:no-data>
         <p class="pl-5 font-weight-bold">
-          {{  lookupNoActiveText  }}
+          {{ lookupNoActiveText }}
         </p>
         <p class="pl-5">
           Ensure you have entered the correct business name or number.
--- a/src/components/new-request/business-lookup.vue
+++ b/src/components/new-request/business-lookup.vue
@@ -50,7 +50,7 @@ import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
 import { debounce } from 'lodash'
 import { BusinessLookupResultIF, BusinessSearchIF } from '@/interfaces'
 import BusinessLookupServices from '@/services/business-lookup-services'
-import { EntityType } from '@/enums'
+import { EntityStates, EntityType } from '@/enums'
 
 enum States {
   INITIAL = 'initial',
@@ -66,7 +66,7 @@ enum States {
 @Component({})
 export default class BusinessLookup extends Vue {
   // Status of businesses to search for prop
-  @Prop({ default: 'ACTIVE' }) readonly businessStatus!: string
+  @Prop({ default: EntityStates.ACTIVE }) readonly searchStatus!: string
 
   // enum for template
   readonly States = States
@@ -84,26 +84,19 @@ export default class BusinessLookup extends Vue {
   lookupLabel = ''
   lookupNoActiveText = ''
 
-  /** Set business lookup text on mount. */
-  mounted (): void {
-    this.setLookupText()
-  }
-
   /** Called when searchField property has changed. */
   @Watch('searchField')
   onSearchFieldChanged (): void {
     this.onSearchInputDebounced(this)
   }
 
-  /** Called when business lookup search status prop has changed. */
-  @Watch('businessStatus')
+  /**
+   * Called when business lookup search status prop has changed.
+   * Business lookup text based on the business status to search for.
+   */
+  @Watch('businessStatus', { immediate: true })
   onStatuschanged (): void {
-    this.setLookupText()
-  }
-
-  /** Business lookup text based on the business status to search for. */
-  private setLookupText (): void {
-    if (this.businessStatus === 'HISTORICAL') {
+    if (this.searchStatus === EntityStates.HISTORICAL) {
       this.lookupLabel = 'Find a historical business'
       this.lookupNoActiveText = 'No historical business found'
     } else {
@@ -116,8 +109,7 @@ export default class BusinessLookup extends Vue {
     // safety check
     if (that.searchField && that.searchField.length > 2) {
       that.state = States.SEARCHING
-      const searchStatus = this.businessStatus // search for status based on prop
-      that.searchResults = await BusinessLookupServices.search(that.searchField, searchStatus).catch(() => [])
+      that.searchResults = await BusinessLookupServices.search(that.searchField, this.searchStatus).catch(() => [])
 
       // display appropriate section
       that.state = (that.searchResults.length > 0) ? States.SHOW_RESULTS : States.NO_RESULTS
--- a/src/components/new-request/business-lookup.vue
+++ b/src/components/new-request/business-lookup.vue
@@ -94,7 +94,7 @@ export default class BusinessLookup extends Vue {
    * Called when business lookup search status prop has changed.
    * Business lookup text based on the business status to search for.
    */
-  @Watch('businessStatus', { immediate: true })
+  @Watch('searchStatus', { immediate: true })
   onStatuschanged (): void {
     if (this.searchStatus === EntityStates.HISTORICAL) {
       this.lookupLabel = 'Find a historical business'
--- a/src/components/new-request/business-lookup.vue
+++ b/src/components/new-request/business-lookup.vue
@@ -34,7 +34,7 @@
 
       <template v-slot:no-data>
         <p class="pl-5 font-weight-bold">
-          {{  lookupNoActiveText }}
+          {{  lookupNoActiveText  }}
         </p>
         <p class="pl-5">
           Ensure you have entered the correct business name or number.
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -35,29 +35,30 @@
       </v-col>
 
       <!-- Business Lookup/Fetch -->
-      <v-col v-if="showBusinessLookup" cols="12" md="6">
+      <v-col v-if="showBusinessLookup" cols="12" md="6" class="business-lookup">
         <template v-if="!business">
           <BusinessLookup
             v-if="getIsAuthenticated"
             :businessStatus="fetchActiveOrHistorical"
-            @business="business=$event"
+            @business="onBusiness($event)"
           />
-          <BusinessFetch v-else @business="business=$event"/>
+          <BusinessFetch v-else @business="onBusiness($event)"/>
         </template>
-        <div v-else class="d-flex justify-space-between align-center">
-          <v-text-field
-            filled
-            hide-details
-            :value="business.name"
-          />
-          <div @click="business=null">
-            <v-icon color="primary">mdi-close</v-icon>
-          </div>
-        </div>
+        <v-text-field
+          v-else
+          append-icon="mdi-close"
+          readonly
+          filled
+          hide-details
+          :label="business.identifier"
+          :value="business.legalName"
+          @click:append="onBusiness(null)"
+          @keyup.delete="onBusiness(null)"
+        />
       </v-col>
 
       <!-- Jurisdiction -->
-      <v-col v-if="showJurisdiction" cols="12" :md="isSelectedXproRestorable ? 4 : 6">
+      <v-col v-if="showJurisdiction" cols="12" :md="isSelectedXproAndRestorable ? 4 : 6">
         <NestedSelect
           label="Select your home jurisdiction"
           :menuItems="jurisdictionOptions"
@@ -119,27 +120,8 @@
         </v-tooltip>
       </v-col>
 
-      <!-- Business Lookup/Fetch -->
-      <v-col v-if="showBusinessLookup" cols="12" md="6" class="business-lookup">
-        <template v-if="!business">
-          <BusinessLookup v-if="getIsAuthenticated" @business="onBusiness($event)"/>
-          <BusinessFetch v-else @business="onBusiness($event)"/>
-        </template>
-        <v-text-field
-          v-else
-          append-icon="mdi-close"
-          readonly
-          filled
-          hide-details
-          :label="business.identifier"
-          :value="business.legalName"
-          @click:append="onBusiness(null)"
-          @keyup.delete="onBusiness(null)"
-        />
-      </v-col>
-
-      <!-- once an entity type is selected (or Federal)... -->
-      <template v-if="entity_type_cd || isFederal">
+      <!-- once an entity type or business is selected (or Federal) -->
+      <template v-if="entity_type_cd || isFederal || business">
         <!-- Company Type -->
         <v-col v-if="companyRadioBtnApplicable" cols="12">
           <p class="font-weight-bold h6">Select a company type:</p>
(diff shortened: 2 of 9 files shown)


src/components/new-request/business-lookup.vue
```diff
@@ -14,7 +14,7 @@
       hint="Search by name, incorporation or registration number of existing business"
       item-text="name"
       item-value="identifier"
-      label="Find an existing business"
+      :label="lookupLabel"
```
JazzarKarim, line 17: Now the label is set depending on whether we're searching for Active or Historical businesses. Ditto line 37 where the error depends on whether we found active/historical businesses or not.

src/components/new-request/business-lookup.vue
```diff
@@ -65,6 +65,9 @@ enum States {
  */
 @Component({})
 export default class BusinessLookup extends Vue {
+  // Status of businesses to search for prop
+  @Prop({ default: 'ACTIVE' }) readonly businessStatus!: string
```
JazzarKarim, line 69: I've set the default to 'Active' since this what is generally used and hence other Devs have no need to change anything.

src/components/new-request/business-lookup.vue
```diff
@@ -77,17 +80,43 @@ export default class BusinessLookup extends Vue {
...
   }
 
+  /** Called when business lookup search status prop has changed. */
+  @Watch('businessStatus')
+  onStatuschanged (): void {
+    this.setLookupText()
+  }
+
+  /** Business lookup text based on the business status to search for. */
+  private setLookupText (): void {
```
JazzarKarim, line 105: Felt like this is the cleanest way to do it instead of having `v-ifs` in the template.

src/components/new-request/search.vue
```diff
@@ -34,8 +34,33 @@
         <v-text-field filled disabled hide-details label="Select an action first" />
       </v-col>
 
+      <!-- Business Lookup/Fetch -->
+      <v-col v-if="showBusinessLookup" cols="12" md="6" class="business-lookup">
+        <template v-if="!business">
+          <BusinessLookup
+            v-if="getIsAuthenticated"
+            :businessStatus="fetchActiveOrHistorical"
```
JazzarKarim, line 42: This is where whether we are checking if we'll be searching for Active or Historical. The function (line 410) checks if we're in the restoration action. If yes, we check for historical businesses.

src/components/new-request/search.vue
```diff
@@ -97,27 +122,8 @@
...
-          :value="business.legalName"
-          @click:append="onBusiness(null)"
-          @keyup.delete="onBusiness(null)"
-        />
-      </v-col>
-
-      <!-- once an entity type is selected (or Federal)... -->
-      <template v-if="entity_type_cd || isFederal">
+      <!-- once an entity type is selected, is Federal, or is Restorable -->
+      <template v-if="entity_type_cd || isFederal || isRestorable">
```
JazzarKarim, line 126: The reason why I added `isRestorable` here is because when a user selects an entity that's not restorable (like LLP for example), we don't want to show this entire block.

src/components/new-request/search.vue
```diff
@@ -144,11 +150,22 @@
...
               <li>Federally incorporated businesses do not need a Name Request.</li>
-              <li>You may register your extraprovincial business immediately using its existing name
-                at Corporate Online.</li>
+              <li v-if="!isSelectedXproAndRestorable">
+                You may register your extraprovincial business immediately using its existing name
+                at Corporate Online.
+              </li>
+              <li v-else>
+                To reinstate your business, complete
+                <a :href="fullReinstatementFormLink">
```
JazzarKarim, line 162: If a selected restorable business that is also an XPRO is selected, we show this text and form links to download as per the UXPin.

src/components/new-request/search.vue
```diff
@@ -392,11 +426,12 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   get showJurisdiction (): boolean {
     // if (this.isAmalgamation) return true // *** FUTURE
     if (this.isNewXproBusiness) return true
+    if (this.isSelectedXproAndRestorable) return true
```
JazzarKarim, line 429: New conditional whether to show the jurisdiction menu (when the business selected is restorable and an XPRO).

src/components/new-request/search.vue
```diff
@@ -566,16 +611,61 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     this.business = business
     this.entity_type_cd = this.business?.legalType || null
     this.setCorpNum(business?.identifier || null)
+
+    if (this.isRestoration) {
+      // Check if not XPRO and BC restorable
+      if (!this.isSelectedXproAndRestorable && this.isBcRestorable) {
```
JazzarKarim, line 617: - First conditional: If the selected business is restorable and BC based (BENs for example).
- Second Conditional: If the selected business is restorable but XPRO (Limited companies, LLCs for example).
- Third conditional (else statement): If the selected business is not restorable.

src/components/new-request/search.vue
```diff
@@ -566,16 +611,61 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
+        if (this.isRestoration) {
+          return this.isRestorable || 'This business cannot be restored.'
+        }
+        return true
+      }]
+  }
+
+  /** Returns whether the selected XPRO is restorable. */
+  get isSelectedXproAndRestorable (): boolean {
+    return XproMapping.REH.includes(this.business?.legalType)
```
JazzarKarim, line 642: Checking if the legal type of the selected business is included in one of the XPROs that are restorable (I've updated the mapping based on the UXPin).

src/components/new-request/search.vue
```diff
@@ -566,16 +611,61 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
+  }
+
+  /** Returns whether the selected business' legal type is BC and restorable. */
+  get isBcRestorable (): boolean {
+    return BcMapping.REH.includes(this.corpTypeToEntityType(this.business?.legalType as unknown as CorpTypeCd))
```
JazzarKarim, line 647: Checking if the legal type of the selected business is included in one of the BCs.

src/components/new-request/search.vue
```diff
@@ -566,16 +611,61 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
    */
   async incorporateNowClicked () {
     const legalType = this.entityTypeAlternateCode(this.entity_type_cd)
     if (this.getIsAuthenticated) {
-      await this.incorporateNow(legalType)
+      if (!this.isRestoration && !this.business) {
+        await this.incorporateNow(legalType)
+      } else {
+        const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
+        Navigate(`${dashboardUrl}${this.business.identifier}`)
```
JazzarKarim, line 667: When pressing on `Restore Now` button, redirect to dashboard (Filings UI).

src/components/new-request/search.vue
```diff
@@ -657,6 +747,10 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
 
     this.setLocation(jurisdiction.group === 0 ? Location.CA : Location.IN)
     this.setJurisdictionCd(jurisdiction.value)
+    // Prevent entity type from being cleared when jurisdiction is changed
+    if (this.business) {
+      this.setEntityTypeCd(this.business.legalType)
```
JazzarKarim, line 752: When user changes jurisdiction (from Manitoba to United States for example), the entity type was being cleared. In doing that, the check your name button was not being shown (everything is so tangled together).

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-666-cq5bs87k.web.app

src/components/new-request/business-lookup.vue
```diff
@@ -77,17 +80,43 @@ export default class BusinessLookup extends Vue {
   /** State of this component. */
   state = States.INITIAL
 
+  /** Business Lookup labels and warnings. */
+  lookupLabel = ''
+  lookupNoActiveText = ''
+
+  /** Set business lookup text on mount. */
+  mounted (): void {
+    this.setLookupText()
```
severinbeauvais, line 89: You may be able to delete this and add "immediate: true" to the watcher instead.

src/components/new-request/business-lookup.vue
severinbeauvais: If this ends up being called from only 1 place, I recommend moving this code there.

src/components/new-request/business-lookup.vue
severinbeauvais: Can this of type `EntityStates`? (and cascade this type to code below)

severinbeauvais: Also I suggest renaming this `searchStatus` as that's what it is (not a business status).

src/components/new-request/business-lookup.vue
```diff
@@ -77,17 +80,43 @@ export default class BusinessLookup extends Vue {
...
+      this.lookupNoActiveText = 'No active B.C. business found'
+    }
+  }
+
   private onSearchInputDebounced = debounce(async (that: this) => {
     // safety check
     if (that.searchField && that.searchField.length > 2) {
       that.state = States.SEARCHING
-      const searchStatus = 'ACTIVE' // search for ACTIVE
+      const searchStatus = this.businessStatus // search for status based on prop
```
severinbeauvais, line 119: Why not move this right into line 120? 😉

src/components/new-request/search.vue
```diff
@@ -365,6 +384,12 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   activeActionGroup = NaN
   showRequestActionTooltip = false
   business = null as BusinessSearchIF
+  fullReinstatementFormLink = 'https://www2.gov.bc.ca/assets/gov/employment-business-and-economic-development/' +
+    'business-management/permits-licences-and-registration/registries-forms/' +
+    'form_31_xco_-_full_reinstatement_application.pdf'
+  limitedReinstatementFormLink = 'https://www2.gov.bc.ca/assets/gov/employment-business-and-economic-development/' +
+    'business-management/permits-licences-and-registration/registries-forms/' +
+    'form_29_xco_-_limited_reinstatement_application.pdf'
```
severinbeauvais, line 392: 😲

src/components/new-request/search.vue
```diff
@@ -381,6 +406,15 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
   }
 
+  /** Fetch businesses with business lookup depending on the Action selected */
+  get fetchActiveOrHistorical (): String {
+    if (this.isRestoration) {
+      return 'HISTORICAL'
+    } else {
+      return 'ACTIVE'
+    }
+  }
```
severinbeauvais, line 416: Please rename this to "lookup" something, since business-fetch is something else.

Also, how about using a ternary here? 😉

src/components/new-request/search.vue
severinbeauvais: Thanks for the code comments, too!

src/components/new-request/search.vue
```diff
@@ -566,16 +611,61 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
    * If user is authenticated, create draft business and redirect to Dashboard.
+   * If restoration/reinstatement selected, go to business dashboard.
    * If user is not authenticated, redirect to login screen then redirect back.
    */
   async incorporateNowClicked () {
     const legalType = this.entityTypeAlternateCode(this.entity_type_cd)
     if (this.getIsAuthenticated) {
-      await this.incorporateNow(legalType)
+      if (!this.isRestoration && !this.business) {
+        await this.incorporateNow(legalType)
```
severinbeauvais, line 664: I think it would be easier to understand positive logic here, ie, in which cases to we go to business dashboard? Or, in which cases do we incorporate now?

Also, why are you checking "and not business"?

src/components/new-request/business-lookup.vue
JazzarKarim: I was debating on whether I should create an enum for them or not. Apparently, there's already one. The weird thing is I remember searching if there was one. Anyway, thanks for pointing this out Sev! I'll be making this change 🙌

src/components/new-request/search.vue
severinbeauvais: Yes, very. I'm trying for us to be careful to not extend this pattern.

Can you tell what code cleared the legal type, and maybe pre-empt that?

src/components/new-request/search.vue
```diff
@@ -678,6 +772,11 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   padding: 20px 8px !important;
 }
 
+// Change color of the open form icon
+.open-form-icon {
+  color: $app-blue;
+}
```
severinbeauvais, line 778: An alternate way to do this (and might be better here since there are no other class properties) is to set `color="primary"` in the v-icon.

src/enums/entity-type.ts
```diff
@@ -21,12 +21,17 @@ export enum EntityType {
...
   XLL = CorpTypeCd.XPRO_LIMITED_LIABILITY_PARTNERSHIP,
   XLP = CorpTypeCd.XPRO_LIMITED_PARTNERSHIP,
   XSO = CorpTypeCd.XPRO_SOCIETY_NR,
   XUL = CorpTypeCd.XPRO_UNLIMITED_LIABILITY_COMPANY,
 
+  // XPRO Entity types + restorable:
+  A = CorpTypeCd.EXTRA_PRO_A,
+  LLC = CorpTypeCd.LIMITED_CO,
+  XCP = CorpTypeCd.XPRO_COOP,
+  XS = CorpTypeCd.XPRO_SOCIETY,
```
severinbeauvais, line 34: This isn't the place to say they're restorable. I would just add these to the xpro types section above.

src/mixins/common-mixin.ts
```diff
@@ -68,6 +68,21 @@ export class CommonMixin extends Vue {
     }
   }
 
+  /**
+   * Entities UI codes to Name Request Code
+   * @example ULC --> UL
+   */
+  corpTypeToEntityType (entityType: CorpTypeCd): EntityType {
```
severinbeauvais, line 75: Great name!

How about renaming "entityTypeAlternateCode" likewise?

src/store/getters.ts
```diff
@@ -196,10 +196,11 @@ export const isRestoration = (state: StateIF): boolean => {
   return (getRequestActionCd(state) === NrRequestActionCodes.RESTORE)
 }
 
-/** True if current flow is XPRO (includes AMALGAMATION, NEW_BUSINESS, and possibly others). */
+/** True if current flow is XPRO (includes AMALGAMATION, NEW_BUSINESS, RESTORATION, and possibly others). */
 export const getIsXproFlow = (state: StateIF): boolean => {
   return (
-    (getRequestActionCd(state) !== NrRequestActionCodes.MOVE) &&
+    ((getRequestActionCd(state) !== NrRequestActionCodes.MOVE) ||
+    (getRequestActionCd(state) !== NrRequestActionCodes.RESTORE)) &&
```
severinbeauvais, line 203: Sadly, the comment doesn't match the code.

I think  your intent is to **include** RESTORE, but the code is **excluding** it.

It would be better to include the respective action codes but I think that's risky.

RESTORE is not MOVE so the old code here should work, right?

src/components/new-request/business-lookup.vue
JazzarKarim: I'm calling it twice Sev. I'm calling it on mount and when the status is changed (for example moving from Name change to Restoration). Both of them use the business lookup.

src/components/new-request/business-lookup.vue
JazzarKarim: I like this a lot Sev. Never knew that. It works!

src/components/new-request/business-lookup.vue
severinbeauvais: 😁

src/components/new-request/business-lookup.vue
severinbeauvais: 👍 

FYI, the other useful watcher option is "deep" (to watch changed object properties, but there are caveats).

src/store/getters.ts
JazzarKarim: Sev, maybe I'm not understanding correctly. 

So this here, I added an `||` to check if the flow is XPRO (in case the selected business is restorable and XPRO). The reason why I need this is to have access to the select jurisdiction dropdown menu and other fields that are related to XPRO (like MRAS search etc..).

In `search.vue`, there are lots of instances of this getter where we check if we're in XPRO. If this is false, many of the components would not be accessible.

So, I added this to say that if we are in Restoration flow --> there is potential to be in XPRO flow. Hence, this getter would be true. The other logic whether we are in restoration XPRO or not are being handled in `search.vue`.

severinbeauvais: To be honest, I'm kinda confused, but it looks like the old code returned True if this **was not** a MOVE (ie, it was a AML, NEW, REH, etc).

Now it looks it's True if it's not a REH either...? In other words, False for a REH?

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-666-cq5bs87k.web.app

src/store/getters.ts
severinbeauvais: I think this logic is messed up anyway.

`if (!a || !b)` is the same as `if !(a && b)`, right? But it can't be both a and b.

Another way to look at it is, if it's not a MOVE then it will match everything else (including REH), or if it's not a REH then it will match everything else (including MOVE).

src/components/new-request/business-lookup.vue
severinbeauvais: Use of enum + new prop name looks good. OK to resolve.

src/components/new-request/business-lookup.vue
severinbeauvais: Looks good.

src/components/new-request/business-lookup.vue
severinbeauvais: Looks good.

src/components/new-request/business-lookup.vue
severinbeauvais: Looks good.

src/components/new-request/search.vue
severinbeauvais: No change needed. It's just some ugly URLs.

src/components/new-request/search.vue
severinbeauvais: Looks good.

src/components/new-request/search.vue
severinbeauvais: Looks good.

src/mixins/common-mixin.ts
severinbeauvais: Looks good. Thanks!

src/enums/entity-type.ts
severinbeauvais: Looks good.

src/store/getters.ts
JazzarKarim: The logic currently is:
if (a || b) && c. The old code was if (a && c).

My code only adds on additional case to the expression. If MOVE or REH is true, then the OR expression is true. Meaning that, if we are in MOVE flow or RESTORE flow, we want the XPRO flow to essentially be true.

Now, we could have REH and not XPRO. That is being handled in `search.vue` not here. Here, it's just to add that REH can be in an XPRO flow.

severinbeauvais: The old code was `if (!a && c)`. The new code is `if (!a || !b) && c`.

That first part, I think, is always True. Example: "if not blue or not green".

src/components/new-request/search.vue
JazzarKarim: The code that's clearing it is line 744: `this.setLocation(jurisdiction.group === 0 ? Location.CA : Location.IN)`.

However, I'd prefer to leave it as is Sev in case other flows might need it to re-initialize/re-populate some of the variables. Please let me know if that's not OK with you Sev.

severinbeauvais: OK but please change your comment -- it's not "preventing" anything; it's resetting the entity type code etc.

src/store/getters.ts
JazzarKarim: Wait, you are absolutely right. Ohhh, it should be:
`((getRequestActionCd(state) === NrRequestActionCodes.MOVE) || (getRequestActionCd(state) === NrRequestActionCodes.RESTORE))`
This'll only include MOVE and RESTORE. Sev, great great catch. Sorry about that.

severinbeauvais: Wait! The original logic was "everything except MOVE".

I think it would be clearer to say what's included instead of what's excluded.

Also, if it's everything except Move then Restore is already included...?

src/components/new-request/search.vue
severinbeauvais: Thanks.

src/store/getters.ts
JazzarKarim: Yes, you're right. Sev, I'm taking a look at it now again and double checking everything.

JazzarKarim: Sev, like you mentioned. I've changed the getter to specify which flows will include the XPRO flow. This way, it's easier to understand and easier to add to. I've also tested the different work flows, looks OK.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-666-cq5bs87k.web.app

JazzarKarim:
@severinbeauvais @chenhongjing @ketaki-deodhar Please let me know if I can get an approval (if there are no more comments). I'm hoping to merge soon. Thanks a lot!

severinbeauvais (approved):
Thanks for the updates. LGTM!

src/components/new-request/business-lookup.vue
```diff
@@ -34,7 +34,7 @@
 
       <template v-slot:no-data>
         <p class="pl-5 font-weight-bold">
-          No active B.C. business found
+          {{  lookupNoActiveText  }}
```
severinbeauvais, line 37: You don't need double spaces in here. (Actually, you don't need any spaces, but our code style is to include one leading and one trailing space.)

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-666-cq5bs87k.web.app

PR #667, opened 2023-08-31
16641 Update Alteration NR flow

chenhongjing:
*Issue #:* bcgov/entity#16641

*Description of changes:*

- Updated Alteration NR flow


bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

src/components/existing-request/existing-request-display.vue
```diff
@@ -427,6 +432,25 @@ export default class ExistingRequestDisplay extends Mixins(
...
+      this.nr.requestTypeCd === NrRequestTypeCodes.CONVERT_ULBE)
+  }
+
+  get isOpenExternal (): boolean {
+    if (this.showAlterNowButton) {
+      const supportedAlterationTypes = GetFeatureFlag('supported-alteration-types')
+      return !supportedAlterationTypes.includes(this.nr.requestTypeCd)
+    }
+    return false
+  }
```
severinbeauvais, line 452: Please rename this to something that tells me what this value is. Maybe "isSupportedAlteration"?

See also ComminMixin::isSupportedEntity.

Should ine 449 check `requestTypeCd` or `entity_type_cd`?

src/components/existing-request/nr-approved-gray-box.vue
```diff
@@ -36,15 +36,55 @@
         </v-btn>
       </div>
 
+      <div v-else-if="showAlterNowButton" class="my-1">
+        <div v-if="isAllowAlterOnline" class="d-flex justify-center alter-now-btn">
```
severinbeauvais, line 40: Probably remove class "alter-now-btn" here (it's already on line 54).

src/components/existing-request/nr-approved-gray-box.vue
```diff
@@ -36,15 +36,55 @@
...
+            class="alter-now-btn"
+            min-width="20rem"
+            :disabled="disabled"
+            @click="$emit('goToEntityDashboard')"
+          >
+            <strong>Alter Now</strong>
+          </v-btn>
+        </div>
+
+        <div v-else class="contact-registries">
```
severinbeauvais, line 63: Fix indents below. Should be 2 spaces per level.

src/components/existing-request/nr-approved-gray-box.vue
```diff
@@ -36,15 +36,55 @@
...
+              To complete this alteration, please contact us at:
+              <p>
+                <br/>
```
severinbeauvais, line 66: Don't use this for whitespace in this paragraph. Instead, use class mt-x or pt-x in the paragraph element itself.

src/components/existing-request/nr-approved-gray-box.vue
```diff
@@ -82,3 +131,24 @@ export default class NrApprovedGrayBox extends Vue {
...
+  margin-top: 30px;
+  .v-icon.v-icon {
+    color: $app-dk-blue;
+  }
+}
+
+.important-note {
+  margin-top: 30px;
+  margin-bottom: 30px;
+}
```
severinbeauvais, line 153: Since we have multiple designs that spec out 30px whitespace, let's make this generic and add it to layout.scss.

I recommend 2 new classes: `mt-30` and `mb-30` that you can use on line 85 instead.

src/components/new-request/search.vue
```diff
@@ -381,6 +419,31 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
+        }
+        return true
+      }]
+  }
+
+  get isAlterOnline (): boolean {
+    if (!this.isConversion) return true
+    return !(this.getConversionType === NrRequestTypeCodes.CONVERT_BEN ||
+      this.getConversionType === NrRequestTypeCodes.CONVERT_CORP ||
+      this.getConversionType === NrRequestTypeCodes.CONVERT_ULBE)
```
severinbeauvais, line 440: You already have this logic at existing-request-display.vue::isAllowAlterOnline(). Could it be moved to getters.ts instead?

src/components/new-request/search.vue
```diff
@@ -396,6 +459,9 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   }
 
   get showEntityType (): boolean {
+    if (this.isConversion) {
+      return this.business != null && this.isPassValidation
```
severinbeauvais, line 463: 1. This should be `!==` (value and type comparison, even though null has no type because lint may complain)
2. This can actually be `return this.business && ...`

Also on line 504 below.

src/components/new-request/search.vue
```diff
@@ -497,11 +566,21 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
+
   /** Whether to show "Go to COLIN" button (otherwise will show "Incorporate Now" button). */
   get showColinButton (): boolean {
     if (this.showContinueInButton) return true
     if (this.isFederal) return true
 
+    if (this.isConversion) {
+      const supportedAlterationTypes = GetFeatureFlag('supported-alteration-types')
+      return !supportedAlterationTypes.includes(this.getConversionType)
+    }
```
severinbeauvais, line 582: As per comment above, I think you should have this as a common mixin function.

src/components/existing-request/existing-request-display.vue
```diff
@@ -196,10 +196,15 @@
             :emailAddress="nr && nr.applicants && nr.applicants.emailAddress"
             :showIncorporateNowButton="showIncorporateButton"
             :showRegisterButton="showRegisterButton"
+            :showAlterNowButton="showAlterNowButton"
+            :isAllowAlterOnline="isAllowAlterOnline"
+            :isOpenExternal="isOpenExternal"
             :showGoToSocietiesButton="showGoToSocietiesButton"
```
severinbeauvais, line 202: This is now getting messy (too many props that mean almost the same thing). Could their logic be moved into NrApprovedGrayBox instead (ie, make the subcomponent smarter so we don't have to pass in so many props)?

src/components/new-request/search.vue
```diff
@@ -45,9 +45,42 @@
...
+        />
+      </v-col>
+
       <!-- Entity Type -->
-      <v-col v-if="showEntityType" cols="12" :md="getIsXproFlow ? 4 : 6">
+      <v-col
+        v-if="showEntityType" cols="12"
+        :md="getIsXproFlow ? 4 : ( isConversion ? 12 : 6)">
+        <v-text-field
+          v-if="isConversion && isBenBusiness"
```
severinbeauvais, line 77: The logic on lines 74 and 77 is getting complicated (nested ifs). (Also line 75.)

Can you lay out the logic cleaner? I'm thinking maybe using `<template v-if="...">` or separate columns for BEN conversion vs others?

src/components/new-request/search.vue
```diff
@@ -97,29 +130,10 @@
...
-          :label="business.identifier"
-          :value="business.legalName"
-          @click:append="onBusiness(null)"
-          @keyup.delete="onBusiness(null)"
-        />
-      </v-col>
-
       <!-- once an entity type is selected (or Federal)... -->
-      <template v-if="entity_type_cd || isFederal">
+      <template v-if="companyRadioBtnApplicable && (entity_type_cd || getConversionType || isFederal)">
```
severinbeauvais, line 134: I recommend moving the extra logic into `companyRadioBtnApplicable`.

src/components/new-request/search.vue
```diff
@@ -210,41 +225,59 @@
...
-        >
-          <v-btn
-            v-if="showColinButton"
-            class="px-9"
-            id="go-to-colin-button"
-            :href="colinLink"
-            target="_blank"
+          <!-- Go to COLIN / Incorporate Now buttons -->
+          <v-col v-if = "isAlterOnline" cols="12"
+            class="d-flex justify-center"
```
severinbeauvais, line 255: Are you sure you need to make this a flex container? I think v-col already is.

src/components/new-request/search.vue
```diff
@@ -274,7 +307,7 @@
 
 <script lang="ts">
 import { Component, Mixins, Vue, Watch } from 'vue-property-decorator'
-import { Action, Getter } from 'vuex-class'
+import { Action, Getter, State } from 'vuex-class'
```
severinbeauvais, line 310: Remove unused State.

src/components/new-request/search.vue
```diff
@@ -285,12 +318,16 @@ import BusinessFetch from '@/components/new-request/business-fetch.vue'
...
-import { AccountType, CompanyType, EntityType, Location, NrRequestActionCodes, NrRequestTypeCodes } from '@/enums'
+import { AccountType, CompanyType, EntityType, Location,
+  NrRequestActionCodes, NrRequestTypeCodes } from '@/enums'
 import { CommonMixin, NrAffiliationMixin } from '@/mixins'
+import { CorpTypeCd } from '@bcrs-shared-components/enums'
 import { CanJurisdictions, ConversionTypes, Designations, IntlJurisdictions, RequestActions } from '@/list-data'
 import { GetFeatureFlag, Navigate } from '@/plugins'
 import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
 
+import { getVuexStore } from '@/store' // DEGUB
```
severinbeauvais, line 329: Remove.

src/components/new-request/search.vue
```diff
@@ -365,6 +402,7 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
   activeActionGroup = NaN
   showRequestActionTooltip = false
   business = null as BusinessSearchIF
+  isPassValidation = false
```
severinbeauvais, line 405: This is weird because its value is set in a computed (line 427 below), which may not be evaluated when you think. (Computed is evaluated when it's first used, and thereafter if anything it uses has changed.)

Can this be changed to a computed also?

src/components/new-request/search.vue
```diff
@@ -210,41 +225,59 @@
...
-            <li>It is not possible to request a specific Incorporation Number.</li>
-          </ul>
-        </v-col>
+        <template v-if="selectedCompanyType === CompanyType.NUMBERED_COMPANY" cols="12">
+          <v-col v-if="isConversion && !isAlterOnline">
+            <div class="contact-registries">
+              To complete this alteration, please contact us at:
+              <p>
+                <br/>
+                  <v-icon small>mdi-phone</v-icon>  Canada and U.S. Toll Free:
```
severinbeauvais, line 234: Don't use spaces for spacing. Use padding instead (or use the HTML entity `&nbsp;` once).

src/components/new-request/search.vue
```diff
@@ -664,6 +765,10 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
 
+.contact-registries .v-icon.v-icon {
```
severinbeauvais, line 768: I'd be tempted to select the icons directly, like this:
```
.v-icon.mdi-phone,
.v-icon.mdi-email {
 ...
}
```

src/components/existing-request/existing-request-display.vue
chenhongjing: It should check `requestTypeCd`.

src/mixins/common-mixin.ts
```diff
@@ -116,6 +116,12 @@ export class CommonMixin extends Vue {
       nr?.legalType === EntityType.GP)
   }
 
+  /** Returns true if the specified NR is for alteration (conversion). */
+  isSupportedAlteration (type: any): boolean {
```
severinbeauvais, line 120: The other functions here accept a "nr" with unknown type, but here you should know the type of "type" so please add it.

Or change this function to accept a "nr" like the others.

src/components/existing-request/existing-request-display.vue
severinbeauvais: I like that you've moved this logic to CommonMixin.

I'd still like to see "isOpenExternal" named something more specific.

src/components/existing-request/nr-approved-gray-box.vue
severinbeauvais: Awesome, thanks!

src/components/new-request/search.vue
severinbeauvais: This looks fine now, but another way is to add `color="primary"` to the v-icon.

(The primary color is assigned in plugins/vuetify.ts, and is the same value as $app-blue.)

src/components/new-request/search.vue
chenhongjing: I moved it to common-mixin.ts instead of getters.ts because these two are slightly different. They check different variables.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

src/interfaces/new-request-interface.ts
```diff
@@ -17,7 +17,7 @@ export interface NewRequestIF {
   assumedNameOriginal: string
   conditionsModalVisible: boolean
   conflictId: string
-  conversionType: EntityType
+  conversionType: NrRequestTypeCodes
```
severinbeauvais, line 20: 👍

src/mixins/common-mixin.ts
```diff
@@ -92,6 +92,13 @@ export class CommonMixin extends Vue {
     }
   }
 
+  /** Returns true if the specified alteration NR is allowed to be done online. */
+  isAlterOnline (type: NrRequestTypeCodes): boolean {
+    return !(type === NrRequestTypeCodes.CONVERT_BEN ||
+      type === NrRequestTypeCodes.CONVERT_CORP ||
+      type === NrRequestTypeCodes.CONVERT_ULBE)
+  }
```
severinbeauvais, line 100: Do you think we should use a FF for this, so that we can expand it easily in the future?

src/list-data/conversion-types.ts
```diff
@@ -37,7 +37,7 @@ export const ConversionTypes: ConversionTypesI[] = [
   {
     desc: 'Benefit company to a limited company',
     text: 'Benefit company to a limited company',
-    entity_type_cd: EntityType.CR,
+    entity_type_cd: EntityType.BEN,
```
severinbeauvais, line 40: I wonder if you should use `CommonMixin::entityTypeAlternateCode()` to convert from EntityType to CorpTypeCd here (or vice versa)?

I know these code are almost the same thing (EntityType is a subset of CorpTypeCd) but it strikes me as strange that "BEN" or "CorpTypeCd.BENEFIT_COMPANY" isn't used anywhere else in Namerequest -- I think for a BEN NR it actually uses the "BC" code.

src/list-data/conversion-types.ts
```diff
@@ -5,7 +5,7 @@ export const ConversionTypes: ConversionTypesI[] = [
   {
     desc: 'Limited company to an unlimited liability company',
     text: 'Limited company to an unlimited liability company',
-    entity_type_cd: EntityType.UL,
+    entity_type_cd: EntityType.BC,
```
severinbeauvais, line 8: I see you're changing the meaning of entity_type_cd to mean the "from" company instead of the "to" company...

Does the NR that get created have the correct fields set? Please compare with a conversion NR in Dev or Test environment.

src/enums/entity-type.ts
```diff
@@ -4,6 +4,7 @@ import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
 export enum EntityType {
   // BC Entity Types:
   BC = CorpTypeCd.BC_COMPANY,
+  BEN = CorpTypeCd.BENEFIT_COMPANY,
```
severinbeauvais, line 7: This may not be needed, depending on whether conversions are done in conversion-types.ts.

src/mixins/common-mixin.ts
chenhongjing: I don't think we should use FF in this case. Because now anything that is from/to BEN cannot be done online, once it can be done, we can remove this part.

src/list-data/conversion-types.ts
severinbeauvais: You can also use ConversionTypes.value to determine the "from" and "to" value.

In this code, the value is NrRequestTypeCodes.CONVERT_CORP, which means "from BEN to LTD".

What do you think?

src/mixins/common-mixin.ts
severinbeauvais: But that's the point of a FF: we can enable it in real-time without needing to change any code and deploy.
Maybe check with Mihai.

src/components/existing-request/existing-request-display.vue
severinbeauvais: This will be handled in bcgov/entity#17608.

src/components/new-request/search.vue
severinbeauvais: This will be handled in bcgov/entity#17608.

src/components/new-request/search.vue
severinbeauvais: Have a quick look in this PR whether you need the d-flex class (I think you might not) and update the code or reply accordingly. Thanks!

src/components/new-request/search.vue
severinbeauvais: Did this get cleaned up?

src/components/new-request/search.vue
severinbeauvais: Easy fix in this PR, please.

src/mixins/common-mixin.ts
severinbeauvais: Thanks for the type! Resolving.

src/components/new-request/search.vue
severinbeauvais: I see you changed this to a getter (computed). Looks good. Resolving.

src/mixins/common-mixin.ts
severinbeauvais: Summary of what we discussed:

The FF will say which request type codes (eg, BECV, UC) will show "Alter Now" instead of "Go To COLIN"... **but** you also have code that says "from or to BEN will show a special Contact Us panel".

So, BECV will show the panel, UC will show Alter Now, but ULCB (not in FF atm) will show Go To COLIN.

src/list-data/conversion-types.ts
severinbeauvais: As discussed, revert this and maybe use the request type code to check "from" and "to" entity type (since the request type codes have a unique from and to).

src/list-data/conversion-types.ts
severinbeauvais: This is another thing we discussed: we need to compare a NR created with your code with a NR created in Dev and make sure all NR fields look the same.

src/components/new-request/search.vue
chenhongjing: Yes, need this one.

src/components/new-request/search.vue
chenhongjing: It has been fixed, but it doesn't update at this page. It should be outdated.

severinbeauvais: The latest files don't show the fix. Maybe you have a commit to push?

src/components/new-request/search.vue
```diff
@@ -389,13 +439,23 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     return (this.getRequestActionCd === NrRequestActionCodes.NEW_BUSINESS && this.request?.group === 1)
   }
 
+  get isPassAlterValidation (): boolean {
+    const legalType = this.business.legalType as string
```
JazzarKarim, line 443: Hongjing, one other way that you can cast this since you want to compare to CorpTypeCd is:
`const legalType = this.business.legalType as unknown as CorpTypeCd`

This is instead of casting it as a string.

JazzarKarim (approved):
LGTM after fixing remaining comments!

src/components/new-request/search.vue
```diff
@@ -65,7 +100,7 @@
                 @change="setClearErrors()"
                 hide-details="auto"
                 filled
-                v-model="entity_type_cd"
+                v-model="entity_type_options_select_bind"
```
chenhongjing, line 103: I got a rendering issue if binding `entity_type_cd` for alteration because it changed as we chose different alterations. Current solution for it is to change it to `origin_entity_type_cd` for alteration only.

severinbeauvais: Let's go with this solution for now.

Could you add a comment on line 516 explaining why it's different for conversion? Thanks.

src/components/new-request/search.vue
```diff
@@ -639,13 +654,19 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
     this.setEntityTypeCd(this.business?.legalType)
-    if (this.isConversion && this.isBenBusiness) {
-      this.setConversionType(NrRequestTypeCodes.CONVERT_CORP)
-    }
-    // clear all related fields when clearing business search/fetch for alter
-    if (this.isConversion && !this.business) {
-      this.setConversionType(null)
-      this.selectedCompanyType = null
+
+    if (this.isConversion) {
```
severinbeauvais, line 658: Add comment explaining what you're doing here.

src/interfaces/new-request-interface.ts
```diff
@@ -55,6 +55,7 @@ export interface NewRequestIF {
   nrOriginal: Partial<NameRequestI>
   nrRequestNameMap: RequestNameMapI[]
   nrRequiredModalVisible: boolean
+  origin_entity_type_cd: EntityType
```
chenhongjing, line 58: I add this to save the `from` entity type because the request type cannot be determined before we select one from the allowable alteration list. However, `from` entity type determines which should be shown at this list.

src/list-data/conversion-types.ts
severinbeauvais: I see you've added a variable for "original cd". That's fine by me for now.

src/enums/entity-type.ts
severinbeauvais: Never mind. It's now needed.

src/list-data/conversion-types.ts
severinbeauvais: ^^ lmk result of your testing.

src/mixins/common-mixin.ts
severinbeauvais: Hongjing, please verify that your code matches what we discussed, and then you can resolve this conversation. Thanks.

src/enums/entity-type.ts
chenhongjing: Is it better to keep it since it's referred by `origin_entity_type_cd`?

severinbeauvais: Yes, or use the common mixin method to convert BC -> BEN (or however that works).

Karim, do you have a strong opinion on this? It's the old NR vs Entity type code mismatch thing.

src/components/new-request/search.vue
```diff
@@ -510,6 +512,15 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     return this.getEntityBlurbs?.find(type => type.value === entity_type_cd)?.blurbs
   }
 
+  get entity_type_options_select_bind (): EntityType | NrRequestTypeCodes {
+    if (this.isConversion) return this.getOriginEntityTypeCd
```
chenhongjing, line 516: > I got a rendering issue if binding `entity_type_cd` for alteration because it changed as we chose different alterations. Current solution for it is to change it to `origin_entity_type_cd` for alteration only.

So the issue is when we choose type of business alter into, the box cannot show value we select the first time we click on this value. But the value can be shown if we click on the same value twice.
image

src/components/new-request/search.vue
severinbeauvais: Add comment to code, please :)

src/mixins/common-mixin.ts
chenhongjing: Yes, verified!

src/enums/entity-type.ts
JazzarKarim: No strong opinions, both approaches are OK. I think keeping it as is should be fine Sev.

src/components/new-request/search.vue
severinbeauvais: Hmm, that needs to be fixed.

It sounds like a reactivity issue: something is not updating right away.

severinbeauvais: Do you want to merge ASAP and then look into this one (same ticket, new PR)?

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

src/components/new-request/search.vue
severinbeauvais: Is it broken for other NR request  types? More specifically, what is broken right now?

src/components/new-request/search.vue
severinbeauvais: The new comments aren't as detailed as your comments above (which I think are super useful).

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

JazzarKarim:
I just verified, all my changes are working perfectly after rebasing. I know this must've been not fun at all. Thanks Hongjing! 🙌

src/components/new-request/search.vue
chenhongjing: No, it works well now 😸

severinbeauvais: OK, you can resolve this if you want.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

severinbeauvais:
Shouldn't this prompt me for designations?

image

severinbeauvais:
I created an alteration NR on the temp URL and got this. The NR type (BEN) is incorrect. It should be BC (which is a BEN in Namerequest).

image

src/components/new-request/search.vue
```diff
@@ -607,6 +709,25 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
...
     this.entity_type_cd = this.business?.legalType || null
     this.setCorpNum(business?.identifier || null)
+    this.setEntityTypeCd(this.business?.legalType)
+
+    if (this.isConversion) {
+      if (this.business) {
+        // set the from business for alteration (conversion)
+        this.setOriginEntityTypeCd(this.business.legalType)
+        // special case if the from business is BEN
+        // set conversionType & entity_type_cd because there's only one alteration type for it
```
severinbeauvais, line 719: Good comment, thanks!

I'll test the "BEN" error as soon as the CI-PR build finishes.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

src/components/new-request/search.vue
severinbeauvais: Nice! This change fixed both issues I found (designation + BEN). 🎉 

image

severinbeauvais: (This conversation can be resolved.)

JazzarKarim:
@severinbeauvais The corp number should also be populated with the business identifier here as with restorations right?
corpnum empty

severinbeauvais:
> @severinbeauvais The corp number should also be populated with the business identifier here as with restoration workflow right? corpnum empty

Exactly. Alteration, restoration, name change.

JazzarKarim:
> Exactly. Alteration, restoration, name change.

Alright. Hongjing, don't worry about that for now (so that you can move this ticket through). I'll fix that in both our flows. I'll create a new ticket for it. Is that OK with you Sev? @severinbeauvais

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.0.34",
+  "version": "5.0.35",
   "lockfileVersion": 1,
```
severinbeauvais, line 4: I don't know how I missed this, but this should be lockfile version 2.

I think you are using an old version of Node. It should be:

```
severin@DESKTOP-2XZPXD3:~$ node --version
v16.14.2
```

severinbeauvais: Also make sure you use `npm i` to re-create the package lock file.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-667-z8piuhtc.web.app

package-lock.json
severinbeauvais: Looks fixed to me :)

chenhongjing:
Ready to merge?! 👀

src/list-data/conversion-types.ts
severinbeauvais: This will be specifically tested in #17614. Resolving.

PR #671, opened 2023-09-07
17553 business lookup update

vysakh-menon-aot:
*Issue #:* /bcgov/entity#17553

*Description of changes:*


src/components/new-request/business-fetch.vue
```diff
@@ -61,7 +61,7 @@ export default class BusinessFetch extends Vue {
   private validate (): boolean {
     if (!this.searchField) {
       this.errorMessages = ['Required field']
-    } else if (!/^(A|BC|C|CP|FM)\d{7}$/.test(this.searchField)) {
+    } else if (!/^(A|BC|C|CP|FM|XCP|CCC|CUL)( |)\d{7}$/i.test(this.searchField)) {
```
severinbeauvais, line 64: I think it would be better to trim searchField than to allow a trailing space.

I believe the field is already bound like this: `v-model.trim="searchField"` but it may still require an explicit trim() before validating it. (Vuetify has always had this behaviour.)

severinbeauvais: Please test lower-case identifiers with the "COLIN search" (the search that is used when not logged in). Eg, try "cp0000840".

src/services/namex-services.ts
```diff
@@ -141,9 +141,8 @@ export default class NamexServices {
    */
   static async searchColin (corpNum: string): Promise<any> {
     // remove BC prefix as COLIN only supports base number with no prefix for BC's
-    const cleanedCorpNum = corpNum.replace(/^BC+/i, '')
-    const url = `${appBaseURL}/colin/${cleanedCorpNum}`
-    return this.axios.post(url).then(response => response.data)
+    const url = `${appBaseURL}/colin/${corpNum}`
+    return this.axios.get(url).then(response => response.data)
```
severinbeauvais, line 145: Awesome! But shouldn't this be a new Namex API endpoint?

severinbeauvais (approved):
I'm happy with this when you're happy with it.

As you mentioned this morning, there may be future changes in this area (same ticket) and that's OK by me.

severinbeauvais:
PS You need to rebase to main branch.

src/services/namex-services.ts
vysakh-menon-aot: this should be the new endpoint https://github.com/bcgov/namex/pull/1413#discussion_r1321763306

severinbeauvais: It looks like the same endpoint to me...?

vysakh-menon-aot: existing endpoint is POST and the new one is GET

severinbeauvais: Ah, OK, good.

Would it make sense to call it something other than "COLIN"?

vysakh-menon-aot: I think colin make more sense since we are searching in colin db

severinbeauvais: OK, leave it as is then.

(We search both LEAR and COLIN, right?)

src/store/actions.ts
```diff
@@ -313,15 +313,11 @@ export async function searchBusiness ({ getters }, corpNum: string): Promise<Bus
...
-        // *** TODO: update this when new endpoint is available
         return {
-          identifier: corpNum, // data['incorp #'],
-          legalName: 'Future Name From COLIN',
-          legalType: EntityTypes.BC,
-          state: EntityStates.ACTIVE
+          identifier: data.business.identifier,
+          legalName: data.business.legalName,
+          legalType: data.business.legalType,
+          state: data.business.state
```
vysakh-menon-aot, line 320: @severinbeauvais do we require state value? it doesn't seems to be used anywhere. this api call doesn't return state (pr is still in review)

src/services/namex-services.ts
vysakh-menon-aot: yes

src/store/actions.ts
severinbeauvais: I think we need to know if the business is active or historical ...?

vysakh-menon-aot: not sure if colin has the same list of state as lear (i believe active should be there)

severinbeauvais:
Vysakh, when you're happy with this...
- consider merging into `feature-search-refactor` branch (where current development is happening)
- update the app version

src/store/actions.ts
severinbeauvais: Does the UI need to know whether the returned business is active or historical? Or that is a search query parameter?

vysakh-menon-aot: I don't see its been used in the UI, probably not required

severinbeauvais: Check with @JazzarKarim . I don't remember if the UI needs to filter for active/historical or whether that's handled by the API call (ie, only return active or historical or all).

severinbeauvais:
@vysakh-menon-aot We are done with the feature branch. Please rebase to main branch. Sorry for the back and forth!

src/enums/entity-types.ts
```diff
@@ -23,7 +23,6 @@ export enum EntityTypes {
   SP = CorpTypeCd.SOLE_PROP,
   UL = CorpTypeCd.UNLIMITED_LIABILITY_COMPANY,
   ULC = CorpTypeCd.BC_ULC_COMPANY,
-  C = CorpTypeCd.CONTINUE_IN,
```
severinbeauvais: I'm surprised this duplicate key didn't cause an error, or did it? Anyway, good fix 👍

src/services/namex-services.ts
```diff
@@ -141,7 +141,8 @@ export default class NamexServices {
    */
   static async searchColin (corpNum: string): Promise<any> {
     // remove BC prefix as COLIN only supports base number with no prefix for BC's
-    const url = `${appBaseURL}/colin/${corpNum}`
+    const cleanedCorpNum = corpNum.replace(/^BC+/i, '')
+    const url = `${appBaseURL}/colin/${cleanedCorpNum}`
```
severinbeauvais, line 145: Does this only search by corp num, or also by name?

Is it OK to let through ids like CPxxx or FMxxx, etc?

vysakh-menon-aot: search only with corp num. looks like only for BC prefix we need to search with number

severinbeauvais: Oh, right. This is called when a user enters a business id to find (for name change, alteration, etc). All good 👍

severinbeauvais (approved):
LGTM

PR #674, opened 2023-09-07
17468 Refactored search.vue + implemented xpro aml flow

severinbeauvais:
*Issue #:* bcgov/entity#17468

**This is incomplete code! What I have in mind is merging to this to a feature branch where several of us can rebuild the flows using the new architecture. And we can compare these flows to what's currently in Dev.**

*Description of changes:*
    - app version = 5.1.0
    - implemented sub-components for search.vue blocks
    - refactored search.vue to separate flows using sub-components (STILL WIP)
    - fixed tab title size on small screens
    - added search variables to store
    - created search-mixin.ts for common getters, setters, etc.
    - misc cleanup


src/list-data/entity-type-data.ts
```diff
@@ -296,14 +299,8 @@ export const EntityTypesXproData: EntityI[] = [
...
     ],
     amlBlurbs: [
-      [
-        `Register an amalgamation that occurred in another jurisdiction where at least one of the companies is
-           extraprovincially registered in BC.`
-      ],
-      [
-        `Register an amalgamation or merger that occurred in another jurisdiction where at least one of the companies
-          is extraprovincially registered in BC.`
-      ]
```
severinbeauvais, line 306: This used to be an array to distinguish between Canadian and International xpro amalgamations. Since the dialog tooltip is displayed before the home jurisdiction is selected, I combined these blurbs.

Now looks like this:
image

src/list-data/entity-type-data.ts
```diff
@@ -235,6 +235,9 @@ export const EntityTypesBcData: EntityI[] = [
       'Has name protection in BC',
       'Must use Societies Online to register a name and incorporate'
     ],
+    amlBlurbs: [
+      'Society amalgamation'
+    ],
```
severinbeauvais, line 240: This was needed otherwise an empty tooltip displayed.

src/list-data/entity-type-data.ts
```diff
@@ -280,7 +283,7 @@ export const EntityTypesBcData: EntityI[] = [
 
 export const EntityTypesXproData: EntityI[] = [
   {
-    text: 'Limited Company',
+    text: 'Extraprovincial Limited Company',
```
severinbeauvais, line 286: See screenshot attached to comment above for the new labels (to match UI design).

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -34,29 +35,41 @@
         </v-card-text>
       </template>
 
-      <!--  List Tables -->
-      <template v-else-if="!showSocietiesInfo">
-        <v-card-text class="">
+      <!-- Special case for societies -->
+      <template v-if="showSocietiesInfo">
```
severinbeauvais, line 39: I simply moved this up from below.

(I think it's easier to understand "if true else everything else" instead of "if not true ...".)

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -68,17 +81,40 @@
         </v-card-text>
       </template>
 
-      <template v-else>
```
severinbeauvais, line 71: This block now starts on line 54 >>>

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -68,17 +81,40 @@
...
+      <template v-if="isAmalgamation">
+        <v-row class="pt-4">
+          <v-col cols="12">
+            <span class="copy-small">Extraprovincial:</span>
+          </v-col>
+        </v-row>
+
+        <v-card-text class="pt-4">
+          <v-row no-gutters>
+            <v-col v-for="(category, i) in tableDataXPRO" :key="`cat-xpro-${i}`">
```
severinbeauvais, line 94: This is the second block for amalgamations only. Note that it displays Xpro data directly. (The block above, for amalgamations, displayed BC data.)

See screenshot attached to comment above for example.

src/components/dialogs/pick-entity-or-conversion.vue
```diff
@@ -194,20 +233,21 @@ export default class PickEntityOrConversionDialog extends CommonMixin {
     return output
   }
 
-  get width (): string {
-    if (this.showSocietiesInfo || this.isConversion) {
+  get maxWidth (): string {
+    if (this.showSocietiesInfo || this.isConversion || this.isAmalgamation) {
```
severinbeauvais, line 237: Needed for double-wide table layout.

See screenshot attached to comment above for example.

src/components/existing-request/existing-request-display.vue
```diff
@@ -194,16 +194,10 @@
             :nrNum="nr && nr.nrNum"
             :approvedName="approvedName && approvedName.name"
             :emailAddress="nr && nr.applicants && nr.applicants.emailAddress"
-            :showIncorporateNowButton="showIncorporateNowButton"
-            :showRegisterButton="showRegisterButton"
-            :showAlterNowButton="showAlterNowButton"
-            :isAllowAlterOnline="isAlterOnline(nr.requestTypeCd)"
-            :showOpenExternalIcon="showOpenExternalIcon"
-            :showGoToSocietiesButton="showGoToSocietiesButton"
```
severinbeauvais, line 202: I moved these (and the supporting getters/methods below) to the sub-component itself.

src/components/new-request/search.vue
```diff
@@ -739,17 +754,17 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin) {
     }
   }
 
-  /** Returns whether the selected XPRO is restorable. */
```
severinbeauvais, line 742: Small technical point: this doesn't "return" anything (like a function would). Instead, it "is" a value.

(FYI, getters aka computed are evaluated the first time they're needed, then they're cached until something in them changes. So they can be faster than a function!)

JazzarKarim:
Sev, I'll stop doing any work in this UI after merging my latest small PR.

src/components/dialogs/mras-search-info.vue
```diff
@@ -16,10 +16,7 @@
         <p>{{ resultDesc  }}</p>
         <p>{{ resultAct }}</p>
 
-        <NameInput
-          id="name-input-component"
-          :is-mras-search="!isNameSearch"
-        />
+        <NameInput :is-mras-search="!isNameSearch" />
```
severinbeauvais, line 19: `id` is already set in the component.

src/components/tabs.vue
```diff
@@ -89,7 +89,6 @@ export default class Tabs extends Mixins(CommonMixin) {
 
 .mobile-font {
   font-size: $px-14;
-  max-width: 120px;
```
severinbeauvais, line 92: Fix for tab title:

image

src/interfaces/new-request-interface.ts
```diff
@@ -61,6 +60,12 @@ export interface NewRequestIF {
   request_action_cd: NrRequestActionCodes
   request_jurisdiction_cd: string
   requestExaminationOrProvideConsent: RequestOrConsentIF
+  search: {
+    business: BusinessSearchIF
+    companyType: CompanyTypes
+    jurisdiction: any
+    request: RequestActionsI
+  }
```
severinbeauvais, line 68: I struggled for a while with reactivity on variables in the search mixin and the components sometimes not seeing the changes. Ultimately, it looks like there can only be 1 watcher or reactive statement for them. So, I moved the variables to the store instead, and now reactivity works as expected: everyone can see all changes,

src/views/landing.vue
```diff
@@ -16,6 +16,12 @@
             </keep-alive>
           </transition>
         </div>
+
+        <!-- FOR DEBUGGING -->
```
severinbeauvais, line 20: Once again, I'll delete this and the block below before my final commit.

src/components/new-request/search.vue
severinbeauvais, line 1: I'm doing some more cleanup in here before we merge this PR (to make it easier for several people to work in here at the same time).

JazzarKarim (approved):
I get now the direction you want to push the code towards. The code already looks much better and easier to work in / understand.

src/components/new-request/search-components/corp-number-checkbox.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <div id="corp-number-checkbox">
```
severinbeauvais, line 2: Note: this component is NOT a v-col like the others.

src/components/new-request/search-components/entity-type.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+  <v-col id="entity-type" :cols="cols" :md="md">
+    <v-text-field
+      v-if="isConversion && isBenBusiness"
```
severinbeauvais, line 4: This logic should be moved to the parent.

src/components/new-request/search-components/numbered-company-bullets.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <v-col id="numbered-company-bullets" :cols="cols" :md="md">
+    <template v-if="isConversion && !isAlterOnline(getConversionType)">
```
severinbeauvais, line 3: Here's another special case, where this component always displays but, depending on the conditionals, some code block or another is displayed.

My point is that logic on whether to display a component belongs in the parent, but within a component there can be some special cases.

src/components/new-request/search.vue
severinbeauvais, line 1: OK, big changes here.

1. most code blocks are now sub-components
2. there is a search-mixin that contains common imports and getters, etc
3. some components still contain some logic that is specific to them only
4. some of the common search variables are now in the store (for reactivity)
5. all flows are now separate!
6. there is a bit of common code at the top of the template (request action menu) and also at the bottom (buttons)
7. logic for displaying components should be in here, not in the sub-component (except for some special cases)

severinbeauvais:
There's no point in creating a temp URL for this as we're all going to run it locally shortly :)

Once this is merged, I'll set up a quick meeting to walk you through the changes and how I expect the new code to look.

JazzarKarim:
> There's no point in creating a temp URL for this as we're all going to run it locally shortly :)
> 
> Once this is merged, I'll set up a quick meeting to walk you through the changes and how I expect the new code to look.

Sounds good!

leodube-aot (approved):
LGTM!

ketaki-deodhar (approved):
👍

PR #675, opened 2023-09-07
Updated Restoration/Reinstatement Flow in response to Yui's comment

JazzarKarim:
*Issue #:* /bcgov/entity#16640

*Description of changes:*
- **Yui's comment**: When firm registration numbers are entered, the UI returns a result but the "This business cannot be restored" message does not appear until the field loses focus. The error message should appear right after the business name is retrieved.
- This PR also fixes this issue for alterations. It fixes the field focus and validation when we select a business that cannot be altered.

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -47,20 +47,19 @@
             @business="onBusiness($event)"
           />
         </template>
-        <v-text-field
-          v-else
-          append-icon="mdi-close"
-          readonly
-          filled
-          :label="businessLookupLabel"
-          :value="business.legalName"
-          :rules="rules"
-          :hint="businessLookupHint"
-          persistent-hint
-          autofocus
-          @click:append="onBusiness(null)"
-          @keyup.delete="onBusiness(null)"
-        />
+        <v-form v-else ref="businessLookupTextFieldForm">
+          <v-text-field
+            append-icon="mdi-close"
+            readonly
+            filled
+            :label="businessLookupLabel"
+            :value="business.legalName"
+            :rules="rules"
+            :hint="businessLookupHint"
+            @click:append="onBusiness(null)"
+            @keyup.delete="onBusiness(null)"
+          />
+        </v-form>
       </v-col>
 
       <!-- Jurisdiction -->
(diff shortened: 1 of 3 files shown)


src/components/new-request/search.vue
```diff
@@ -47,20 +47,19 @@
...
-          :label="businessLookupLabel"
-          :value="business.legalName"
-          :rules="rules"
-          :hint="businessLookupHint"
-          persistent-hint
-          autofocus
-          @click:append="onBusiness(null)"
-          @keyup.delete="onBusiness(null)"
-        />
+        <v-form v-else ref="businessLookupTextFieldForm">
```
JazzarKarim, line 50: I wrapped the text field in a v-form so that I can call the `validate()` function of the ref.

JazzarKarim:
[screen-capture (4).webm](https://github.com/bcgov/namerequest/assets/122301442/e39997b9-2b0f-4bc4-a385-094656eb08fc)

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-675-jao3khdh.web.app

src/components/new-request/search.vue
severinbeauvais: You can call `validate()` on the component itself.

FYI, forms catch events (like Enter or Escape) and do something with that (like reload the page). If you use a form, test thoroughly to avoid unwanted behaviours.

severinbeauvais (approved):
I approve the general solution, but I think you shouldn't use a form 😛

src/components/new-request/search.vue
JazzarKarim: It didn't work. I tried calling the function on the v-text-field itself, the text field wasn't triggering the validation.

On it. I'll test thoroughly.

severinbeauvais: Weird it didn't work.

OK, leaving it with you. Form is OK if that works fine. OK to resolve.

JazzarKarim: I tested thoroughly, everything looks good Sev! 🙌

chenhongjing (approved):
LGTM 😺

ketaki-deodhar (approved):
Looks good!

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-675-jao3khdh.web.app

PR #676, opened 2023-09-08
Fixed corp num not being populated in CHG, REH, and CNV

JazzarKarim:
*Issue #:* /bcgov/entity#17633

*Description of changes:*
- The `corpNum` field is now being populated with the business identifier after submitting a CHG, REH, or CNV
- An example after submitting a Restoration:
corp num now works

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -164,7 +164,6 @@ export const resetAnalyzeName = ({ commit, getters }): void => {
   if (!getters.getUserCancelledAnalysis) {
     commit('mutateAnalysisJSON', null)
   }
-  commit('mutateCorpNum', '')
   commit('mutateEditMode', false)
   commit('mutateSubmissionType', 'normal')
   commit('mutateShowActualInput', false)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "5.0.40",
+  "version": "5.0.41",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "5.0.40",
+      "version": "5.0.41",
       "dependencies": {
         "@babel/compat-data": "^7.12.13",
         "@bcrs-shared-components/breadcrumb": "2.1.24",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.0.40",
+  "version": "5.0.41",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",


src/store/actions.ts
```diff
@@ -164,7 +164,6 @@ export const resetAnalyzeName = ({ commit, getters }): void => {
   if (!getters.getUserCancelledAnalysis) {
     commit('mutateAnalysisJSON', null)
   }
-  commit('mutateCorpNum', '')
```
JazzarKarim, line 167: After doing lots of digging, I found out that every time, the `corpNum` is being set as an empty string here. There was nothing to counter this for Alterations, Restorations, and Name Change.

So, I moved this line to the else statement below.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-676-mfoo44bx.web.app

src/store/actions.ts
```diff
@@ -268,6 +267,8 @@ const commitExistingData = ({ commit, getters }): void => {
   }
   if (getters.getNr.corpNum) {
     commit('mutateCorpNum', getters.getNr.corpNum)
+  } else {
+    commit('mutateCorpNum', '')
```
severinbeauvais, line 271: I don't think the purpose of this action ("commitExistingData") is to clear fields in the newRequestModel object, so I don't understand why you're doing this here.

src/store/actions.ts
JazzarKarim: Sev, please check latest commit.

Here's my thought process: The `corpNum` is being manually set in the code in a couple of places. When it's been manually set (like in REH for example), I don't want to run mutate the `corpNum` to being an empty string.

severinbeauvais: This "feels" better. It's clearing the value in the same place as the others, and there's similar code to yours just above. So it's probably OK.

If we had time, it would be nice to understand why `resetAnalyzeName` is being called after we set a value. But we don't, so if this works then I'm fine with it.

severinbeauvais: How about adding a comment that we set corpNum earlier in the path and we don't want to reset it here while analyzing the name?

(Now it "feels" like corpNum has nothing to do with name analysis, so why is it being cleared here at all? Sigh.)

severinbeauvais (approved):

src/store/actions.ts
severinbeauvais: See comments above. Resolving.

severinbeauvais:
Karim, I think you need to rebase and update the app version, and then you can merge this.

src/store/actions.ts
JazzarKarim: You mean add a comment in the code explaining why is the if statement there Sev and why are we setting the corpNum as an empty string? I didn't quite get that.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-676-mfoo44bx.web.app

src/store/actions.ts
severinbeauvais: Yes, why is this here / needed? But this may all be obsolete if you remove corpNum from the name analysis flow (since it's not at all related, I think).

The root cause is that we previously did name analysis and then prompted for the corpNum, but now we look up the corpNum and then do the name analysis.

JazzarKarim: Yes, agreed. This line was written 2 years ago. I think it is indeed obsolete. Going to take a deeper dive and make sure if it's obsolete, I can remove the line all together.

JazzarKarim: I've confirmed Sev. This line is indeed obsolete.

I've tried going through the process of creating an NR through various flows after removing this line. The corpNum is being properly set when REH is selected for example while it's empty in a NEW flow.

severinbeauvais: Thanks for testing. Are the other 2 flows OK? And so this PR is not needed?

severinbeauvais: Awesome. This conversation is fine by me (ok to resolve).

severinbeauvais: Why don't you merge this fix and open another PR for other fixes?

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-676-mfoo44bx.web.app

src/store/actions.ts
JazzarKarim: Sounds good. I'm just waiting for another approval!

ketaki-deodhar (approved):
LGTM!

PR #685, opened 2023-09-13
Refactored start a new BC based business

JazzarKarim:
*Issue #:* /bcgov/entity#17727

*Description of changes:*
- Completed refactoring of the start a new BC based business WoN

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -24,7 +24,7 @@
               @emit-corp-num-validity="corpNumValid = $event"
             />
           </v-col>
-          <Designation cols="12" md="4" />
+          <Designation v-if="showDesignation" cols="12" md="4" />
         </template>
       </template>
 
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -298,6 +298,8 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
 
   get showActionNowButton (): boolean {
     if (
+
+    // If we're in Start a new BC based and the entity is supported, show incorporate now button.
       this.isAmalgamation &&
       this.isNumberedCompany &&
       this.isSupportedAmalgamation(this.getEntityTypeCd)
@@ -342,6 +344,8 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
     ) return true
 
     // *** TODO: add your logic here instead of the spaghetti below
+    // If we're in Start a new BC based and the entity is not supported, show go to Colin button.
+
     if (this.isNewBcBusiness && this.isNumberedCompany) {
       return !this.showActionButton
     }
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -17,7 +17,7 @@
         <CompanyType v-if="getEntityTypeCd && isNumberedEntityType" />
         <NumberedCompanyBullets v-if="isNumberedCompany && isNumberedEntityType" />
 
-        <template v-if="isNamedCompany || !isNumberedEntityType">
+        <template v-if="(isNamedCompany || !isNumberedEntityType) && entity_type_cd">
           <v-col cols="12" :md="showDesignation ? '8' : '12'">
             <NameInput
               :is-mras-search="(isXproFlow && isMrasJurisdiction && !getHasNoCorpNum)"
@@ -415,7 +415,7 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
 
   get showCheckNameButton (): boolean {
     // Show button if we're in "Start a new BC-based business" and non-numbered entity is selected.
-    if (this.isNewBcBusiness && !this.isNumberedEntityType) {
+    if (this.isNewBcBusiness && !this.isNumberedEntityType && this.entity_type_cd) {
       return true
     }
 
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -207,10 +207,10 @@ import NumberedCompanyBullets from '@/components/new-request/search-components/n
 import RequestAction from '@/components/new-request/search-components/request-action.vue'
 import XproFederalBullets from '@/components/new-request/search-components/xpro-federal-bullets.vue'
 
-import { CompanyTypes, EntityTypes } from '@/enums'
+import { EntityTypes } from '@/enums'
 import { CommonMixin, NrAffiliationMixin, SearchMixin } from '@/mixins'
 import { Designations } from '@/list-data'
-import { GetFeatureFlag, Navigate } from '@/plugins'
+import { Navigate } from '@/plugins'
 
 /**
  * This is the component that displays the new NR menus and flows.
@@ -297,9 +297,8 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
   }
 
   get showActionNowButton (): boolean {
+    // Conditional for Amalgamation Flow.
     if (
-
-    // If we're in Start a new BC based and the entity is supported, show incorporate now button.
       this.isAmalgamation &&
       this.isNumberedCompany &&
       this.isSupportedAmalgamation(this.getEntityTypeCd)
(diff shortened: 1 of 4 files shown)


JazzarKarim:
**I've tested every single possible business type and compared it to DEV. It looks exactly similar now.**

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-685-avcajisb.web.app

src/components/new-request/search.vue
```diff
@@ -14,7 +14,7 @@
       <!-- New BC Business flow -->
       <template v-if="isNewBcBusiness">
         <EntityType v-if="getLocation" />
-        <CompanyType v-if="entity_type_cd" />
+        <CompanyType v-if="showCompanyTypeRadioButtons" />
```
severinbeauvais, line 17: Do you want to take a stab at simplifying this getter?

```
get showCompanyTypeRadioButtons (): boolean {
  if (this.isNewBcBusiness && ...) return true
```

^^ The fallback return value should be False... probably need to comment out the existing code like I did for the other buttons.

severinbeauvais: I have a better idea!

Instead of using the old getter above, include the logic right in the template. It should be simple for this flow. Try this:
```
<CompanyType v-if="getEntityTypeCd && isNumberedEntityType" />
```

JazzarKarim: It works perfectly! So, are we going to get rid of that weird `showCompanyTypeRadioButtons`?

Also, this PR is incomplete. Apologies for that. I still need to handle the buttons.

severinbeauvais: You said it -- it's a weird (gross) getter. I'd love to get rid of it.

JazzarKarim: Sounds good.

src/components/new-request/search.vue
```diff
@@ -237,22 +237,29 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
...
-    if (!this.isNumberedEntityType || isSociety) {
-      this.setSearchCompanyType(CompanyTypes.NAMED_COMPANY)
-      return false
-    }
-    return true
+    // if (!this.isNumberedEntityType || isSociety) {
+    //   this.setSearchCompanyType(CompanyTypes.NAMED_COMPANY)
+    //   return false
+    // }
+    return false
```
severinbeauvais, line 250: Yes, thanks. This should discourage anyone from using this code :)

src/components/new-request/search.vue
```diff
@@ -14,7 +14,7 @@
       <!-- New BC Business flow -->
       <template v-if="isNewBcBusiness">
         <EntityType v-if="getLocation" />
-        <CompanyType v-if="entity_type_cd" />
+        <CompanyType v-if="getEntityTypeCd && isNumberedEntityType" />
```
severinbeauvais, line 17: You probably need to add a check for Societies, since it's conditional on a FF.

JazzarKarim: Sev, in my last commit, I fixed all the issues resulting from conditionals by adding them to the template.

I don't know if it's kind of ugly, but I don't think there's escaping them. Please let me know what you think. Now, all business are looking exactly like dev.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-685-avcajisb.web.app

src/components/new-request/search.vue
severinbeauvais: You have very little ugliness in the template 👍 

I think the logic in the 3 global button getters is fine.

However, I think you still need to handle Societies.

JazzarKarim: Ahh, I missed that, my bad. There's the `isSocietyEnabled` FF. I'm on it.

ketaki-deodhar:
I think UI is not matching DEV:

In DEV:
image

Temp URL:
image

JazzarKarim:
> I think UI is not matching DEV.

Thanks for catching this Ketaki. On it.

src/components/new-request/search.vue
JazzarKarim: I am yet to fix the society dialog when FF is false. The code is written in a very weird way and I'm trying to figure that out.

src/components/new-request/search.vue
severinbeauvais: I "disabled" this getter in my previous PR (merged), so this conversation is obsolete.

src/components/new-request/search.vue
JazzarKarim: Never mind, it works totally fine right now. If the FF is disabled, the dialog will show. I tried that by removing the ! on line 255 in `pick-entity-or-conversion.vue` https://github.com/JazzarKarim/namerequest/blob/ce013f6d0ad827699da28c3a6374d2c7416bc0de/src/components/dialogs/pick-entity-or-conversion.vue#L255

So, societies should be OK after latest commit.

src/components/new-request/search.vue
```diff
@@ -14,17 +14,17 @@
       <!-- New BC Business flow -->
       <template v-if="isNewBcBusiness">
         <EntityType v-if="getLocation" />
-        <CompanyType v-if="getEntityTypeCd" />
-        <NumberedCompanyBullets v-if="isNumberedCompany" />
+        <CompanyType v-if="getEntityTypeCd && isNumberedEntityType" />
+        <NumberedCompanyBullets v-if="isNumberedCompany && isNumberedEntityType" />
 
-        <template v-if="isNamedCompany">
+        <template v-if="(isNamedCompany || !isNumberedEntityType) && entity_type_cd">
```
severinbeauvais, line 20: Does this block display for a society (with the FF enabled)?

JazzarKarim: What do you mean Sev?

severinbeauvais: With your latest code, start a new BC business with type = Society.

Do  you see a name to request, and the Check this Name button?

JazzarKarim: Yes.

If FF true:
if ff true

If FF false:
if ff false

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-685-avcajisb.web.app

src/components/new-request/search.vue
severinbeauvais: Oh, I see how you've done it...

Society is NOT set as a named company so, in my flow, I added the check for `isSociety`. In your flow, you check if it's an entity that cannot be numbered and that entity is set. I think that can be simplified (and more concise) from
```
<template v-if="(isNamedCompany || !isNumberedEntityType) && entity_type_cd">
```
to
```
<template v-if="isNamedCompany || isSociety">
```

leodube-aot:
I think it's missing the Check This Name button, at least for BENs.

Screenshot 2023-09-14 084245

src/components/new-request/search.vue
JazzarKarim: This doesn't seem to work Sev.

The reason is that if we're in an SP for example, the Name Input won't be shown. This is since it's neither a named company nor a society.

sp not working

JazzarKarim:
> I think it's missing the Check This Name button, at least for BENs.

Thanks for catching that Leo! Hopefully, it should be OK in latest temp URL.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-685-avcajisb.web.app

src/components/new-request/search.vue
severinbeauvais: Oh, I assumed SP was automatically set as a Named company (without showing the company type component since it can never be numbered).

In amalgamations, SPs don't exist, so I didn't see this issue.

Leave it as-is then.

severinbeauvais (approved):
LGTM

JazzarKarim:
@leodube-aot @ketaki-deodhar OK with you to merge now (I want to start with the restoration flow after rebasing from here)?

ketaki-deodhar (approved):

PR #688, opened 2023-09-13
17727 Refactored Conversion/Alteration Flow

leodube-aot:
*Issue #:* /bcgov/entity#17727

*Description of changes:*
- Completed refactoring of the conversion/alteration flow

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -95,7 +95,7 @@
       <template v-else-if="isConversion">
         <BusinessLookupFetch />
         <EntityType v-if="getSearchBusiness && isAlterable" cols="12" md="12" />
-        <CompanyType v-if="showCompanyTypeRadioButtons" />
+        <CompanyType v-if="!!getSearchBusiness && !!getConversionType" />
 
         <template v-if="isNamedCompany">
           <v-col cols="12" :md="showDesignation ? '8' : '12'">
@@ -104,7 +104,7 @@
               @emit-corp-num-validity="corpNumValid = $event"
             />
           </v-col>
-          <Designation cols="12" md="4" />
+          <Designation v-if="showDesignation" cols="12" md="4" />
         </template>
 
         <template v-if="isNumberedCompany">
@@ -285,12 +285,6 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
   }
 
   get showCompanyTypeRadioButtons (): boolean {
-    if (
-      this.isConversion &&
-      !!this.getSearchBusiness &&
-      !!this.getConversionType
-    ) return true
-
     // *** TODO: add your logic in the template instead of the spaghetti below
     // if (!this.getEntityTypeCd && !this.getConversionType && !this.isFederal) return false
     // if (this.isConversion) {
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -93,9 +93,23 @@
 
       <!-- Conversion (aka Alteration) flow -->
       <template v-else-if="isConversion">
-        <EntityType v-if="getSearchBusiness && isAlterable" />
         <BusinessLookupFetch />
+        <EntityType v-if="getSearchBusiness && isAlterable" cols="12" md="12" />
         <CompanyType v-if="showCompanyTypeRadioButtons" />
+
+        <template v-if="isNamedCompany">
+          <v-col cols="12" :md="showDesignation ? '8' : '12'">
+            <NameInput
+              :is-mras-search="(isXproFlow && isMrasJurisdiction && !getHasNoCorpNum)"
+              @emit-corp-num-validity="corpNumValid = $event"
+            />
+          </v-col>
+          <Designation cols="12" md="4" />
+        </template>
+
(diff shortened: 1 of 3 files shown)


src/components/new-request/search.vue
```diff
@@ -291,6 +311,12 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
       this.isSupportedAmalgamation(this.getEntityTypeCd)
     ) return true
 
+    if (
+      this.isConversion &&
+      this.isNumberedCompany &&
+      this.isSupportedAlteration(this.getConversionType)
+    ) return true
+
```
leodube-aot, line 319: @severinbeauvais Was this sort of how you pictured us implementing logic for each flow? A separate if statement as needed?

src/components/new-request/search.vue
```diff
@@ -93,9 +93,23 @@
 
       <!-- Conversion (aka Alteration) flow -->
       <template v-else-if="isConversion">
-        <EntityType v-if="getSearchBusiness && isAlterable" />
         <BusinessLookupFetch />
+        <EntityType v-if="getSearchBusiness && isAlterable" cols="12" md="12" />
         <CompanyType v-if="showCompanyTypeRadioButtons" />
+
+        <template v-if="isNamedCompany">
```
severinbeauvais, line 100: You probably need to add `|| isSociety` here.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-688-1g710hqa.web.app

src/components/new-request/search.vue
```diff
@@ -93,9 +93,23 @@
...
+          <v-col cols="12" :md="showDesignation ? '8' : '12'">
+            <NameInput
+              :is-mras-search="(isXproFlow && isMrasJurisdiction && !getHasNoCorpNum)"
+              @emit-corp-num-validity="corpNumValid = $event"
+            />
+          </v-col>
+          <Designation cols="12" md="4" />
```
severinbeauvais, line 107: You probably need to add `v-if="showDesignation"` here.

src/components/new-request/search.vue
severinbeauvais: Yup, exactly 👍

src/components/new-request/search.vue
```diff
@@ -271,6 +285,12 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
   }
 
   get showCompanyTypeRadioButtons (): boolean {
+    if (
+      this.isConversion &&
+      !!this.getSearchBusiness &&
+      !!this.getConversionType
+    ) return true
```
severinbeauvais, line 292: I think you can simply add the needed logic to line 98 and then we can eventually get rid of this getter.

severinbeauvais (commented):
Looks good. Do you need to support xpro or societies?

severinbeauvais:
Did you compare exactly with Dev Namerequest?

leodube-aot:
> Looks good. Do you need to support xpro or societies?

I'm not 100% sure. The UXPin doesn't state anything about those being part of allowable alterations but I'll double check.

> Did you compare exactly with Dev Namerequest?

I compared it using a BEN (BC0870877), a ULC (BC0871323), and a Limited Company (BC0870316) and it matched with dev

severinbeauvais:
> > Looks good. Do you need to support xpro or societies?
> 
> I'm not 100% sure. The UXPin doesn't state anything about those being part of allowable alterations but I'll double check.
> 
> Edit: On dev I tested with an Expro (A0002744) and it gives an error message stating the business cannot alter it's type, which matches with the refactor. Weirdly, dev will also display the Jurisdiction select field despite the error (which I think is likely incorrect), but that does not happen with the refactored version. Will need to try with a society.
> 
> > Did you compare exactly with Dev Namerequest?
> 
> I compared it using a BEN (BC0870877), a ULC (BC0871323), and a Limited Company (BC0870316) and it matched with dev

Actually, I remembered: societies can't alter their type,

I think you're right that that's a bug in Dev,

Looks good to me!

src/components/new-request/search.vue
severinbeauvais: Societies cannot change their type -- don't add this condition. Resolving.

src/components/new-request/search.vue
severinbeauvais: Clean! I love it!

src/components/new-request/search.vue
```diff
@@ -93,9 +93,23 @@
...
+              :is-mras-search="(isXproFlow && isMrasJurisdiction && !getHasNoCorpNum)"
+              @emit-corp-num-validity="corpNumValid = $event"
+            />
+          </v-col>
+          <Designation v-if="showDesignation" cols="12" md="4" />
+        </template>
+
+        <template v-if="isNumberedCompany">
+          <NumberedCompanyBullets />
+        </template>
```
severinbeauvais, line 112: I suggest moving the v-if into line 111 and getting rid of the template, but this isn't a deal-breaker.

severinbeauvais (approved):

PR #708, opened 2023-09-22
17843 Fix Register Now button

JazzarKarim:
*Issue #:* /bcgov/entity#17843

*Description of changes:*
- Fixed the register now button

If you want to test in temporary URL:
SP: NR 7348035, phone number: 1112223333
GP: NR 1346615, phone number: 1112223333
DBA: NR 7241339, phone number: 1112223333

### DIFF
--- a/src/mixins/nr-affiliation-mixin.ts
+++ b/src/mixins/nr-affiliation-mixin.ts
@@ -134,33 +134,36 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
 
   /** Returns business request object. */
   private getBusinessRequest (accountId: number, nr: NameRequestI): BusinessRequest {
-    let name = 'registration'
-    let legalType = nr.legalType
-    // If we're doing an incorporation (Incorporate Now button).
-    if (this.isSupportedIncorporationRegistration(nr.entity_type_cd) && !this.isFirm) {
-      name = 'incorporationApplication'
-      legalType = this.entityTypeToCorpType(nr.entity_type_cd)
-    }
+    let name = ''
+    let legalType = ''
+    let businessRequest = {} as BusinessRequest
     const nrNumber = nr.nrNum
 
-    const businessRequest = {
-      filing: {
-        business: { legalType },
-        header: { accountId, name }
-      }
-    } as BusinessRequest
-
     if (!this.isFirm(nr)) {
-      businessRequest.filing.incorporationApplication = {
-        nameRequest: { legalType, nrNumber }
-      }
-    }
-
-    if (this.isFirm(nr)) {
-      businessRequest.filing.registration = {
-        business: { natureOfBusiness: nr.natureOfBusiness },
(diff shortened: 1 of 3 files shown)


src/mixins/nr-affiliation-mixin.ts
```diff
@@ -136,7 +136,7 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
   private getBusinessRequest (accountId: number, nr: NameRequestI): BusinessRequest {
     let name = 'registration'
     let legalType = nr.legalType
-    if (this.isSupportedIncorporationRegistration(nr.entity_type_cd)) {
+    if (this.isSupportedIncorporationRegistration(nr.entity_type_cd) && !this.isFirm) {
```
JazzarKarim, line 139: The logic here was incorrect. Basically, we were always going into the if statement whether we're incorporating (BC, ULC etc...) or doing a registration (SPs, GPs, DBAs) since they're all in the FF.

Adding the `isFirm` prevents from going into the if statement when we're doing registration of SPs, GPs, or DBAs.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-708-0smuwtar.web.app

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-708-0smuwtar.web.app

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Ah, OK... so if this is a firm then it keeps the name and legalType from lines 137-138 and then successfully creates a draft registration filing?

Man, this entire method could be cleaned up... too much back and forth logic! But it is Friday.

severinbeauvais (approved):

src/mixins/nr-affiliation-mixin.ts
JazzarKarim: Yup, you're precisely right Sev.

I'll have a look. See if I can clean it up in a nice way 🙌

JazzarKarim: Sev, give me your thoughts on the code in the latest commit!

I've separated the logic into an if and else. That's easier to read and much cleaner I think.

Also, I noticed that the FF being used here is not needed. This is since we're controlling whether we're showing the button or not in the first place using the exact same FF.

severinbeauvais: Oh yeah! Wayyyy better! 🦜

severinbeauvais: I think you're right about the FF. Nice.

It was actually a bug before: if the FF was false then it would create a draft registration!

Looks fine to me.

JazzarKarim: Awesome, thanks Sev!

ketaki-deodhar (approved):
👍

PR #711, opened 2023-09-26
9392 capture edit-lock-error

eve-git:
*Issue #:* https://github.com/bcgov/entity/issues/9392
*Description of changes:*
capture error when edit/refund an NR with state other than 'DRAFT'

### DIFF
--- a/src/components/dialogs/error.vue
+++ b/src/components/dialogs/error.vue
@@ -5,15 +5,25 @@
         <div>Name Request encountered an error</div>
       </v-card-title>
 
-      <v-card-text class="copy-normal pt-8">
-        If you have paid for a new NR, please do not try submitting your payment again.
-      </v-card-text>
+      <div v-if="isEditLockError">
+        <v-card-text class="copy-normal pt-8">
+          The Name Request is presently undergoing examination or editing by another user
+          and cannot be modified or canceled at this time.
+          <br>
+          <br>
+          For assistance, please contact BC Registries staff:
+        </v-card-text>
+      </div>
+      <div v-else>
+        <v-card-text class="copy-normal pt-8">
+          If you have paid for a new NR, please do not try submitting your payment again.
+        </v-card-text>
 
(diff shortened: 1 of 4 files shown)


src/services/namex-services.ts
```diff
@@ -466,7 +471,13 @@ export default class NamexServices {
     } catch (err) {
       const msg = await this.handleApiError(err, 'Could not patch name requests by action')
       console.error('patchNameRequestsByAction() =', msg) // eslint-disable-line no-console
-      await errorModule.setAppError({ id: 'patch-name-requests-by-action-error', error: msg } as ErrorI)
+
+      let error_id = 'patch-name-requests-by-action-error'
+      if (err instanceof AxiosError && err.response.status === 423) {
+        // cannot be cancelled if the NR is not DRAFT state
```
severinbeauvais, line 477: I like this comment. Can you add it to the code block on line 213 also please?

src/components/dialogs/error.vue
```diff
@@ -48,6 +58,11 @@ export default class ErrorDialog extends Vue {
   async hideModal () {
     await ErrorModule.clearAppErrors()
   }
+
+  get isEditLockError () {
+    const errors = ErrorModule[ErrorTypes.GET_ERRORS]
+    return errors[0] && errors[0].id === 'edit-lock-error'
```
severinbeauvais, line 64: Can `errors` ever be null or undefined?

eve-git:
I don't think the previous word in Error dialog (for payment) works for the scenario. After discuss with Yui, I opened https://github.com/bcgov/entity/issues/17945 for UX to review and decide the error message.
The modifications in error.vue may be temporary, with the hope that they will suffice until UX provides further guidance.

src/components/dialogs/error.vue
eve-git: no, it is initialed as an empty array.

severinbeauvais (approved):

PR #725, opened 2023-10-05
18023 incorrect name in breadcrumb

jamespaologarcia:
*Issue #:* /bcgov/entity#18023

*Description of changes:*
Created a separate breadcrumb for logged in users

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -228,7 +228,6 @@ export default class App extends Mixins(
 
     // Set base crumbs based on user role
     // Staff don't want the home landing page and they can't access the Manage Business Dashboard
-    console.log(this.isAuthenticated)
     if (this.isAuthenticated) {
       if (this.isRoleStaff) {
         crumbs.unshift(getStaffDashboardBreadcrumb())
--- a/src/App.vue
+++ b/src/App.vue
@@ -112,7 +112,7 @@ import { BreadcrumbIF } from '@/interfaces'
 import {
   getRegistryDashboardBreadcrumb,
   getStaffDashboardBreadcrumb,
-  getUserDashboardBreadcrumb
+  getRegistryHomeBreadcrumb
 } from '@/resources'
 import axios from 'axios'
 
@@ -173,7 +173,7 @@ export default class App extends Mixins(
   @Getter getNrId!: number
   @Getter isRoleStaff!: boolean
   @Getter isMobile!: boolean
-  @Getter isAuthenticated!: boolean
+  @Getter isRoleBasic!: boolean
 
   // Global actions
   @Action resetAnalyzeName!: ActionBindingIF
@@ -227,14 +227,13 @@ export default class App extends Mixins(
 
     // Set base crumbs based on user role
     // Staff don't want the home landing page and they can't access the Manage Business Dashboard
-    if (this.isAuthenticated) {
-      if (this.isRoleStaff) {
-        crumbs.unshift(getStaffDashboardBreadcrumb())
-      } else {
-        crumbs.unshift(getUserDashboardBreadcrumb())
-      }
+
+    if (this.isRoleStaff) {
+      crumbs.unshift(getStaffDashboardBreadcrumb())
+    } else if (this.isRoleBasic) {
+      crumbs.unshift(getRegistryHomeBreadcrumb())
     } else {
-      // For non-staff, set Home crumb
+      // If not logged in, set Registry Home Breadcrumb
       crumbs.unshift(getRegistryDashboardBreadcrumb())
     }
     return crumbs
--- a/src/App.vue
+++ b/src/App.vue
@@ -173,7 +173,6 @@ export default class App extends Mixins(
   @Getter getNrId!: number
   @Getter isRoleStaff!: boolean
   @Getter isMobile!: boolean
-  @Getter getFolioNumber!: string
   @Getter isAuthenticated!: boolean
 
   // Global actions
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -225,9 +227,12 @@ export default class App extends Mixins(
...
     // Set base crumbs based on user role
     // Staff don't want the home landing page and they can't access the Manage Business Dashboard
-    if (this.isRoleStaff) {
-      // If staff, set StaffDashboard as home crumb
-      crumbs.unshift(getStaffDashboardBreadcrumb())
+    if (this.isAuthenticated) {
+      if (this.isRoleStaff) {
+        crumbs.unshift(getStaffDashboardBreadcrumb())
+      } else {
+        crumbs.unshift(getUserDashboardBreadcrumb())
```
severinbeauvais, line 234: I think you don't need to do this.

If you're not logged in as staff, the breadcrumb link takes you to this page, which is titled "BC Registries Dashboard". It's the same page as now, but the text is wrong.

src/App.vue
```diff
@@ -225,9 +227,12 @@ export default class App extends Mixins(
...
+      }
     } else {
       // For non-staff, set Home crumb
```
severinbeauvais, line 237: "For non-staff, or not logged in, set Registry Dashboard Breadcrumb"

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-725-eeac624z.web.app

janisrogers:
The changes requested in the ticket look great. I did however notice that the non-logged in user is taken to the login page, instead of the BCROS home page (https://dev.bcregistry.gov.bc.ca/) Can that be updated?

severinbeauvais:
> The changes requested in the ticket look great. I did however notice that the non-logged in user is taken to the login page, instead of the BCROS home page (https://dev.bcregistry.gov.bc.ca/) Can that be updated?

Is it correct to tell the user in the breadcrumb that they will go to the login page? After login, they go to the BCROS dashboard (same as logged in).

severinbeauvais:
> The changes requested in the ticket look great. I did however notice that the non-logged in user is taken to the login page, instead of the BCROS home page (https://dev.bcregistry.gov.bc.ca/) Can that be updated?

@janisrogers Is it correct to tell the user in the breadcrumb that they will go to the login page? After login, they go to the BCROS dashboard (same as logged in).

janisrogers:
There is a non logged in version of the BCROS dashboard. I don't think we should direct the user to the login page if they haven't chosen to login.

severinbeauvais:
> There is a non logged in version of the BCROS ~dashboard~ Main page. This is where the breadcrumb should take them. I don't think we should direct the user to the login page if they haven't chosen to login.

That's a good point / idea. Janis, let us know whether to proceed with that.

James, when Janis has approved this requirement, the non-logged-in breadcrumb should go to `registryHomeUrl` and its text should be "BC Registries and Online Services". The logged-in breadcrumb should go to `${registryHomeUrl}dashboard/${getParams()}` and its text should be "BC Registries Dashboard".

janisrogers:
Yes, please proceed with that

src/App.vue
severinbeauvais: Please update comment. And please rename the variable to `getRegistryHomeBreadcrumb`.

src/resources/breadcrumb-resources.ts
```diff
@@ -10,10 +10,16 @@ export function getRegistryDashboardBreadcrumb (): BreadcrumbIF {
...
+  }
+}
+export function getUserDashboardBreadcrumb (): BreadcrumbIF {
+  const registryHomeUrl = sessionStorage.getItem('REGISTRY_HOME_URL')
+  return {
+    text: 'BC Registries Dashboard',
     href: `${registryHomeUrl}dashboard/${getParams()}`
   }
 }
-
```
severinbeauvais, line 16: Please restore blank line to keep this style consistent.

severinbeauvais: Also add blank line above line 16.

src/App.vue
severinbeauvais: The requirement changed. This conversation is obsolete.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-725-eeac624z.web.app

severinbeauvais:
James, there might be a problem with how the code determines whether or not the user is authenticated.

I tested this change in the preview build above (temp URL), and it looks fine for logged out, but after logged in, the breadcrumb is incorrect until I refresh the page again. Try it yourself. (I have an idea how to fix it.)

severinbeauvais:
> I have an idea how to fix it.

I think the error is that `isAuthenticated` checks for the presence of the keycloak token before it's populated. Because it's a getter, its value is cached until something it depends on changes (ie, responsive), but it doesn't recognize changes to the session storage so it keeps its original False value.

Ref: https://stackoverflow.com/questions/69789213/any-way-to-make-sessionstorage-reactive-in-vue3

severinbeauvais (approved):
This is good enough for now but I might pick away at the code a bit tomorrow to clean something up.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-725-eeac624z.web.app

severinbeauvais:
> This is good enough for now but I might pick away at the code a bit tomorrow to clean something up.

There was actually a bug in this (wrong getter), but I want to keep THIS change separate from the login detection change (*), so I fixed the bug using `isAuthenticated` (as part of THIS ticket) and created a new ticket for the login detection.

(*) I want us to fix the login detection properly, which is out of scope for THIS ticket.

PR #702, opened 2023-09-18
Updated package.json to create a dummy PR

severinbeauvais:
Do not commit this change. It is only to create a dummy PR so that we can build a previous tempory version of this code to facilitate testing.

PR #728, opened 2023-10-11
Updated package.json to create a dummy PR

severinbeauvais:
Do not commit this change. It is only to create a dummy PR so that we can build a previous tempory version of this code to facilitate testing.


severinbeauvais:
This PR didn't build the preview site (temporary URL) so I'm closing it and will try to reuse #702.

PR #735, opened 2023-10-23
show both named and numbered option for LTD company

tshyun24:
*Issue #:* /bcgov/entity#18202

*Description of changes:* Using `isNumberedCompany` to determine whether we have the COLIN option, instead of using `isSupportedRestoration` previously 


bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-735-526soleu.web.app

severinbeauvais:
Please rebase and update the package files.

Please provide some sample businesses that:
- are and are not supported by FF
- can be only named and can be named/numbered

src/components/new-request/search.vue
```diff
@@ -291,7 +291,7 @@
             :href="colinLink"
             target="_blank"
           >
-            Go to Corporate Online to {{ isConversion ? 'Alter' : 'Register' }}
+            Go to Corporate Online to {{ isConversion ? 'Alter' : isRestoration ? 'Restore' : 'Register'}}
```
severinbeauvais, line 294: Please change this to a computed (get) and also include the "Change Name" option.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-735-526soleu.web.app

tshyun24:
Some sample business
image

- Supported by FF : BC0013141 (FF value: UL), BC0871194 (FF value: BC)
- Not supported by FF: CP1002580 (FF value: CP)
- Can only be named: A0076494 (Xpro business)
- Can be named/numbered: BC0024185

JazzarKarim:
> * Supported by FF : BC0013141 (FF value: UL), BC0871194 (FF value: BC)
> * Not supported by FF: CP1002580 (FF value: CP)
> * Can only be named: A0076494 (Xpro business)
> * Can be named/numbered: BC0024185

Shaoyun, for coops, we don't want the numbered and named radio buttons.
In DEV:
coop in dev
Unless something has changed, this is the expected behavior for coops. They're like a special case. I was having a look in your temp URL, I can see that we got the radio button options.

Also, for "A0076494", I got stuck:
A business shaoyun

Please have a look in DEV and see the expected result. There's the "Business's full legal name in home jurisdiction" component missing.

src/components/new-request/search.vue
severinbeauvais: Fixed, thanks.

JazzarKarim:
> @JazzarKarim @severinbeauvais push a new comment, hopefully fix all the bugs above

So I've tested the different workflows, looks great! Shaoyun, can you please take a SC from your local when the `supported-restoration-entities` FF is false and when a ULC for example is selected (just like how we discussed yesterday)?

src/components/new-request/search.vue
```diff
@@ -240,7 +240,7 @@
       <!-- Restoration / Reinstatement flow -->
       <template v-else-if="isRestoration">
         <BusinessLookupFetch />
-        <CompanyType v-if="getSearchBusiness && isBcRestorable && isSupportedRestoration(getEntityTypeCd)" />
+        <CompanyType v-if="getSearchBusiness && isBcRestorable && isNumberedEntityType" />
```
severinbeauvais, line 243: I am wondering why the statement `isBcRestorable` is here... if the company is not restorable (due to its type) then CompanyType is not rendered, but what about the other components below?

1. Please test a non-restorable business type and see what components you see.

2. You may need to put some of these elements in a block with "v-if=isBcRestorable".

severinbeauvais: This looks correct:

severinbeauvais: This XP looks correct:

image

severinbeauvais: I guess this code is OK. Resolving this conversation.

tshyun24:
> So I've tested different types of companies, looks great! Shaoyun, can you please take a SC from your local when the `supported-restoration-entities` FF is false and when a ULC for example is selected (just like how we discussed yesterday)?

Yes, I just create a ULC business (BC0871496) and make it historical
image
It looks like that in DEV. I think I should not have the Restore Now option? Cause it doesn't include the FF for ULC
image
And the below is in local, same with DEV
image

JazzarKarim:
> > So I've tested different types of companies, looks great! Shaoyun, can you please take a SC from your local when the `supported-restoration-entities` FF is false and when a ULC for example is selected (just like how we discussed yesterday)?
> 
> Yes, I just create a ULC business (BC0871496) and make it historical

What if the FF value is false? Will it show the same thing but instead of "Restore Now", it's "Go To Corporate Online to Register"?

src/components/new-request/search.vue
```diff
@@ -250,7 +250,7 @@
         <!-- federal sub-flow -->
         <XproFederalBullets v-if="isFederal && getSearchBusiness" />
 
-        <template v-if="(isRestorable && !isFederal) && (isNamedCompany || !isSupportedRestoration(getEntityTypeCd))">
+        <template v-if="(isRestorable && !isFederal) && (isNamedCompany || isCooperative || isCreditUnion || isSelectedCompanyXPro)">
```
tshyun24, line 253: add it to make sure Extrapos, Coop, and Credit Union can enter their name
image
image
image

tshyun24:
> What if the FF value is false? Will it show the same thing but instead of "Restore Now", it's "Go To Corporate Online to Register"?
> 
> You can set this line to return false to test that! https://github.com/tshyun24/namerequest/blob/f495ad9e87a33ba472479575c3edab334510b7e8/src/mixins/common-mixin.ts#L209

Show  "Go To Corporate Online to Register" when I set FF as false
image

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-735-526soleu.web.app

src/components/new-request/search.vue
severinbeauvais: > add it to make sure Extrapos, Coop, and Credit Union can enter their name

The existing getter `isNamedCompany` should be True for these already, so you shouldn't need to check for them explicitly.

In fact, you shouldn't check for entity types here at all. The purpose of the lists in request-action-mappings.ts is to declare which types are valid for which filings. The getter that uses those lists is `isRestorable`, which is already checked in your if statement here. So, the logic here needs to be cleaned up.

Finally, if you see a lot of logic like this in the template then it should be extracted into a local getter (computed), so that the template is as clean as possible.

JazzarKarim:
> > What if the FF value is false? Will it show the same thing but instead of "Restore Now", it's "Go To Corporate Online to Register"?
> > You can set this line to return false to test that! https://github.com/tshyun24/namerequest/blob/f495ad9e87a33ba472479575c3edab334510b7e8/src/mixins/common-mixin.ts#L209
> 
> Show "Go To Corporate Online to Register" when I set FF as false image

Looks great! This is basically what we want.

src/components/new-request/search.vue
tshyun24: > > add it to make sure Extrapos, Coop, and Credit Union can enter their name
> 
> The existing getter `isNamedCompany` should be True for these already, so you shouldn't need to check for them explicitly.
> 
> In fact, you shouldn't check for entity types here at all. The purpose of the lists in request-action-mappings.ts is to declare which types are valid for which filings. The getter that uses those lists is `isRestorable`, which is already checked in your if statement here. So, the logic here needs to be cleaned up.
> 
> Finally, if you see a lot of logic like this in the template then it should be extracted into a local getter (computed), so that the template is as clean as possible.

I find that if I only keep `isNamedCompany`, it does not work for Coop and Xpro, no name input
image

tshyun24: I notice that if I only keep `isNamedCompany`, it doesn't work for Coop and XPro as no place for name input
image

So maybe I should great a getter to nest all the logic at here?

severinbeauvais: Re: `(isRestorable && !isFederal) && (isNamedCompany || isCooperative || isCreditUnion || isSelectedCompanyXPro)`

I worked through this logic and I think you are correct. The issue is that CP and FI are special cases -- they are handled differently from the top and bottom group here:

image

So, your code needs to allow them, as you've done 👍 

However, I am not sure about `isSelectedCompanyXPro`. That getter uses `isSelectedXproAndRestorable`, which is already used by `isRestorable`, so it feels like you're checking the same thing twice. Can you explain why you need it? An example would be great.

tshyun24: I keep `isSelectedCompanyXPro` because the XPro company should be in different logic, and I didn't find a function that can only check whether it's XPro and avoid `isRestorable`. So should I create a new getter that check XPro only or ?

If I remove `isSelectedCompanyXPro`, it will show, no name input here
image

severinbeauvais: All right. The logic is complicated because of the different types of checks. Let's move this to a local getter (computed) where you can explicitly label the different conditions and explain things for future developers.

See following example. This is essentially what you have but with comments. Also I'm using `isSelectedXproAndRestorable` instead of `isSelectedCompanyXPro` (and I think you can delete `isSelectedCompanyXPro` as it's not used for anything else).

```
const isPromptNameInput = (
  this.isNamedCompany || // from radio buttons
  this.isCooperative || // specially include CP
  this.isCreditUnion || // specially include FI
  this.isSelectedXproAndRestorable // include all xpros
)

return (
  this.isRestorable && // check mapping lists for REH
  isPromptNameInput &&
  !this.isFederal // ignore Federal xpros (special case)
)
```

severinbeauvais: Shaoyun, in the end I think you're correct, but I had to wrap my head around it. It really seemed like some of these checks were unnecessary or redundant, but they're checking either types or different parts of the flow. It's complicated. In these cases, I like to see explicit code, comments, etc. You can rewrite the above however you see fit. Thanks for your patience with me :)

tshyun24: Sure, Severin, I will create a local getter. Yes, there are lots of complicated logic so I just push lots of comments to fix the errors. And thank you so much for those reviews<

tshyun24: > const isPromptNameInput = (
>   this.isNamedCompany || // from radio buttons
>   this.isCooperative || // specially include CP
>   this.isCreditUnion || // specially include FI
>   this.isSelectedXproAndRestorable // include all xpros
> )
> 
> return (
>   this.isRestorable && // check mapping lists for REH
>   isPromptNameInput &&
>   !this.isFederal // ignore Federal xpros (special case)
> )

Hi Severin, should I keep those comments in my getter function or just leave one overall comment for the getter?

severinbeauvais: I don't like to tell smart developers how to code, so up to you where the comments go or exactly what they say but I think they're all valuable. (They are the things I had to read into and think about.)

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-735-526soleu.web.app

src/components/new-request/search.vue
```diff
@@ -459,10 +459,22 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
...
   get isSociety (): boolean {
     return (this.isSocietyEnabled() && this.getEntityTypeCd === EntityTypes.SO)
   }
 
+  /** Whether to show the name input field when trying to restore a historical company. */
+  get showRestoreNameInput (): boolean {
+    const isPromptNameInput = (this.isNamedCompany || this.isCooperative || this.isCreditUnion || this.isSelectedXproAndRestorable)
+    return (
+      this.isRestorable && isPromptNameInput && !this.isFederal
+    )
```
severinbeauvais, line 475: The code is fine but is missing the detail that the comments provided, such as "why are we checking these specific variables?". That's what got me in this rabbit hole initially. And there's nothing wrong with the checks but it's not clear that they are different types of checks -- for example, not just entity type check, but also flow checks. Could you include the detailed comments please?

severinbeauvais: In your original conversation, you described (and included a uxpin screenshot) what this check is, and that business logic needs to be described here or a future dev will not understand this code.

src/components/new-request/search.vue
severinbeauvais: Resolving this conversation as it is now continued elsewhere.

src/components/new-request/search.vue
tshyun24: Yes... I should leave more comments here, I also get confused when I first tryin to fix it.
I typed the following comment under `showRestoreNameInput`, not sure is it cleared enough and my poor grammar...
`// In the Uxpin, when company's type is one of coop, credit union, and XPro, it is in different work flow with regular named company.`
`// Thus we need to consider them seprately to determine whether we should have the name input field.   `

severinbeauvais: That's very good, but future developers won't have access to the uxpin, so you'll have to describe the groups of companies that are included here.

1. named when radio button was used
2. specially include CP
3. specially include FI
4. include all xpros

severinbeauvais: That looks great! Commit it and then check with Karim if this PR is OK to merge.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-735-526soleu.web.app

src/components/new-request/search.vue
```diff
@@ -563,6 +578,20 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
...
+    var colinText = ''
+    if (this.isConversion) {
+      colinText = 'Alter'
+    } else if (this.isRestoration) {
+      colinText = 'Restore'
+    } else if (this.isChangeName) {
+      colinText = 'Change Name'
+    } else {
+      colinText = 'Register'
+    } return 'Go to Corporate Online to ' + colinText
```
severinbeauvais, line 592: This return statement should be on its own line.

src/components/new-request/search.vue
```diff
@@ -563,6 +578,20 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
     return false
   }
 
+  /** Retrieve text for COLIN button based on selected action/flow */
+  get colinButtonText (): string {
+    var colinText = ''
```
severinbeauvais, line 583: If this getter is used when this is not a conversion/restoration/change-name then it will return an incomplete string. I recommend initializing this string to "use this NR" as a safe fallback value.

severinbeauvais (approved):

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-735-526soleu.web.app

JazzarKarim (approved):
I just double checked that the different types look good. I also reached out to Shaoyun for confirmation on a couple of stuff.

Looks great! Awesome job 👍

PR #740, opened 2023-10-26
18068 - Namerequest UI: Editing the Resubmit NR has the first name choice as empty.

Rajandeep98:
*Issue #18068 https://github.com/bcgov/entity/issues/18068

*Description of changes:*
Updated condition for applying designation for resubmission for SP/GP

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -489,14 +489,18 @@ export const resubmit = async ({ commit, getters }): Promise<boolean> => {
 
   // add resubmit NR number (for internal use only - API ignores it)
   nrTypeData['resubmitNrNum'] = nrData['nrNum']
-
+  // check against entity types if it should have designation or not : if length returns to be 0 then NR should not have any designation otherwise work as getNameDesignation
+  let flag = false
+  if (Designations[nrData['entity_type_cd']].words.length === 0) {
+    flag = true
+  }
   // clean names data
   nrTypeData['names'] = nrData.names.map(name => ({
     choice: name.choice,
     consent_words: '',
     conflict1: '',
     conflict1_num: '',
-    designation: getNameDesignation(name),
+    designation: !flag ? getNameDesignation(name) : '',
     name: name.name,
     name_type_cd: name.name_type_cd
   }))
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-740-qail5x3t.web.app

severinbeauvais (approved):
LGTM

severinbeauvais (changes requested):
Removing my approval -- I didn't looks closely enough at the code. The code change will always be True because it's not comparing the NR's type with an enum; it's looking at the truthiness of the enums directly, which is always True.

Rajandeep and I looked at this together and we're not sure why this code exists -- can't the designation from the original NR simply be copied in to the new NR? This code suggests that some  NRs have a designation in the name but the designation field is empty. It would be good to have an example. We could then handle those specially.

The other option of exempting some types that we know don't have designations is less optimal but is probably safe and could be implemented as a last resort.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-740-qail5x3t.web.app

src/store/actions.ts
```diff
@@ -489,14 +489,18 @@ export const resubmit = async ({ commit, getters }): Promise<boolean> => {
 
   // add resubmit NR number (for internal use only - API ignores it)
   nrTypeData['resubmitNrNum'] = nrData['nrNum']
-
+  // check against entity types if it should have designation or not : if length returns to be 0 then NR should not have any designation otherwise work as getNameDesignation
+  let flag = false
+  if (Designations[nrData['entity_type_cd']].words.length === 0) {
+    flag = true
+  }
```
severinbeauvais, line 496: I'd name the flag something useful.

Also, you can simplify this...
```
const flag = (your boolean statement)
```

severinbeauvais: Thanks for the very useful comment!

severinbeauvais (approved):
I approve this solution.

eve-git (approved):

severinbeauvais:
Do not merge this until after the release on Tuesday, October 31.

src/store/actions.ts
severinbeauvais: Thanks for the update. Sorry for being so picky. Your code looks good and I've approved this PR but I think you have to wait a couple of days before merging it (check with Omid).

ozamani9gh (approved):
looks good

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-740-qail5x3t.web.app

PR #741, opened 2023-10-26
17914 added popup for socieites

jamespaologarcia:
*Issue #:* /bcgov/entity#17914

Added a popup for SO and XSO for name change and restoration

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "name-request",
-      "version": "5.3
+      "version": "5.3.5",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/breadcrumb": "2.1.24",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "5.3.4",
+  "version": "5.3.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "5.3.4",
+      "version": "5.3.5",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/breadcrumb": "2.1.24",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "name-request",
-  "version": "5.3.3",
+  "version": "5.3.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "name-request",
-      "version": "5.3.3",
+      "version": "5.3.4",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/breadcrumb": "2.1.24",
--- a/src/components/dialogs/societies-info-dialog.vue
+++ b/src/components/dialogs/societies-info-dialog.vue
@@ -44,7 +44,7 @@ import { ActionBindingIF } from '@/interfaces/store-interfaces'
 
 @Component({})
 export default class SocietiesInfo extends Vue {
-  @Prop() readonly type: string
+  @Prop({ required: true }) readonly type: string
   @Getter getSocietiesModalVisible!: boolean
   @Action setSocietiesModalVisible!: ActionBindingIF
 
--- a/src/components/dialogs/societies-info-dialog.vue
+++ b/src/components/dialogs/societies-info-dialog.vue
@@ -20,7 +20,7 @@
           <v-container fluid>
             <v-row no-gutters class="text-center">
               <v-col cols="12">
-                To request a {{ type }} for a Society
+                To {{ type }} a Society
               </v-col>
               <v-col cols="12">
                 please use the Societies Online website
@@ -44,7 +44,7 @@ import { ActionBindingIF } from '@/interfaces/store-interfaces'
 
 @Component({})
 export default class SocietiesInfo extends Vue {
-  @Prop() readonly type: boolean
+  @Prop() readonly type: string
   @Getter getSocietiesModalVisible!: boolean
   @Action setSocietiesModalVisible!: ActionBindingIF
 
@@ -57,7 +57,6 @@ export default class SocietiesInfo extends Vue {
 
   emitClose (): void {
     this.showModal = false
-    this.$parent.closeSocietyDialog()
   }
 }
 </script>
--- a/src/components/dialogs/societies-info-dialog.vue
+++ b/src/components/dialogs/societies-info-dialog.vue
@@ -1,0 +1,63 @@
+<template>
+    <v-dialog v-model="showModal" width="45rem" persistent content-class="custom-dialog">
+      <v-card>
+        <v-row>
+          <v-col cols="10">
+            <span class="copy-small"></span>
+          </v-col>
+          <v-col cols="2">
+            <v-icon
+            class="float-right"
+            md
+            color="primary"
+            @click="emitClose()"
+          >
+          mdi-close
+          </v-icon>
+          </v-col>
+        </v-row>
+        <v-card-text class="mt-n12">
+          <v-container fluid>
+            <v-row no-gutters class="text-center">
+              <v-col cols="12">
+                To request a {{ type }} for a Society
+              </v-col>
+              <v-col cols="12">
+                please use the Societies Online website
+              </v-col>
+              <v-col cols="12">
+                <a href="https://www.bcregistry.ca/societies/">https://www.bcregistry.ca/societies/</a>
+              </v-col>
+            </v-row>
+          </v-container>
+        </v-card-text>
+        <v-card-actions class="pt-0">
+          <v-spacer />
+        </v-card-actions>
+      </v-card>
+    </v-dialog>
+  </template>
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator'
+import { Action, Getter } from 'vuex-class'
+import { ActionBindingIF } from '@/interfaces/store-interfaces'
+
+@Component({})
+export default class SocietiesInfo extends Vue {
+  @Prop() readonly type: boolean
+  @Getter getSocietiesModalVisible!: boolean
+  @Action setSocietiesModalVisible!: ActionBindingIF
+
+  get showModal () {
+    return this.getSocietiesModalVisible
+  }
+  set showModal (value: boolean) {
+    this.setSocietiesModalVisible(value)
+  }
+
+  emitClose (): void {
+    this.showModal = false
+    this.$parent.closeSocietyDialog()
+  }
+}
+</script>
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -107,7 +107,7 @@
           :type="'request a name for'"
           :showDialog="showSocietiesInfo"
         />
-        <template v-if="isNameChangeable && !isSocietyDisabled">
+        <template v-else-if="isNameChangeable">
           <!-- XPRO jurisdiction -->
           <Jurisdiction
             v-if="isChangeNameXpro"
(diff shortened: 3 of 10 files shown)


bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

src/components/new-request/search.vue
```diff
@@ -103,11 +103,44 @@
       <!-- Change Name flow -->
       <template v-else-if="isChangeName">
         <BusinessLookupFetch />
-
+        <v-dialog  v-if="isSocietyDisabled" v-model="showSocietiesInfo"
```
severinbeauvais, line 106: Please move this to its own component, same as the other dialogs in this project.

severinbeauvais, line 106: I think you don't need the v-if here, because you are already controlling the visibility of this dialog with the v-model property.

src/components/new-request/search.vue
```diff
@@ -103,11 +103,44 @@
...
+                  </v-container>
+                </v-card-text>
+              </template>
+          </v-card>
+        </v-dialog>
         <template v-if="isNameChangeable">
           <!-- XPRO jurisdiction -->
           <Jurisdiction
-            v-if="isChangeNameXpro"
+            v-if="isChangeNameXpro && !isSocietyDisabled"
```
severinbeauvais, line 143: Would it make more sense to move this check to the entire template block here (line 140)?

src/components/new-request/search.vue
```diff
@@ -428,6 +497,26 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
...
+  openSocietyDialog () {
+    this.showSocietiesInfo = true
+  }
+
+  closeSocietyDialog () {
+    this.showSocietiesInfo = false
+    this.setSearchBusiness(null)
+  }
+
+  get maxWidth (): string {
```
severinbeauvais, line 509: You don't need this getter because your dialog is a fixed width.

In pick-entity-or-conversion.vue, the dialog could be really wide, ie

image

or narrower:

image

severinbeauvais: Looks like a static value of 550px is correct for your dialog.

image

severinbeauvais:
I have tested this with the "enable-society" FF off and the dialogs display correctly if I select a society for name change or restoration. Alteration is already protected by the mappings lists but I think it should be handled the same way as name change and restoration.

I tested this with the "enable-society" FF on and the name change flow works fine; the alteration flow is blocked by the mapping and should be fixed; the restoration flow works for extra societies (XS) but not BC societies (S) and should be fixed.

severinbeauvais:
Do not merge this until after the release on Tuesday, October 31.

src/components/new-request/search.vue
```diff
@@ -274,40 +242,7 @@
...
-                      <v-col cols="12">
-                        <a href="https://www.bcregistry.ca/societies/">https://www.bcregistry.ca/societies/</a>
-                      </v-col>
-                    </v-row>
-                  </v-container>
-                </v-card-text>
-              </template>
-          </v-card>
-        </v-dialog>
+        <SocietiesInfo type='restoration' />
```
severinbeauvais, line 245: Please check with UI/UX team (eg, Janis or Yui) on whether they want to see "To request a restoration for a Society".

src/interfaces/new-request-interface.ts
```diff
@@ -67,6 +67,7 @@ export interface NewRequestIF {
     request: RequestActionsI
   }
   showActualInput: boolean
+  societiesModalVisible: boolean
```
severinbeauvais, line 70: I had suggested just passing a prop to the dialog to show/hide it as that would be cleanest.

It's OK that you figured out how the other modals in Namerequest UI are show/hidden.

src/components/new-request/search.vue
```diff
@@ -103,40 +102,9 @@
...
-                        <a href="https://www.bcregistry.ca/societies/">https://www.bcregistry.ca/societies/</a>
-                      </v-col>
-                    </v-row>
-                  </v-container>
-                </v-card-text>
-              </template>
-          </v-card>
-        </v-dialog>
+        {{ isSocietyDisabled }}{{ showSocietiesInfo }}
+        <!--  :show-modal="isSocietyDisabled && showSocietiesInfo"-->
```
severinbeauvais, line 106: Remove this debugging code as some point in this PR.

src/components/new-request/search.vue
severinbeauvais: Thanks.

severinbeauvais: Fixed, thanks.

src/components/new-request/search.vue
severinbeauvais: Fixed, thanks.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

severinbeauvais:
I restarted all checks that failed due to environmental issue. They have now passed.

This flow still doesn't work. I should see more controls after this:

image

src/components/new-request/search.vue
severinbeauvais: I have unresolved this conversation because you left in the `madWidth` getter that you had initially added and is no longer used. Please clean this up.

src/components/new-request/search.vue
```diff
@@ -467,6 +499,19 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
...
+    let isSocietyDisabled = (
+      !this.isSocietyEnabled() &&
+      (this.getEntityTypeCd === EntityTypes.SO || this.getEntityTypeCd === EntityTypes.XSO)
+    )
+
+    this.setSocietiesModalVisible(isSocietyDisabled)
+
+    return isSocietyDisabled
+  }
+
```
severinbeauvais, line 514: I'm quite confused by this getter (computed). It should not be doing multiple things (and esp not setting state when you "get" something).

I think you go back to my earlier suggestion of passing a prop to show/hide the dialog, and using an event handler when it's closed to clean up the variables.

After that, if you still need to call something when a variable changes then use a watcher method.

severinbeauvais (changes requested):
Not OK as is. Please rework. Ask for more help or examples if needed.

PS You're close, and Namerequest UI is _frustratingly_ complicated, I know, so I harbour no ill feelings :)

eve-git:
Could you please update the version number to 5.3.5, as https://github.com/bcgov/namerequest/pull/742 is currently using 5.3.4 and merged already.

src/components/dialogs/societies-info-dialog.vue
```diff
@@ -44,7 +44,7 @@ import { ActionBindingIF } from '@/interfaces/store-interfaces'
 
 @Component({})
 export default class SocietiesInfo extends Vue {
-  @Prop() readonly type: boolean
+  @Prop() readonly type: string
```
severinbeauvais, line 47: It's good practise to include a default value, or to make this prop required.

src/components/new-request/search.vue
```diff
@@ -102,9 +102,7 @@
       <!-- Change Name flow -->
       <template v-else-if="isChangeName">
         <BusinessLookupFetch />
-        {{ isSocietyDisabled }}{{ showSocietiesInfo }}
-        <!--  :show-modal="isSocietyDisabled && showSocietiesInfo"-->
-        <SocietiesInfo type='name' />
+        <SocietiesInfo v-if="isSocietyDisabled && showSocietiesInfo" :type="'request a name for'" :showDialog="showSocietiesInfo"/>
```
severinbeauvais, line 105: Looks good!

Please put each attribute on a different line. I don't know if that lint rule is used in this repo but we try to do that in the other UIs.

src/components/new-request/search.vue
```diff
@@ -438,15 +436,20 @@ export default class Search extends Mixins(CommonMixin, NrAffiliationMixin, Sear
...
 
-  openSocietyDialog () {
-    this.showSocietiesInfo = true
+  @Watch('isSocietyDisabled')
+  onIsSocietyDisabledChanged (newVal: boolean, oldVal: boolean) {
+    this.setSocietiesModalVisible(newVal)
+    this.showSocietiesInfo = newVal
+    if (newVal) {
+      this.setSearchBusiness(null)
+    }
```
severinbeauvais, line 445: Looks good!

You can leave out `oldVal` if you don't need it.

I'd rename `newVal` to something more meaningful (like "disabled"?).

src/components/new-request/search.vue
severinbeauvais, line 105: You were asking about best practises for modals...

In this repo, all dialogs (except this one) are declared in App.vue and each one uses its own store getter + action to show/hide. Personally I don't like this approach because a lot of code files are involved; it's not "contained" or localized.

In the other repos, each dialog generally looks like this in the parent component:
```
<MyDialog
  :dialog="whetherToShowThisDialog"
  @close="closeThisDialog()"
/>

<script>
  ...
  whetherToShowThisDialog = false // initial value
  ...
  showThisDialog (): void {
    this.whetherToShowThisDialog = true
  }
  ...
  closeThisDialog (): void {
    this.whetherToShowThisDialog = false
  }
  ...
```

I prefer this approach because I don't have to touch any other system files (App.vue, interfaces, store, state, getters, actions). It's all here. And the parent can respond to close events (or you  could implement "exit" or "retry" events instead).

What do you think? There isn't a correct answer but depending on the app and what you're doing in it, there may be a "better" answer :)

severinbeauvais: Example: https://github.com/bcgov/business-filings-ui/blob/05b5e3b9950dd5d2f598c66d008394eca99466a5/src/views/AgmLocationChg.vue#L8

src/components/new-request/search.vue
severinbeauvais: What did they say about the messages?

src/components/new-request/search.vue
severinbeauvais: Fixed, thanks.

src/components/new-request/search.vue
severinbeauvais: Changed, thanks.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

src/components/new-request/search.vue
jamespaologarcia: You're right. Also saves time.

src/components/dialogs/societies-info-dialog.vue
severinbeauvais: Fixed, thanks.

src/components/new-request/search.vue
severinbeauvais: Fixed, thanks!

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

severinbeauvais:
This flow still doesn't work -- should show other input components after I select this society.

image

src/components/new-request/search.vue
jamespaologarcia: Checked with Yui Sotozaki. Made the changes accordingly.

Should be "To restore a Society..."

src/components/new-request/search.vue
severinbeauvais: I think this will work itself out when the final flow is fixed. Resolving.

src/components/new-request/search.vue
severinbeauvais: Thanks. Resolving.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

severinbeauvais:
> This flow still doesn't work -- should show other input components after I select this society.

Still doesn't work after latest changes.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

package-lock.json
severinbeauvais, line 1: I'm not sure what happened here. The usual procedure is:
1. update app version in package.json
2. enter `npm i`
3. commit package.json and package-lock.json

jamespaologarcia: It was a typho on my part. Will fix in the next push

severinbeauvais: Resolving.

src/components/new-request/search.vue
severinbeauvais: I'm resolving this conversation but it might be good reading for the other reviewers.

src/components/new-request/search.vue
severinbeauvais: Thanks for the changes.

PS Instead of specifying a prop as dynamic, which required you to pass the string as an expression, you could do this.

Before:
image

After:
```
  type="request a name for"
```

This works because the string is static.

There's a similar thing when passing boolean props. You can do this (note: no quotes):
```
  myProp=true
```
or this (dynamic prop so what is in quotes is evaluated):
```
  :myProp="true"
```

severinbeauvais: I'm resolving this conversation because your existing code is OK.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

JazzarKarim:
> This flow still doesn't work -- should show other input components after I select this society.
> 
> image

Sev, I've tried this in DEV, it's the same thing:
dev society

I wonder, what do we exactly need to show in that case? For XS, we show the usual jurisdiction and full legal name like you mentioned (if FF is true). In the case of a regular society though, we didn't have previous behavior. The UXPin also didn't have any mention for that.

severinbeauvais:
> Sev, I've tried this in DEV, it's the same thing:
> 
> I wonder, what do we exactly need to show in that case? For XS, we show the usual jurisdiction and full legal name like you mentioned (if FF is true). In the case of a regular society though, we didn't have previous behavior. The UXPin also didn't have any mention for that.

I've been communicating with James on this. It's partly a FF issue and partly a logic issue. See:
https://github.com/bcgov/namerequest/pull/743/commits/4cea5ca5fb9866e744e33cbb3228e154d0cc55a5?diff=split&w=1

severinbeauvais (approved):
Looks good. Do all your tests pass?

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-741-l6xc5zp1.web.app

JazzarKarim (approved):
LGTM!

PR #743, opened 2023-10-31
DO NOT MERGE - misc fixes

severinbeauvais:
Not applicable

### DIFF
--- a/src/components/new-request/search.vue
+++ b/src/components/new-request/search.vue
@@ -103,11 +103,44 @@
       <!-- Change Name flow -->
       <template v-else-if="isChangeName">
         <BusinessLookupFetch />
-
+        <v-dialog  v-if="isSocietyDisabled" v-model="showSocietiesInfo"
+          :max-width="maxWidth"
+          persistent>
+          <v-card>
+              <v-icon
+                class="float-right"
+                md
+                color="primary"
+                @click="closeSocietyDialog()"
+              >
+              mdi-close
+              </v-icon>
+              <template>
+                <v-card-text class="mt-n12">
+                  <v-container fluid>
+                    <v-row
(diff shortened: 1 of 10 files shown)


src/components/new-request/search.vue
```diff
@@ -107,7 +107,7 @@
           :type="'request a name for'"
           :showDialog="showSocietiesInfo"
         />
-        <template v-if="isNameChangeable && !isSocietyDisabled">
+        <template v-else-if="isNameChangeable">
```
severinbeauvais, line 110: This whole section only needs to be displayed if SocietiesInfo wasn't shown, hence the else. And I left the other condition in there.

src/components/new-request/search.vue
```diff
@@ -242,47 +242,49 @@
...
-            :md="(showDesignation || isSelectedXproAndRestorable) ? '8' : '12'"
-          >
-            <NameInput
-              :is-mras-search="(isXproFlow && isMrasJurisdiction && !getHasNoCorpNum)"
-              @emit-corp-num-validity="corpNumValid = $event"
-            />
-          </v-col>
-          <Designation
-            v-if="showDesignation"
+        <template v-else>
```
severinbeauvais, line 251: As above, I'm using an else here to NOT display the following when the dialog is shown.

src/components/new-request/search.vue
severinbeauvais, line 1: The LaunchDarkly config for the FF is now:

image

severinbeauvais: (But, except for Dev, societies are not enabled so it's a moot point except for getting the code right for the future.)

severinbeauvais:
This PR isn't needed. Closing.

PR #751, opened 2024-01-30
19256 Implemented continuation in start + misc. fixes

JazzarKarim:
*Issue #:* /bcgov/entity#19256

*Description of changes:*
- Implemented the start to continuation in
- Fixed old bug where action after logging in wasn't working properly
- Refactored code for better readability and reusability


src/App.vue
```diff
@@ -273,22 +281,25 @@ export default class App extends Mixins(
       sessionStorage.removeItem('NR_DATA')
     }
 
-    // if there is stored legal type for an IA then incorporate/register it now
+    // if there is stored legal type and request action cd, try to continue
     const legaltype = sessionStorage.getItem('LEGAL_TYPE')
-    if (legaltype && this.isAuthenticated) {
+    const requestActionCd = sessionStorage.getItem('REQUEST_ACTION_CD')
```
JazzarKarim, line 286: This here was an old bug that I fixed on the way.

Basically, when trying to do an action without being logged in, after logging in, it wasn't working at all. The reason is previously, we're using the getters like `this.isNewBusiness` but on load, the request action cd isn't in the store so all the getters are null.

Now, I'm storing the request action cd in session storage when the user tries to do an action and setting that in the store. Now, these getters properly work.

src/App.vue
```diff
@@ -86,6 +86,11 @@
       :dialog="getAmalgamateNowErrorStatus"
       @close="closeAmalgamateNowErrorDialog()"
     />
+    <ContinuationInErrorDialog
```
JazzarKarim, line 89: New error dialog to show when continuation in action fails.

src/components/existing-request/nr-approved-gray-box.vue
```diff
@@ -78,6 +78,34 @@
         </v-btn>
       </div>
 
+      <div
+        v-else-if="showContinueInNowButton"
+        class="d-flex justify-center my-1"
+      >
+        <v-btn
+          v-if="showOpenExternalIcon"
```
JazzarKarim, line 86: Go to Colin if FF doesn't support continuation in for this entity. Otherwise, we'll show the "Continue In Now" button.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -194,101 +210,64 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
   }
 
   /**
-   * Handle "Incorporate Now" button.
+   * Handle the action buttons (numbered selection).
```
JazzarKarim, line 213: I took the liberty of refactoring all the following blocks of code here.

Previously, we had 2 functions for each entity type. It was OK at first when we had only incorporation and registration. But the more stuff we add, the more these functions are increasing.

We went from 2 to 4 when we added amalgamations. And If I added here continuation in, it'll be 6 functions doing something similar. And that number will continue on increasing. It's a pain adding more stuff.

So, I refactored to have only 2 functions that deal with everything. Now, adding more types/filings is by simply adding a conditional and what that type is going to have.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-751-wmrpi3ru.web.app

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.3.12",
+  "version": "5.3.13",
```
severinbeauvais, line 3: This is more than a patch. Take some credit -- go to 5.4.0 :)

src/App.vue
severinbeauvais: These fields are put in session storage because there might be a redirect between now and when the fields are needed?

src/App.vue
severinbeauvais: With 3 similar dialogs now, would it be easy(ish) to combine them?

src/interfaces/new-request-interface.ts
```diff
@@ -17,6 +17,7 @@ export interface NewRequestIF {
   assumedNameOriginal: string
   conditionsModalVisible: boolean
   conflictId: string
+  continuationInError: boolean
```
severinbeauvais, line 20: In an alternate universe, we'd have named this `errorContinuationIn` (and the others similarly) to keep them together when sorted alphabetically.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -141,7 +145,7 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
     let businessRequest = {} as BusinessRequest
     const nrNumber = nr.nrNum
     if (!this.isFirm(nr)) {
```
severinbeauvais, line 147: I recommend you refactor this by moving the firm case first and the others into the else. That keeps the conditional statement and its code block close together for easier understanding.

Alternately, you could leave this line alone and change line 186 to `if (this.isFirm(nr))`.

severinbeauvais: Alternately you could leave this alone 😆

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: I totally agree with you 👍

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -141,7 +145,7 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
...
-      if (nr.request_action_cd === 'AML') {
+      if (nr.request_action_cd === NrRequestActionCodes.AMALGAMATE) {
```
severinbeauvais, line 148: Could you use the store getters here instead?

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -194,101 +210,64 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
...
-      const accountId = +JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id || 0
-      const businessId = await this.createBusinessAA(accountId, legalType)
+      const businessId = await this.createNumberedBusiness(accountId, legalType)
       this.goToEntityDashboard(businessId)
       return
     } catch (error) {
       this.$root.$emit('showSpinner', false)
-      this.setAmalgamateNowErrorStatus(true)
-      throw new Error('Unable to Amalgamate Now ' + error)
+      if (this.isAmalgamation) {
```
severinbeauvais, line 228: So the nr is in the store now, and the getter works?

src/App.vue
JazzarKarim: Yes, exactly. When we get back after logging, these are used to know exactly what the user was trying to do (what action and what was the legal etc..)

src/components/dialogs/continuation-in-error.vue
```diff
@@ -0,0 +1,96 @@
...
+  >
+    <v-card>
+      <v-card-title id="dialog-title">
+        Unable to Continue in
+      </v-card-title>
+
+      <v-card-text id="dialog-text">
+        <!-- display message -->
+        <div class="general-error">
+          <p>Unable to continue in. Please cancel or try again.</p>
```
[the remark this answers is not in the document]
JazzarKarim: Agreed. Flipped them over Sev 💪

src/interfaces/new-request-interface.ts
JazzarKarim: Warped us into that alternate universe Sev 😎 Made this change.

src/mixins/nr-affiliation-mixin.ts
JazzarKarim: Added the change to line 186.

src/components/dialogs/societies-info-dialog.vue
```diff
@@ -1,42 +1,50 @@
 <template>
-    <v-dialog v-model="showModal" width="45rem" persistent content-class="custom-dialog">
```
JazzarKarim, line 2: These all are lint changes. Didn't touch this file myself.

src/components/new-request/business-lookup.vue
```diff
@@ -5,13 +5,13 @@
     id="business-lookup"
   >
     <v-autocomplete
+      v-model:search-input="searchField"
```
JazzarKarim, line 8: `lint` change.

src/components/new-request/search-components/xpro-federal-bullets.vue
```diff
@@ -20,14 +20,20 @@
 
       <li v-else-if="isRestoration && isSelectedXproAndRestorable">
         To reinstate your business, complete
-        <a :href="fullReinstatementFormLink" target="_blank">
+        <a
```
JazzarKarim, line 23: `lint` changes. Didn't touch it myself.

src/components/new-request/search.vue
```diff
@@ -242,7 +242,7 @@
 
       <!-- Restoration / Reinstatement flow -->
       <template v-else-if="isRestoration">
-        <BusinessLookupFetch ref="MyBusinessLookup"/>
+        <BusinessLookupFetch ref="MyBusinessLookup" />
```
JazzarKarim, line 245: `lint` change.

src/components/new-request/business-lookup.vue
severinbeauvais: The new code looks like the Vue3-compatible change. Does it still work in Vue2?

src/interfaces/new-request-interface.ts
severinbeauvais: Haha, nice 🚀

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Yeah, this is cleaner, given the number of if-then-else and curly brackets 👍

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -231,10 +234,11 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
       } else if (this.isContinuationIn) {
         this.setContinuationInErrorStatus(true)
         throw new Error('Unable to Continue In Now ' + error)
-      } else {
+      } else if (this.isNewBusiness) {
         this.setIncorporateNowErrorStatus(true)
         throw new Error('Unable to Incorporate Now ' + error)
       }
+      throw new Error('Unable to do action ' + error)
```
severinbeauvais, line 241: I would change the text to "Error in actionNumberedEntity() = " to make it easier to trouble-shoot later if this case ever happens.

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -264,7 +268,7 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
     } else if (this.isContinuationIn) {
       businessRequest.filing.header.name = FilingTypes.CONTINUATION_IN
       businessRequest.filing.continuationIn = { nameRequest: { legalType } }
-    } else {
+    } else if (this.isNewBusiness) {
       businessRequest.filing.header.name = FilingTypes.INCORPORATION_APPLICATION
       businessRequest.filing.incorporationApplication = { nameRequest: { legalType } }
     }
```
severinbeauvais, line 274: How about an else case here, "Invalid request type" or similar?

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Also, please add a `=` or `-` between the text and the error on lines 233, 236, 239 and 241.

src/components/new-request/business-lookup.vue
JazzarKarim: Thank you so much for catching this Sev. You are absolutely right. It doesn't work. I reverted this change.

src/mixins/nr-affiliation-mixin.ts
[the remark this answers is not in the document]
JazzarKarim: This makes a lot of sense.

I've deleted the error throw on lines 277-278 and instead, I'm throwing now an error in the else case as per your other comment.

For lines 277-278. that's being taken care of in the other function as you pointed out Sev. So, there's no need for that.

src/mixins/nr-affiliation-mixin.ts
JazzarKarim: Sev, I've tested this, I can use the `this.getNr` store getter here. It'll be `this.getNr.request_action_cd`.

But I've been thinking, since we're using the function and we're taking the nr parameter, how beneficial would it be to use the store getter in this case?

src/mixins/nr-affiliation-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Looks great. I'm only walking through the code in my head but, theoretically, all errors should now bubble up and be displayed nicely by actionNumberedEntity() 👍

src/mixins/nr-affiliation-mixin.ts
severinbeauvais: Poke.

JazzarKarim: Or maybe you're thinking to try removing the parameter altogether and using the store getter instead?

severinbeauvais: I was thinking you'd use isAmalgamation/isContinuationIn/isNewBusiness ...?

JazzarKarim: OH, these doesn't work as these getters are catching what the action item is. They don't actually check what the NR was.

severinbeauvais: Aside from the possible code duplication (2 ways to do the same thing) in this module, it bothers me a bit that, here, you're checking a field in a formal parameter, while in the code below, the getters look at the global state.

But maybe these methods are used in totally different flows with different store states? If so, that's really not clear in here. But how much do we care about this? Should understanding this complexity be a rite of passsage for new developers? Dunno.

JazzarKarim: True, I totally get where you're coming from.

From my understanding, the store getters (isAmalgamation/isContinuationIn/isNewBusiness) are checking when the action is selected from the dropdown menu.

This here on the other hand, is checking when we open an NR to see what was it's action. So, two completely different things.

I agree, this can be a bit misleading as one might think that for example `isAmalgamation` will return true when we open an NR amalgamation but that isn't the case. It will return true when amalgamation is selected in the dropdown menu.

How would you like to proceed Sev? One suggestion can be renaming those getters so that it's clearer on what they're checking for. Thoughts?

severinbeauvais: Just leave it. We've spent enough time on this today.

severinbeauvais: Your new code is consistent with the previous code. That's the most important thing here.

severinbeauvais (approved):
LGTM but please get another couple of devs looking at this.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-751-wmrpi3ru.web.app

leodube-aot (approved):
LGTM

PR #761, opened 2024-04-22
20951 Added support for continuation in entity types

severinbeauvais:
*Issue #:* bcgov/entity#20951

*Description of changes:*
- app version = 5.4.7
- added CBEN entity type
- added string/type conversions to common mixin

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.4.6",
+  "version": "5.4.7",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
--- a/src/App.vue
+++ b/src/App.vue
@@ -176,7 +176,7 @@ export default class App extends Mixins(
   @Getter isAuthenticated!: boolean
   @Getter isRoleStaff!: boolean
   @Getter isMobile!: boolean
-  @Getter isNewBusiness!: boolean
+  // @Getter isNewBusiness!: boolean
 
   // Global actions
   @Action resetAnalyzeName!: ActionBindingIF
--- a/src/list-data/request-action-mapping.ts
+++ b/src/list-data/request-action-mapping.ts
@@ -49,6 +49,8 @@ export const BcMapping: RequestActionMappingI = {
   ],
   // every entity type except Parishes and Private Act
   CHG: EntityTypesBC.filter(ent => ent !== EntityTypes.PAR && ent !== EntityTypes.PA),
+  // when a MVE (continuation in) NR is created, the resultant company will have a
+  // different entity type in LEAR, as per comments below
   MVE: [
     EntityTypes.CR, // will become CorpTypeCd.CONTINUE_IN
     EntityTypes.CC, // will become CorpTypeCd.CCC_CONTINUE_IN
--- a/src/list-data/request-action-mapping.ts
+++ b/src/list-data/request-action-mapping.ts
@@ -50,12 +50,12 @@ export const BcMapping: RequestActionMappingI = {
   // every entity type except Parishes and Private Act
   CHG: EntityTypesBC.filter(ent => ent !== EntityTypes.PAR && ent !== EntityTypes.PA),
   MVE: [
-    EntityTypes.CR,
-    EntityTypes.CC,
+    EntityTypes.CR, // will become CorpTypeCd.CONTINUE_IN
+    EntityTypes.CC, // will become CorpTypeCd.CCC_CONTINUE_IN
     EntityTypes.CP,
-    EntityTypes.UL,
+    EntityTypes.UL, // will become CorpTypeCd.ULC_CONTINUE_IN
     EntityTypes.SO,
-    EntityTypes.BC
+    EntityTypes.BC // will become CorpTypeCd.BEN_CONTINUE_IN
   ]
 }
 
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.4.5",
+  "version": "5.4.6",
```
severinbeauvais, line 3: Eve, do you have an upcoming release for Namerequest UI? Can you please add this ticket to that release? Thanks!

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-761-38upl0oq.web.app

src/App.vue
```diff
@@ -176,7 +176,7 @@ export default class App extends Mixins(
   @Getter isAuthenticated!: boolean
   @Getter isRoleStaff!: boolean
   @Getter isMobile!: boolean
-  @Getter isNewBusiness!: boolean
+  // @Getter isNewBusiness!: boolean
```
severinbeauvais, line 179: This getter is already imported via a mixin so declaring it here displays an error in VS Code. However, I like to keep the declaration here so that it's explicit that this code file uses this getter.

src/mixins/common-mixin.ts
```diff
@@ -40,6 +40,12 @@ export class CommonMixin extends Vue {
       case EntityTypes.SP: return 'BC Sole Proprietorship'
       case EntityTypes.UL: return 'BC Unlimited Liability Company'
 
+      // Continuation In Entity Types:
+      case EntityTypes.C: return 'Continuation In (BC Limited Company)'
+      case EntityTypes.CBEN: return 'Continuation In (Benefit Company)'
+      case EntityTypes.CCC: return 'Continuation In (BC Community Contribution Company)'
+      case EntityTypes.CUL: return 'Continuation In (BC Unlimited Liability Company)'
```
severinbeauvais, line 47: To be honest, I think these cases don't get used, because as far as Namerequest UI is concerned, the subject NR is for a "BC Benefit Company" (for example), not a CBEN. I think only LEAR cares that this is a continued in benefit company.

src/mixins/common-mixin.ts
```diff
@@ -60,9 +66,13 @@ export class CommonMixin extends Vue {
   entityTypeToCorpType (entityType: EntityTypes): CorpTypeCd {
     switch (entityType) {
       case EntityTypes.BC: return CorpTypeCd.BENEFIT_COMPANY
+      case EntityTypes.C: return CorpTypeCd.CONTINUE_IN
+      case EntityTypes.CBEN: return CorpTypeCd.BEN_CONTINUE_IN
       case EntityTypes.CC: return CorpTypeCd.BC_CCC
+      case EntityTypes.CCC: return CorpTypeCd.CCC_CONTINUE_IN
       case EntityTypes.CP: return CorpTypeCd.COOP
       case EntityTypes.CR: return CorpTypeCd.BC_COMPANY
+      case EntityTypes.CUL: return CorpTypeCd.ULC_CONTINUE_IN
```
severinbeauvais, line 75: Same comment as above: I think these (and the 4 below) aren't used, but I added them for completeness.

vysakh-menon-aot (approved):

severinbeauvais:
Here's what a continuation in NR for a BC Benefit Company looks like:

image

Note that the entity type code (that Namerequest UI uses) is BC (Benefit Company), and the request action code is MVE (move aka continue in). NameX API filled in legal type (CBEN) and request type code (BECT):

image

jamespaologarcia (approved):
LGTM

src/list-data/request-action-mapping.ts
```diff
@@ -50,12 +50,12 @@ export const BcMapping: RequestActionMappingI = {
   // every entity type except Parishes and Private Act
   CHG: EntityTypesBC.filter(ent => ent !== EntityTypes.PAR && ent !== EntityTypes.PA),
   MVE: [
-    EntityTypes.CR,
-    EntityTypes.CC,
+    EntityTypes.CR, // will become CorpTypeCd.CONTINUE_IN
```
JazzarKarim, line 53: What do you mean by the comments here Sev?

package.json
eve-git: could you update the version to 5.4.7? I merged my code with version = "5.4.6" just now,

JazzarKarim (approved):
LGTM

package.json
severinbeauvais: Done!

src/list-data/request-action-mapping.ts
JazzarKarim: Ah I see, I forgot about that.

I totally get you. It's just the name request code is a bit funky. I guess it's OK to keep as is. It makes sense now to me but the next person looking at this will probably need to do some digging to understand what's meant by that 😆 Resolving.

severinbeauvais: Well, that's not ideal that a comment raises more questions than no comment. Any ideas how I can make this clearer?

eve-git (approved):
LGTM

src/list-data/request-action-mapping.ts
JazzarKarim: Maybe add a comment on a separate line before MVE: When Continuation In is selected, the following mappings to LEAR are done (something along these lines). Thoughts?

PR #763, opened 2024-04-30
21033 Use different entity types for continuation in NRs

severinbeauvais:
*Issue #:* bcgov/entity#21033

*Description of changes:*
- app version = 5.4.9
- added missing entity type (CS) to enum
- added cont in entity type data
- added cont in types to list of BC types
- updated MVE request action mapping for cont in
- updated getters to return alternate entity type data for cont in vs BC
- added CS to society code checks
- added cont in codes to numbered entity types list

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -72,8 +72,8 @@ importers:
         specifier: ^0.4.2
         version: 0.4.4
       sbc-common-components:
-        specifier: 3.0.8
-        version: 3.0.8
+        specifier: 3.0.12
+        version: 3.0.12
       style-loader:
         specifier: ^1.3.0
         version: 1.3.0(webpack@5.78.0)
@@ -5453,8 +5453,8 @@ packages:
     resolution: {integrity: sha512-5LBh1Tls8c9xgGjw3QrMwETmTMVk0oFgvrFSvWx62llR2hcEInrKNZ2GZCCuuy2lvWrdl5jhbpeqc5hRYKFOcw==}
     engines: {node: '>=10'}
 
-  sbc-common-components@3.0.8:
-    resolution: {integrity: sha512-TjZA8xRxtwrynI/EXLbGKr68+NFBlFPXDs1n7Uc+EaGD+zwmi8flSi6FEQNbrfb01YfVmlZ8GNghU7f1YP41rw==}
+  sbc-common-components@3.0.12:
+    resolution: {integrity: sha512-xi+zRXCyhnlKQgqbDoPd64Mrs/peB3+TwBv8H3VoEAVkVVr+z6mxfn+ej4WIlGv8KPQkL30LGBQRuZ0TG/JwxQ==}
 
   schema-utils@2.7.0:
     resolution: {integrity: sha512-0ilKFI6QQF5nxDZLFn2dMjvc4hjg/Wkg7rHd3jK6/A4a1Hl9VFdQWvgB1UMGoU94pad1P/8N7fMcEnLnSiju8A==}
@@ -12915,7 +12915,7 @@ snapshots:
     dependencies:
       xmlchars: 2.2.0
 
-  sbc-common-components@3.0.8:
+  sbc-common-components@3.0.12:
     dependencies:
       '@mdi/font': 4.9.95
       axios: 0.21.4
--- a/src/enums/entity-types.ts
+++ b/src/enums/entity-types.ts
@@ -11,6 +11,7 @@ export enum EntityTypes {
   CCC = CorpTypeCd.CCC_CONTINUE_IN,
   CP = CorpTypeCd.COOP,
   CR = CorpTypeCd.CORPORATION,
+  CS = CorpTypeCd.CONT_IN_SOCIETY,
   CUL = CorpTypeCd.ULC_CONTINUE_IN,
   DBA = CorpTypeCd.DOING_BUSINESS_AS,
   FI = CorpTypeCd.FINANCIAL,
--- a/src/list-data/entity-type-data.ts
+++ b/src/list-data/entity-type-data.ts
@@ -281,6 +281,177 @@ export const EntityTypesBcData: EntityI[] = [
   }
 ]
 
+/** List of entity types for Continuation In NRs. */
+export const EntityTypesContInData: EntityI[] = [
+  {
+    text: 'Limited Company',
+    value: EntityTypes.C,
+    cat: 'Corporations',
+    blurbs: [
+      `A company that may have one or more people who own shares with some personal responsibility for debt and
+        liabilities.`,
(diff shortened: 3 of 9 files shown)


src/enums/entity-types.ts
```diff
@@ -11,6 +11,7 @@ export enum EntityTypes {
   CCC = CorpTypeCd.CCC_CONTINUE_IN,
   CP = CorpTypeCd.COOP,
   CR = CorpTypeCd.CORPORATION,
+  CS = CorpTypeCd.CONT_IN_SOCIETY,
```
severinbeauvais, line 14: ... Even though we have that other feature flag to redirect users straight to Society Online if they choose a society-anything NR.

src/list-data/entity-type-data.ts
```diff
@@ -281,6 +281,177 @@ export const EntityTypesBcData: EntityI[] = [
   }
 ]
 
+/** List of entity types for Continuation In NRs. */
```
severinbeauvais, line 284: This is almost the same as the EntityTypesBcData, above, except with only the 6 cont in entity types. The category and blurbs are the same.

(Yes, cont in coop is still CP. This is an ongoing discussion and may be fixed in a separate ticket.)

src/list-data/request-action-mapping.ts
```diff
@@ -2,25 +2,31 @@ import { EntityTypes, NrRequestActionCodes } from '@/enums'
...
-  EntityTypes.CC,
-  EntityTypes.SO,
+  EntityTypes.LP,
   EntityTypes.PA,
-  EntityTypes.FI,
-  EntityTypes.PAR
+  EntityTypes.PAR,
+  EntityTypes.SO,
+  EntityTypes.UL
 ]
```
severinbeauvais, line 24: I added 5 cont in types (since there is no cont in type for CP), and sorted this list.

C, CBEN, CCC, CUL, and CS.

src/store/getters.ts
```diff
@@ -496,7 +497,13 @@ export const getEntityTypesBC = (state: StateIF): EntityI[] => {
...
       const output = []
       for (const entity of entities) {
-        const obj = EntityTypesBcData.find(ent => ent.value === entity)
+        let obj = null as EntityI
+        // continuation in uses a different list than the others
+        if (isContinuationIn(state) && isLocationBC(state)) {
+          obj = EntityTypesContInData.find(ent => ent.value === entity)
+        } else {
+          obj = EntityTypesBcData.find(ent => ent.value === entity)
+        }
```
severinbeauvais, line 506: @JazzarKarim , @jamespaologarcia :

I don't love this implementation. An alternative is to keep using the EntityTypesBcData list, and the "non-cont in" entity types (eg, ULC instead of CUL), but then swap the entity types later in the processing. What do you think? Which makes the most sense to you?

Without seriously refactoring a lot of stuff, we're kind of stuck with needing to toggle between "BC" and "cont in" types _somewhere_.

src/store/getters.ts
```diff
@@ -528,10 +535,13 @@ export const getEntityTypesBC = (state: StateIF): EntityI[] => {
...
 
-    return EntityTypesBcData
+    // safe fallback
+    // continuation in uses a different list than the others
+    if (isContinuationIn(state) && isLocationBC(state)) return EntityTypesContInData
+    else return EntityTypesBcData
   } catch (err) {
     console.error('entityTypesBC() =', err) // eslint-disable-line no-console
-    return EntityTypesBcData
+    return []
```
severinbeauvais, line 544: The old code masked any errors here. The new code should be safe and make any errors obvious.

src/store/getters.ts
```diff
@@ -606,18 +616,17 @@ export const getXproRequestTypeCd = (state: StateIF): XproNameType => {
...
-    if (
-      (isContinuationIn(state) || isRestoration(state)) &&
-      isLocationBC(state)
-    ) {
+    if (isContinuationIn(state) && isLocationBC(state)) {
+      // Shortlist order: Limited Company, ULC
+      if (x.value === EntityTypes.CUL) {
+        x.shortlist = true
+        x.rank = 2
+      }
```
severinbeauvais, line 624: Limited Company (CR) is automatically added as the first element in the short list.

src/store/getters.ts
```diff
@@ -606,18 +616,17 @@ export const getXproRequestTypeCd = (state: StateIF): XproNameType => {
...
+    } else if (isRestoration(state) && isLocationBC(state)) {
       // Shortlist order: Limited Company, Cooperative Association
       if (x.value === EntityTypes.CP) {
         x.shortlist = true
         x.rank = 2
-      } else if ([EntityTypes.FR, EntityTypes.GP, EntityTypes.UL].includes(x.value)) {
-        x.shortlist = null
-        x.rank = null
```
severinbeauvais, line 620: This else-if was not needed.

src/store/getters.ts
severinbeauvais: Note that any error here should be a code error that we should catch during testing, so displaying "nothing" is good option.

If this error was a runtime error then it should be handled differently as a regular user might see it.

src/store/getters.ts
JazzarKarim: I'm OK with this as is. I'm thinking that since this is a very special case which I don't think will re-occur, we can keep as is here. 

Now, if we stumble upon something like this in the future, we might want to reconsider but we don't want to have lots of conditions here. Thoughts? I'm OK with the alternative as well.

src/store/getters.ts
JazzarKarim: 👍👍 👍 👍

pnpm-lock.yaml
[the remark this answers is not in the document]
JazzarKarim: This is good to know. Thanks.

JazzarKarim (approved):
LGTM!

src/store/getters.ts
jamespaologarcia: I think the readability makes up for the potential code duplication because it makes it clear right away why you're changing entity types.

severinbeauvais: Great point, James. Thanks.

Remember this when you following this architecture for your next project 😉

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-763-m2wjrmib.web.app

JazzarKarim:
Last commit LGTM Sev 👍

jamespaologarcia (approved):
LGTM

severinbeauvais:
Folks, sorry, but this solution got messy -- it was sending the continuation in codes (eg, "C") instead of the regular codes (eg, "CR") to NameX API, which was crashing (HTTP 500).

So I went with the alternate solution, which turns out to be much simpler. It will be in another PR.

PR #764, opened 2024-05-01
20133 Convert entity types in numbered Continuation In drafts

severinbeauvais:
*Issue #:* bcgov/entity#20133

(I abandoned my previous PR because that solution was fairly messy and still not working correctly 🙄)

*Description of changes:*
- app version = 5.4.9
- added missing entity type (CS) to enum
- added comments to mapping table
- adding missing entity type to common mixin methods
- updated createNumberedBusiness to convert "entity type" to "continuation in entity type"

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -72,8 +72,8 @@ importers:
         specifier: ^0.4.2
         version: 0.4.4
       sbc-common-components:
-        specifier: 3.0.8
-        version: 3.0.8
+        specifier: 3.0.12
+        version: 3.0.12
       style-loader:
         specifier: ^1.3.0
         version: 1.3.0(webpack@5.78.0)
@@ -5453,8 +5453,8 @@ packages:
     resolution: {integrity: sha512-5LBh1Tls8c9xgGjw3QrMwETmTMVk0oFgvrFSvWx62llR2hcEInrKNZ2GZCCuuy2lvWrdl5jhbpeqc5hRYKFOcw==}
     engines: {node: '>=10'}
 
-  sbc-common-components@3.0.8:
-    resolution: {integrity: sha512-TjZA8xRxtwrynI/EXLbGKr68+NFBlFPXDs1n7Uc+EaGD+zwmi8flSi6FEQNbrfb01YfVmlZ8GNghU7f1YP41rw==}
+  sbc-common-components@3.0.12:
+    resolution: {integrity: sha512-xi+zRXCyhnlKQgqbDoPd64Mrs/peB3+TwBv8H3VoEAVkVVr+z6mxfn+ej4WIlGv8KPQkL30LGBQRuZ0TG/JwxQ==}
 
   schema-utils@2.7.0:
     resolution: {integrity: sha512-0ilKFI6QQF5nxDZLFn2dMjvc4hjg/Wkg7rHd3jK6/A4a1Hl9VFdQWvgB1UMGoU94pad1P/8N7fMcEnLnSiju8A==}
@@ -12915,7 +12915,7 @@ snapshots:
     dependencies:
       xmlchars: 2.2.0
 
-  sbc-common-components@3.0.8:
+  sbc-common-components@3.0.12:
     dependencies:
       '@mdi/font': 4.9.95
       axios: 0.21.4
--- a/src/enums/entity-types.ts
+++ b/src/enums/entity-types.ts
@@ -11,6 +11,7 @@ export enum EntityTypes {
   CCC = CorpTypeCd.CCC_CONTINUE_IN,
   CP = CorpTypeCd.COOP,
   CR = CorpTypeCd.CORPORATION,
+  CS = CorpTypeCd.CONT_IN_SOCIETY,
   CUL = CorpTypeCd.ULC_CONTINUE_IN,
   DBA = CorpTypeCd.DOING_BUSINESS_AS,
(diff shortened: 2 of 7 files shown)


pnpm-lock.yaml
```diff
@@ -72,8 +72,8 @@ importers:
         specifier: ^0.4.2
         version: 0.4.4
       sbc-common-components:
-        specifier: 3.0.8
-        version: 3.0.8
+        specifier: 3.0.12
+        version: 3.0.12
```
severinbeauvais, line 76: I'm not sure why the old lock file specified v3.0.8, because the previous package file specified v3.0.12. Anyways, looks correct now.
🤷‍♂️

src/enums/entity-types.ts
```diff
@@ -11,6 +11,7 @@ export enum EntityTypes {
   CCC = CorpTypeCd.CCC_CONTINUE_IN,
   CP = CorpTypeCd.COOP,
   CR = CorpTypeCd.CORPORATION,
+  CS = CorpTypeCd.CONT_IN_SOCIETY,
```
severinbeauvais, line 14: I added this for completeness, since there's a feature flag that redirects users to Societies Online.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-764-hllvd3sa.web.app

src/mixins/common-mixin.ts
```diff
@@ -40,13 +40,14 @@ export class CommonMixin extends Vue {
...
-      case EntityTypes.C: return 'Continuation In (BC Limited Company)'
-      case EntityTypes.CBEN: return 'Continuation In (Benefit Company)'
-      case EntityTypes.CCC: return 'Continuation In (BC Community Contribution Company)'
-      case EntityTypes.CUL: return 'Continuation In (BC Unlimited Liability Company)'
+      // Continuation In entity types:
+      case EntityTypes.C: return 'BC Limited Company (Continuation In)'
+      case EntityTypes.CBEN: return 'Benefit Company (Continuation In)'
+      case EntityTypes.CCC: return 'BC Community Contribution Company (Continuation In)'
+      case EntityTypes.CS: return 'BC Social Enterprise (Continuation In)'
+      case EntityTypes.CUL: return 'BC Unlimited Liability Company (Continuation In)'
```
severinbeauvais, line 48: I added CS for completeness (also in 2 other methods below).

I changed the text because the name is more important than the type, but this really doesn't matter because these strings are not used in Namerequest UI.

Example (where name and type are displayed separately):

image

src/mixins/nr-affiliation-mixin.ts
```diff
@@ -244,19 +244,26 @@ export class NrAffiliationMixin extends Mixins(CommonMixin) {
...
+    if (this.isContinuationIn) {
+      // convert regular legal type to continuation in legal type
+      switch (legalType) {
+        case CorpTypeCd.BC_COMPANY: legalType = CorpTypeCd.CONTINUE_IN; break
+        case CorpTypeCd.BENEFIT_COMPANY: legalType = CorpTypeCd.BEN_CONTINUE_IN; break
+        case CorpTypeCd.BC_CCC: legalType = CorpTypeCd.CCC_CONTINUE_IN; break
+        case CorpTypeCd.BC_ULC_COMPANY: legalType = CorpTypeCd.ULC_CONTINUE_IN; break
+        case CorpTypeCd.SOCIETY: legalType = CorpTypeCd.CONT_IN_SOCIETY; break
+      }
+    }
```
severinbeauvais, line 261: So here's the gist of it. This is right where we create the draft filing in Legal API.

As far as Namerequest UI is concerned, up to this point, the entity type / corp type was CR/BC/CC/UL/SO.

Maybe you're wondering how this works for NRs? It's similar -- Namerequest UI uses CR/BC/CC/UL/CO, but then Namex API returns the new NR with a converted `entityType` (C/CBEN/CCC/CUL/CS).

src/store/getters.ts
```diff
@@ -531,7 +531,7 @@ export const getEntityTypesBC = (state: StateIF): EntityI[] => {
     return EntityTypesBcData
   } catch (err) {
     console.error('entityTypesBC() =', err) // eslint-disable-line no-console
-    return EntityTypesBcData
+    return []
```
severinbeauvais, line 534: I carried over this change from my previous (closed) PR.

The old code masked any errors by returning what looks like legitimate data. I think it's better to return "nothing" as that will stand out during testing and we can fix any problems.

src/store/getters.ts
```diff
@@ -606,11 +606,16 @@ export const getXproRequestTypeCd = (state: StateIF): XproNameType => {
...
-    ) {
+    if (isContinuationIn(state) && isLocationBC(state)) {
+      // Shortlist order: Limited Company, Unlimited Liability Company
+      if (x.value === EntityTypes.UL) {
+        x.shortlist = true
+        x.rank = 2
+      } else if ([EntityTypes.FR, EntityTypes.GP, EntityTypes.UL].includes(x.value)) {
+        x.shortlist = null
+        x.rank = null
+      }
```
severinbeauvais, line 617: Yes, I left the else-if block in here, even though I'm pretty sure it does nothing. This is the "don't make waves" approach 😁

rarmitag (approved):

ketaki-deodhar (approved):
Looks good Sev!

JazzarKarim (approved):
LGTM

PR #766, opened 2024-05-23
21397 - Build in a refetch mechanism for PENDING_PAYMENT

seeker25:
https://github.com/bcgov/entity/issues/21397


src/mixins/payment-mixin.ts
```diff
@@ -536,7 +536,7 @@ export class PaymentMixin extends Mixins(ActionMixin) {
       if (!paymentResponse) throw new Error('Got error from getNameRequestPayment()')
 
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { payment, sbcPayment = { receipts: [], status_code: '' }, statusCode, completionDate } = paymentResponse
+      const { payment, sbcPayment = { receipts: [], status_code: '' }, statusCode, completionDate } = paymentResponse[0]
```
seeker25, line 539: We'll need to discuss this in the meeting, Andriy found this

seeker25:
I just realized, probably need some more thought around the states

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.5.1",
+  "version": "5.5.2",
```
severinbeauvais, line 3: Also commit package-lock.json please.

seeker25: `pnpm-lock.yaml` ahh ok

seeker25: image

src/components/dialogs/payment-complete.vue
```diff
@@ -109,7 +111,20 @@ export default class PaymentCompleteDialog extends Mixins(
   }
 
   async fetchNr (): Promise<void> {
-    const nrData = await NamexServices.getNameRequest(true)
+    let nrData = await NamexServices.getNameRequest(true)
+    if (nrData.state === NrState.PENDING_PAYMENT && this.incompletePaymentRetries > 0) {
```
seeker25, line 115: I haven't tested this on a true `PENDING_PAYMENT` will try it now, perhaps there are additional conditions we should check here?

src/mixins/payment-mixin.ts
severinbeauvais: The old and new code probably evaluate the same (thanks, Javascript).

package.json
severinbeauvais: Oh yeah. Please and thanks!

src/mixins/payment-mixin.ts
seeker25: Yup, I think we'll probably have to go through all the namex routes and do a side by side to figure out which responses changes to an array or not

severinbeauvais: What I mean is that I think `paymentResponse` and `paymentResponse[0]` are equivalent,

src/components/dialogs/payment-complete.vue
severinbeauvais: 🤷‍♂️

seeker25: image

src/mixins/payment-mixin.ts
seeker25: I had no idea

bolyachevets: Screenshot 2024-05-23 at 3 27 19 PM
Screenshot 2024-05-23 at 3 27 14 PM
running namerequest locally against test api looks like values are not initalized

severinbeauvais: I'm pretty sure I saw that somewhere. Maybe the TS -> JS cross-compilation does something different.

So that complicates things if different routes give different payment responses...

src/components/dialogs/payment-complete.vue
seeker25: I'll probably let this one sit overnight and think about it

src/mixins/payment-mixin.ts
severinbeauvais: Darn, I could have sworn I saw this behaviour before.

Thanks for the correction, Andrew.

bolyachevets: Severin, do you agree this means there may be a bug in namerequest right now? Not sure how these variables are used currently

severinbeauvais: Yes. If payment response is an array, and the Namerequest UI code tries to spread the array instead of the first element, that's a bug.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-766-mr5z71do.web.app

src/mixins/payment-mixin.ts
seeker25, line 539: I think adding this, makes this show again:

image

bolyachevets: maybe changing this to:

```
      let resp = paymentResponse
      if (resp instanceof Array) {
        resp = resp[0]
      }

      // eslint-disable-next-line @typescript-eslint/no-unused-vars
      const { payment, sbcPayment = { receipts: [], status_code: '' }, statusCode, completionDate } = resp
```
will be safer, in case there is a None value

seeker25: I've removed this section of code, I doubt it was called. I checked on TEST it wasn't called.

seeker25: already a check above it @bolyachevets       if (!paymentResponse) throw new Error('Got error from getNameRequestPayment()')

package.json
severinbeauvais: I'm leaving this conversation open until I see the pnpm-lock.yaml file in this PR. Thanks.

src/components/dialogs/payment-complete.vue
```diff
@@ -55,7 +55,7 @@ import { Action, Getter } from 'vuex-class'
 import PaymentConfirm from '@/components/payment/payment-confirm.vue'
 import { NameRequestPayment } from '@/modules/payment/models'
```
severinbeauvais, line 56: Travis, in the GH PR Files changes page, do you see a bunch of GitHub Actions warnings? Could you please clean them up while you're here?

src/mixins/payment-mixin.ts
```diff
@@ -536,7 +536,8 @@ export class PaymentMixin extends Mixins(ActionMixin) {
       if (!paymentResponse) throw new Error('Got error from getNameRequestPayment()')
 
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { payment, sbcPayment = { receipts: [], status_code: '' }, statusCode, completionDate } = paymentResponse
+      // Note this was broken for a while. It was fixed by adding [0] to the response.
```
severinbeauvais, line 539: I think we don't need this comment. However, a comment indicating that paymentResponse may be an object or an array or null might be useful (if true).

src/mixins/payment-mixin.ts
seeker25: Too risky, put this this back to what it was, which was unfortunately broken

src/mixins/payment-mixin.ts
seeker25: I've changed it back since, it can affect too many pathways

package.json
seeker25: I don't think it needs an update, let me check the other PRs?

seeker25: Yeah it doesn't need an update? Looking from your example here:

https://github.com/bcgov/namerequest/pull/764

severinbeauvais: I see that you're right! I thought the app version was in there but I guess it's not. Resolving.

src/components/dialogs/payment-complete.vue
seeker25: Yup

severinbeauvais (approved):

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-766-mr5z71do.web.app

PR #785, opened 2024-12-12
fix: additional address use maxCharRules to prevent errors when empty

EPortman:
*Issue #:* https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/24605

*Description of changes:*
The `maxCharRules` was causing problems when the additional address field was empty. This was because the check was attempting to get the length.
```
 maxCharRules = [
    v => (v.length <= 50) || 'Cannot exceed 50 characters'
  ]
```
This means for the `additionalAddressField` - which had this rule attached, would give the following error when there was nothing in this field preventing the user from moving on when editing an NR. 
image

This PR changes this rule to be as follows:
```
  maxCharRules = [
    v => (!v || v.length <= 50) || 'Cannot exceed 50 characters'
  ]
```
This checks that the value is either falsy (when nothing is there) or that it has less than 50 characters. 

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.5.13",
+  "version": "5.5.14",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",
--- a/app/src/components/common/applicant-info-1.vue
+++ b/app/src/components/common/applicant-info-1.vue
@@ -37,7 +37,7 @@
               <v-text-field
                 id="firstname"
                 :messages="messages['firstName']"
-                :rules="firstMiddleNameRules"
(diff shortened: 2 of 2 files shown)


EPortman:
To test this:

First in Dev environment bring up this NR and try to edit it and observe the error:
**NR Number:** NR 3718858
**Phone number:** 123

Then use the /gcbrun build and do the same thing. It should now work with no errors in the console.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-785-c4we7061.web.app

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-785-c4we7061.web.app

shaangill025 (approved):

PR #789, opened 2024-12-24
25067 Removed unneeded storage sync check

severinbeauvais:
*Issue #:* /bcgov/entity#25067

*Description of changes:*
- updated Status API URL in example env file
- app version = 5.5.20
- updated updateUser error text
- check SESSION_SYNCED before fetching user info
- misc cleanup


app/.env.example
```diff
@@ -18,7 +18,7 @@ VUE_APP_LEGAL_API_URL="https://legal-api-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_LEGAL_API_VERSION_2="/api/v2"
 VUE_APP_NAMEX_API_URL="https://namex-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_NAMEX_API_VERSION="/api/v1"
-VUE_APP_STATUS_API_URL="https://status-api-dev.apps.silver.devops.gov.bc.ca"
+VUE_APP_STATUS_API_URL="https://status-api-dev.apps.gold.devops.gov.bc.ca"
```
severinbeauvais, line 21: Note to developers: please update your `.env` file if needed.

app/src/mixins/update-user-mixin.ts
```diff
@@ -16,7 +16,7 @@ export class UpdateUserMixin extends Vue {
       }
     } catch (err) {
       // just log the error -- no need to halt app
-      console.log('Launch Darkly update error =', err) // eslint-disable-line no-console
+      console.log('Error updating user =', err) // eslint-disable-line no-console
```
severinbeauvais, line 19: The error could be in `fetchUserInfo` or `updateLaunchDarkly` so I generalized this message.

app/src/services/auth-services.ts
```diff
@@ -53,24 +53,19 @@ export default class AuthServices {
    * Throws on error.
    */
   static async fetchUserInfo (): Promise<any> {
-    // PREVENT_STORAGE_SYNC flag is set during signout in sbc common components
-    // when session is state cleared, as a mechanism to prevent returning async
-    // functions to set tokens
-    if (sessionStorage.getItem('PREVENT_STORAGE_SYNC')) return null
```
severinbeauvais, line 59: Code seems to work without console messages after logging in, logging out, logging in again, and logging out again.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-789-bt6gljou.web.app

JazzarKarim (approved):

app/src/services/auth-services.ts
bolyachevets: I tried to reproduce the original issue locally, and could not as well. The check was put in place to take advantage of the flag in the logout function: https://github.com/bcgov/sbc-common-components/blob/d640dc75ea51add2d611a30259d15d93d8654381/vue/sbc-common-components/src/services/keycloak.services.ts#L112 -- this would guard against race conditions in async contexts, It might be possible to reproduce locally if a delay is added to the logout function, e.g. smth along the line:

```
async logout(redirectUrl?: string) {
  let token = ConfigHelper.getFromSession(SessionStorageKeys.KeyCloakToken) || undefined;
  if (token) {
    this.kc = Keycloak(ConfigHelper.getKeycloakConfigUrl());
    let kcOptions: KeycloakInitOptions = {
    ...
    };
    const siteminderLogoutUrl = ConfigHelper.getFromSession(SessionStorageKeys.SiteminderLogoutUrl) || undefined;
    
    // Simulate delay before clearing session
    setTimeout(() => {
      ConfigHelper.clearSession();  // Clear session after delay
      ConfigHelper.addToSession(SessionStorageKeys.PreventStorageSync, true);
    }, 1000); // Delay in milliseconds
    
    return new Promise<void>((resolve, reject) => {
      this.kc && this.kc.init(kcOptions)
        .then(authenticated => {
          ...
        });
    });
  }
}

```

severinbeauvais: Thanks for looking into this, Andrew.

Were you able to demonstrate that the original issue still exists?

Is there any code that removed the session storage variable, PREVENT_STORAGE_SYNC?

Why block the call to fetch user info? Is there anything else that could be done instead, so that the user info feature keeps working correctly? (See ticket for bug info.)

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-789-x6o9lylg.web.app

app/src/services/auth-services.ts
bolyachevets: I did not verify whether the issue would reoccur. Currently there is no issue - no sentry errors. I looked at the ticket you are working on right now, and I wonder if PREVENT_STORAGE_SYNC value in session storage is True would that not mean that the user is logged out, in which case why would we want to still keep track of user's profile? I don't have any suggestions about how to make this better, you probably understand code at the moment better than me (as I worked on a handful of tickets for namerequest back when i was on names team). i also heard that namerequest will be rewritten in the near future.

severinbeauvais: Ah! That would make sense, but it's not working that way currently:

image

I suppose I could check whether SESSION_SYNCED is True, and if so then disregard PREVENT_STORAGE_SYNC ...?

severinbeauvais: @bolyachevets , should PREVENT_STORAGE_SYNC be False (or deleted) if the user logs in again? Maybe this is another bug, that I don't mind reporting if you confirm that's what's supposed to happen.

In the meantime, should I:
- ignore the value of PREVENT_STORAGE_SYNC ? or,
- ignore the value of PREVENT_STORAGE_SYNC if SESSION_SYNCED is True?

severinbeauvais: It looks like PREVENT_STORAGE_SYNC is set to True if the user logs out, and is never cleared (or reset to False) if the user logs in again. I believe this is incorrect.

I have another solution. See following conversation. Resolving this one.

app/src/services/auth-services.ts
```diff
@@ -53,24 +53,22 @@ export default class AuthServices {
...
-    const url = `${this.authApiUrl}/users/@me`
+    // don't fetch user info if session isn't synced (ie, user is not logged in)
+    // to pre-empt a console error and Sentry log
+    if (sessionStorage.getItem('SESSION_SYNCED') !== 'true') return null
```
severinbeauvais, line 58: I'm checking this other session variable, same as in auth-web, ie.

image

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-789-x6o9lylg.web.app

PR #802, opened 2025-03-03
25546 - Remove numbered option when restoration coop

ArwenQin:
*Issue #:* /bcgov/entity#25546

*Description of changes:*
- When restore a Coop, remove the radio button for selecting company type (only allow named)

### DIFF
--- a/app/src/components/new-request/search.vue
+++ b/app/src/components/new-request/search.vue
@@ -248,7 +248,6 @@
           :type="'restore'"
           :showDialog="showSocietiesInfo"
         />
-        <CompanyType v-if="getSearchBusiness && isBcRestorable" />
         <Jurisdiction
           v-if="isSelectedXproAndRestorable"
           cols="12"
--- a/app/src/components/new-request/search.vue
+++ b/app/src/components/new-request/search.vue
@@ -248,6 +248,7 @@
           :type="'restore'"
           :showDialog="showSocietiesInfo"
         />
+        <CompanyType v-if="getSearchBusiness && isBcRestorable && isNumberedEntityType" />
         <Jurisdiction
           v-if="isSelectedXproAndRestorable"
           cols="12"
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "name-request",
-  "version": "5.5.26",
+  "version": "5.5.27",
   "private": true,
   "appName": "Name Request UI",
   "sbcName": "SBC Common Components",


ArwenQin:
After updates:
image

app/src/components/new-request/search.vue
```diff
@@ -248,7 +248,6 @@
           :type="'restore'"
           :showDialog="showSocietiesInfo"
         />
-        <CompanyType v-if="getSearchBusiness && isBcRestorable" />
```
severinbeauvais, line 251: Won't this break the flow for non-coop companies?

severinbeauvais: Look at the code for a new BC business flow (ie, isNumberedEntityType).

ArwenQin: Thanks, just updated, added a check for isNumberedEntityType when show the CompanyType

app/src/components/new-request/search.vue
```diff
@@ -248,7 +248,7 @@
...
+        <CompanyType v-if="getSearchBusiness && isBcRestorable && isNumberedEntityType" />
         <Jurisdiction
```
ArwenQin, line 252: added a check for isNumberedEntityType when show the CompanyType

severinbeauvais: Can you please document some good tests to do for this (in the ticket, for QA), and/or show some screenshots here to validate that this flow works for all business types? Thanks!

ArwenQin:
Test:
- COOP: CP0001040, CP0001042
image


- BC Limited: BC0887683
image


- BC CCC: BC0887687
image


- BC Benefit: BC0883648
image


- BC Unlimited: BC0870872
image



- Continued Limited: C9900818
image



- Continued Unlimited: C9900868
image


- Continued CCC: C9900822
image


- BC Cooperative Association: CP1002469
image

app/src/components/new-request/search.vue
ArwenQin: Sure, just added below

severinbeauvais (approved):

ArwenQin:
@severinbeauvais could you please merge it, thank you

severinbeauvais:
> @severinbeauvais could you please merge it, thank you

Please wait for at least one more review before we merge this PR.

Once merged and you've tested it in Dev, please let Names Team (Omid) know to include your ticket in their next release.

ArwenQin:
Can I get more reviewers take a look please? Thank you

EPortman (approved):

ArwenQin:
Hi Severin, @severinbeauvais 
Do you know if there is a specific reason, why EntityTypes.SO is not included in the special cases?
SO is restorable
image
and SO does not support the numbered company option
image

but SO are not handled here as a special case
image

image
I think we can remove the special cases "this.isCooperative || this.isCreditUnion" and use !this.isNumberedEntityType to include all business types that can't be numbered.
I'm not sure if there is a reason why SO is not included

severinbeauvais:
> I'm not sure if there is a reason why SO is not included

We don't handle SO (societies) at all in Namerequest. The user should not be able to get to the point where they are able to select a NR type for a society, so that's probably why it was missed.

I recommend against changing any code that you don't need to, without a good understanding/analysis of what it's doing and the requirements around it.

ArwenQin:
@severinbeauvais 

> > I'm not sure if there is a reason why SO is not included
> 
> We don't handle SO (societies) at all in Namerequest. The user should not be able to get to the point where they are able to select a NR type for a society, so that's probably why it was missed.
> 
> I recommend against changing any code that you don't need to, without a good understanding/analysis of what it's doing and the requirements around it.

Thank you! This PR is ready to merge, could you please help, thanks

ArwenQin:
@ozamani9gh 
Hi Omid, could you please include this ticket in Names next release?
Thank you!

severinbeauvais:
Just a quick note that Namerequest UI in Dev is pointing to a feature branch, not main branch, so you might not see this fix in Dev right away. (Might have to keep ticket in Review pipeline for a bit.)

ArwenQin:
> Just a quick note that Namerequest UI in Dev is pointing to a feature branch, not main branch, so you might not see this fix in Dev right away. (Might have to keep ticket in Review pipeline for a bit.)

Somehow I saw the new changes in dev now
image

EPortman:
Hi yeah, its the API that name request is pointing to that has changed. Not the UI.

PR #804, opened 2025-03-07
26407 fixed URL not fetched in session storage

severinbeauvais:
*Issue #:* bcgov/entity#26407

*Description of changes:*
- app version = 5.5.28
- updated Auth API URL and Namex API URL in example env file
- updated Auth Services to fetch URL when needed (not when class is used which is before session storage is populated)

### DIFF
--- a/app/src/services/auth-services.ts
+++ b/app/src/services/auth-services.ts
@@ -6,7 +6,9 @@ const axios = AddAxiosInterceptors(Axios.create())
 
 export default class AuthServices {
   /** The Auth API URL. */
-  static authApiUrl = sessionStorage.getItem('AUTH_API_URL')
+  static get authApiUrl (): string {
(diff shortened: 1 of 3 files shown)


app/src/services/auth-services.ts
```diff
@@ -6,7 +6,9 @@ const axios = AddAxiosInterceptors(Axios.create())
 
 export default class AuthServices {
   /** The Auth API URL. */
-  static authApiUrl = sessionStorage.getItem('AUTH_API_URL')
+  static get authApiUrl (): string {
+    return sessionStorage.getItem('AUTH_API_URL')
+  }
```
severinbeauvais, line 11: I believe the old code assigned the property when this class is instantiated, which is before the session storage variable was populated.

Oddly, the old code works fine locally, so I need to deploy this to verify it.

eve-git (approved):

PR #830, opened 2025-06-17
29012 add ff and support for gcp api url for the legal api

EPortman:
*Issue #:* https://github.com/bcgov/entity/issues/29012

*Description of changes:*
Legal-API is moving to GCP behind an API Gateway (to avoid the `...run.app` domain).

This PR updates Name Request to support the new URL using a feature flag:
 - If `use-business-api-gw-url` is true, the app will use the gateway URL (to GCP) and attach `Account-ID` and `X-Apikey` headers.
 - Otherwise, it will continue to use the OpenShift URL (to OCP) without those headers.

**Note:**
`BusinessServices.createBusiness` is the only place where Legal API is called in Name Request, and it only runs when the user is logged in — so account-id will always be available.

**Potential future improvement:**
It might be cleaner to move the header logic into the `AddAxiosInterceptors` call. However, that interceptor is shared with namex, auth, and reg-search APIs — which are not yet using the gateway, so probably better to do that at a later time. 
 
**Tests**
**CALL BEFORE REDIRECT**
This logic runs during NR affiliation (both named and numbered) when logged in.
Since a redirect to the dashboard follows immediately, I paused execution to verify the API call and headers.
image
image

**REDIRECTION**
Redirection fails in dev because the feature flag is off — so the dashboard still uses the OCP DB, which doesn’t recognize the new GCP business.
This should resolve once the feature flag is enabled on both dashboard and name-request.
image


app/.env.example
```diff
@@ -17,6 +17,9 @@ VUE_APP_AUTH_API_URL="https://auth-api-dev-142173140222.northamerica-northeast1.
 VUE_APP_AUTH_API_VERSION="/api/v1"
 VUE_APP_LEGAL_API_URL="https://legal-api-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_LEGAL_API_VERSION_2="/api/v2"
+VUE_APP_BUSINESS_API_GW_URL="https://test.api.connect.gov.bc.ca/business-dev"
+VUE_APP_BUSINESS_API_VERSION_2="/api/v2"
+VUE_APP_BUSINESS_API_KEY=""
```
EPortman, line 22: ^ Add to your .env file

app/src/plugins/featureFlags.ts
```diff
@@ -19,7 +19,8 @@ const defaultFlagSet: LDFlagSet = {
   'sentry-enable': false, // by default, no sentry logs
   'supported-amalgamation-entities': [],
   'supported-continuation-in-entities': [],
-  'supported-incorporation-registration-entities': []
+  'supported-incorporation-registration-entities': [],
+  'use-business-api-gw-url': false
```
EPortman, line 23: To test this locally:
`use-business-api-gw-url` = true  ----> to use the gateway
`supported-incorporation-registration-entities` = ['BC', 'CC', 'CP', 'CR', 'DBA', 'FR', 'GP', 'UL'] ---> to have incorporate now btn

app/src/plugins/getConfig.ts
```diff
@@ -26,6 +26,9 @@ export async function getConfig (): Promise<EnvConfigI> {
   const legalApiUrl: string = process.env.VUE_APP_LEGAL_API_URL + process.env.VUE_APP_LEGAL_API_VERSION_2
   sessionStorage.setItem('LEGAL_API_URL', legalApiUrl)
 
+  const businessApiGwUrl: string = process.env.VUE_APP_BUSINESS_API_GW_URL + process.env.VUE_APP_LEGAL_API_VERSION_2
+  sessionStorage.setItem('BUSINESS_API_GW_URL', businessApiGwUrl)
```
EPortman, line 30: Note: The API key is NOT stored in local storage.

app/src/services/business-services.ts
```diff
@@ -1,12 +1,18 @@
...
 const axios = AddAxiosInterceptors(Axios.create())
 
 export default class BusinessServices {
   /** Legal API URL. */
-  static legalApiUrl = sessionStorage.getItem('LEGAL_API_URL')
+  static get legalApiUrl (): string {
+    return GetFeatureFlag('use-business-api-gw-url')
+      ? sessionStorage.getItem(SessionStorageKeys.BusinessApiGwUrl)
+      : sessionStorage.getItem(SessionStorageKeys.LegalApiUrl)
+  }
```
EPortman, line 15: Changed to a getter to evaluate the feature flag at run-time.

severinbeauvais:
> **Potential future improvement:** It might be cleaner to move the header logic into the `AddAxiosInterceptors` call. However, that interceptor is shared with namex, auth, and reg-search APIs — which are not yet using the gateway, so probably better to do that at a later time.

Something like this might work.
```
if (request.url?.includes('/auth/')) {
   request.headers.common['X-Apikey'] = import.meta.env.VUE_APP_AUTH_API_KEY
} else if (request.url?.includes('/pay/')) {
   request.headers.common['X-Apikey'] = import.meta.env.VUE_APP_PAY_API_KEY
}
```

There's also a rumour that the API keys work for all APIs.

app/src/services/business-services.ts
```diff
@@ -15,6 +21,13 @@ export default class BusinessServices {
...
     const url = `${BusinessServices.legalApiUrl}/businesses?draft=true`
-    return axios.post(url, businessRequest)
+
+    // Add API gateway-specific headers (in addition to interceptor)
+    const extraHeaders = GetFeatureFlag('use-business-api-gw-url') ? {
+      'Account-Id': String(businessRequest.filing.header.accountId),
+      'X-Apikey': process.env.VUE_APP_BUSINESS_API_KEY || ''
+    } : {}
+
+    return axios.post(url, businessRequest, { headers: extraHeaders })
```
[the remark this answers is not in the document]
severinbeauvais: Great solution!

app/src/plugins/featureFlags.ts
EPortman: I guess if you had the LD key it would work, but I don't have that in my .env file.

app/src/services/business-services.ts
severinbeauvais: If you're not sure or you think this is fragile, I recommend not using a getter.

app/src/plugins/featureFlags.ts
severinbeauvais: BTW, I just created that FF for you in LD. It's currently False in all environments.

image

Let me know if you don't see it in the LD response.

severinbeauvais (approved):

app/src/services/business-services.ts
EPortman: Good point. In this case, the getter only runs during the `BusinessServices.createBusiness` call, which happens after the feature flags and session storage values have been set — so it should be safe.

EPortman:
> > **Potential future improvement:** It might be cleaner to move the header logic into the `AddAxiosInterceptors` call. However, that interceptor is shared with namex, auth, and reg-search APIs — which are not yet using the gateway, so probably better to do that at a later time.
> 
> Something like this might work.
> 
> ```
> if (request.url?.includes('/auth/')) {
>    request.headers.common['X-Apikey'] = import.meta.env.VUE_APP_AUTH_API_KEY
> } else if (request.url?.includes('/pay/')) {
>    request.headers.common['X-Apikey'] = import.meta.env.VUE_APP_PAY_API_KEY
> }
> ```
> 
> There's also a rumour that the API keys work for all APIs.

Yeah I really had to think about where to put the headers. I imagine its only a matter of time until namex api, auth api, and reg-search apis move behind the gateway (im honestly surprised that everything is working in prod using the `run.app` domain name for namex from name-request).

app/src/plugins/featureFlags.ts
EPortman: Thanks for creating that. I have been kind of operating blind in terms of launch darkly 😆

app/src/services/business-services.ts
EPortman: > If you're not sure or you think this is fragile, I recommend not using a getter.

Will look into this a bit more and make a decision before merging. I thought I was sure but have lost a bit of confidence.

severinbeauvais: `createBusiness` is normally called some time after the UI loads, but try the situation where you create a draft (eg, numbered) business without being logged in. It should redirect for login, then redirect back and call createBusiness.

severinbeauvais: And thanks to this thread, I realized that the changes I made in Filings UI and Create UI will pre-empt the use of user targeting because the FF is checked well before the user info is fetched and reported to LD. I may update that.

FYI, @TVWerdal @bobdev-94 @JazzarKarim

severinbeauvais:
> Yeah I really had to think about where to put the headers. I imagine its only a matter of time until namex api, auth api, and reg-search apis move behind the gateway (im honestly surprised that everything is working in prod using the `run.app` domain name for namex from name-request).

That's because Auth API and Pay API aren't in Prod yet, and there are reverse-proxies in place. But yes, ultimately all APIs should be behind the API GW.

JazzarKarim (approved):
Nice!

app/src/plugins/featureFlags.ts
severinbeauvais: > I guess if you had the LD key it would work, but I don't have that in my .env file.

I recomend adding the key to your env file.

app/src/services/business-services.ts
```diff
@@ -1,20 +1,33 @@
...
 
   /**
    * Creates (posts) a draft (temporary) business record.
    * Must be logged in to use this.
    * Throws an exception on error.
    */
   static async createBusiness (businessRequest: BusinessRequest): Promise<any> {
-    const url = `${BusinessServices.legalApiUrl}/businesses?draft=true`
-    return axios.post(url, businessRequest)
+    const url = `${BusinessServices.legalApiUrl()}/businesses?draft=true`
```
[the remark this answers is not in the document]
EPortman: Updated, tested, and pushed. 

Thanks!

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-830-nfwi0r41.web.app

PR #833, opened 2025-06-26
27143-Investigation: NR for a ULC extraprovincial registration

stevenc987:
*Issue #:* /bcgov/entity###

*Description of changes:*
Removed the front end for extraprovincial ULC but keep the backend mapping for future use.

### DIFF
--- a/app/src/store/getters.ts
+++ b/app/src/store/getters.ts
@@ -543,6 +543,7 @@ export const getEntityTypesBC = (state: StateIF): EntityI[] => {
 export const getEntityTypesXPRO = (state: StateIF): EntityI[] => {
   let _entityTypesXproData = EntityTypesXproData
   if (isLocationCA(state)) {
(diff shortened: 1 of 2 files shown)


eve-git (approved):

app/src/store/getters.ts
```diff
@@ -543,6 +543,7 @@ export const getEntityTypesBC = (state: StateIF): EntityI[] => {
 export const getEntityTypesXPRO = (state: StateIF): EntityI[] => {
   let _entityTypesXproData = EntityTypesXproData
   if (isLocationCA(state)) {
+    _entityTypesXproData = _entityTypesXproData.filter(ent => ent.value !== EntityTypes.XUL)
     _entityTypesXproData = _entityTypesXproData.filter(ent => ent.value !== EntityTypes.RLC)
```
severinbeauvais, line 547: Won't like 547 overwrite the value set in line 546?

stevenc987: No, it won't overwrite the value set in line 546

PR #837, opened 2025-07-15
28799 - Disallow Maximus users to perform certain actions

JazzarKarim:
*Issue #:* /bcgov/entity#28799

*Description of changes:*
Maximus Users will now not have certain actions enabled:
- Amalgamate Now buttons
- Begin Continuation buttons

For example:

### DIFF
--- a/app/src/components/new-request/business-lookup.vue
+++ b/app/src/components/new-request/business-lookup.vue
@@ -5,13 +5,13 @@
     id="business-lookup"
   >
     <v-autocomplete
+      v-model:search-input="searchField"
       filled
       no-filter
       :hide-no-data="state != States.NO_RESULTS"
       :items="searchResults"
       :loading="state === States.SEARCHING"
       :name="Math.random()"
-      :search-input.sync="searchField"
       append-icon="mdi-magnify"
       autocomplete="chrome-off"
       autofocus
--- a/app/src/interfaces/state-interface.ts
+++ b/app/src/interfaces/state-interface.ts
@@ -1,10 +1,12 @@
+import { AuthorizedActions } from '@/enums'
 import { NewRequestIF } from '@/interfaces/new-request-interface'
 import { NameCheckModelIF, StaffPaymentIF, RefundParamsIF } from '@/interfaces'
 
 export interface StateModelIF {
   common: {
     currentJsDate: Date,
-    keycloakRoles: Array<string>
+    keycloakRoles: Array<string>,
+    authorizedActions: Array<AuthorizedActions>
   }
   newRequestModel: NewRequestIF
   staffPayment: StaffPaymentIF
--- a/app/src/services/business-services.ts
+++ b/app/src/services/business-services.ts
@@ -29,4 +29,20 @@ export default class BusinessServices {
 
     return axios.post(url, businessRequest, { headers: extraHeaders })
   }
+
+  /**
+   * Get authorized actions from business/legal api.
+   * @returns response object
+   */
+  static async getAuthorizedActions (): Promise<any> {
+    const url = `${this.legalApiUrl()}/permissions`
+
+    // Add API gateway-specific headers (in addition to interceptor)
+    const extraHeaders = GetFeatureFlag('use-business-api-gw-url') ? {
+      'Account-Id': JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id || '',
+      'X-Apikey': process.env.VUE_APP_BUSINESS_API_KEY || ''
+    } : {}
+
+    return axios.get(url, { headers: extraHeaders })
+  }
 }
--- a/app/src/store/actions.ts
+++ b/app/src/store/actions.ts
@@ -1,6 +1,7 @@
 import querystring from 'qs'
 import axios from 'axios'
 import {
+  AuthorizedActions,
   CompanyTypes,
   EntityStates,
   EntityTypes,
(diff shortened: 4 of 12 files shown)


severinbeauvais:
Are these actions also blocked from the approved NR page?

app/src/components/new-request/business-lookup.vue
```diff
@@ -5,13 +5,13 @@
     id="business-lookup"
   >
     <v-autocomplete
+      v-model:search-input="searchField"
```
severinbeauvais, line 8: Unless this works, you'll want to revert this change.

app/src/interfaces/state-interface.ts
```diff
@@ -1,10 +1,12 @@
+import { AuthorizedActions } from '@/enums'
 import { NewRequestIF } from '@/interfaces/new-request-interface'
 import { NameCheckModelIF, StaffPaymentIF, RefundParamsIF } from '@/interfaces'
 
 export interface StateModelIF {
   common: {
     currentJsDate: Date,
-    keycloakRoles: Array<string>
+    keycloakRoles: Array<string>,
+    authorizedActions: Array<AuthorizedActions>
```
severinbeauvais, line 9: Actually, you don't need the commas on lines 7 and 8.

app/src/services/business-services.ts
```diff
@@ -29,4 +29,20 @@ export default class BusinessServices {
...
+  static async getAuthorizedActions (): Promise<any> {
+    const url = `${this.legalApiUrl()}/permissions`
+
+    // Add API gateway-specific headers (in addition to interceptor)
+    const extraHeaders = GetFeatureFlag('use-business-api-gw-url') ? {
+      'Account-Id': JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id || '',
+      'X-Apikey': process.env.VUE_APP_BUSINESS_API_KEY || ''
+    } : {}
+
+    return axios.get(url, { headers: extraHeaders })
```
severinbeauvais, line 46: Looks good.

app/src/store/actions.ts
```diff
@@ -747,6 +749,24 @@ export const setKeycloakRoles = ({ commit }, keycloakRoles: string[]): void => {
   commit('mutateKeycloakRoles', keycloakRoles)
 }
 
+export const setAuthorizedActions = ({ commit }, authorizedActions: AuthorizedActions[]): void => {
+  commit('mutateAuthorizedActions', authorizedActions)
+}
+
+export const fetchAuthorizedActions = async ({ commit, getters }): Promise<void> => {
+  // Only fetch if not already loaded
```
severinbeauvais, line 757: Why not always load these, same as account info or whatever?

JazzarKarim:
> Are these actions also blocked from the approved NR page?

Yes!
image

Same thing for amalgamation.

app/src/components/new-request/business-lookup.vue
JazzarKarim: Thank you for catching this. I think the linter changed this automatically when I ran `pnpm run lint`. Fixed.

severinbeauvais: Right. I have the linter for our UIs "not fix" by default, for this exact reason. Actually, it's because the linter is using Vue3 rules that this is flagged, so that's another way this could be fixed/mitigated.

app/src/store/actions.ts
JazzarKarim: We will always load these but only one time and that's on initial load of the UI.

The idea is I don't want subsequent API calls every time to the permissions endpoint every time the search component is shown/rendered. These permissions don't change so we only need to make that call once.

severinbeauvais: Why not load the permissions in created() like the other tombstone data?

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-837-pifhn8jf.web.app

app/src/store/actions.ts
severinbeauvais: By the way, there was a bug where the account info (org id) was not loaded about half the time. There's another ticket for that, and I can't find the code I'm thinking about. Do you know what I'm talking about? Where you have to wait for CURRENT_ACCOUNT to be populated asynchronously by SbcHeader?

The reason I mention it is that you'll need the account info before fetching the authorized actions.

JazzarKarim: Sorry, I misunderstood that first comment. You are absolutely right, makes much more sense. Fixed.

bcregistry-sre:
Temporary Url for review: https://namerequest-dev--pr-837-pifhn8jf.web.app

app/src/store/actions.ts
JazzarKarim: > By the way, there was a bug where the account info (org id) was not loaded about half the time. There's another ticket for that, and I can't find the code I'm thinking about. Do you know what I'm talking about? Where you have to wait for CURRENT_ACCOUNT to be populated asynchronously by SbcHeader?
> 
> The reason I mention it is that you'll need the account info before fetching the authorized actions.

I'm not familiar with the issue you've mentioned. I did update this however and I've tested all the different scenarios. This should be fixed now.

app/src/App.vue
```diff
@@ -304,6 +305,9 @@ export default class App extends Mixins(
         this.staffPaymentErrorDialog = true
       }
     })
+
+    // fetch the user's authorized actions
+    this.fetchAuthorizedActions()
```
severinbeauvais, line 310: You have this as an action (instead of a services call) but it's still async -- doesn't it need to be awaited somehow? Or is it automatically awaited because it's an action?

severinbeauvais: Are other async actions awaited?

JazzarKarim: Yes, since it's an action:
image

All the other actions do not have await before them. I believe it's automatically awaited.

app/src/store/actions.ts
JazzarKarim: Got it. This shouldn't be an issue as the call to fetch permissions is after the user is loaded. It's the last call in created.

EPortman (approved):

severinbeauvais (approved):

TVWerdal (approved):
I see there are a few lint warnings here about line length and unused vars, but other than that looks good.

meawong (approved):

PR #839, opened 2025-07-28
29230 calls to namex go through api gw

EPortman:
*Issue #:*  https://github.com/bcgov/entity/issues/29230

*Description of changes:*
Calls to Namex API go through GW with a generic ui key. 
With this API key specifically, a JWT bearer token is not required and anonymous users can call the API from namerequest UI (with the generic api key).


eve-git (approved):
LGTM

PR #843, opened 2025-08-26
30381 cleaned up some env keys and fixed breadcrumbs

severinbeauvais:
*Issue #:* bcgov/entity#30381

*Description of changes:*
- deleted obsolete key VUE_APP_BUSINESSES_URL
- app = 5.5.46
- fixed Business Registry URL in Affiliation Error Modal
- fixed Staff Dashboard breadcrumb

Also:
- updated browserslist

### DIFF
--- a/app/.env.example
+++ b/app/.env.example
@@ -3,7 +3,6 @@ VUE_APP_PATH="/"
 #vaults web-url
 VUE_APP_REGISTRY_HOME_URL="https://dev.bcregistry.gov.bc.ca/"
 VUE_APP_AUTH_WEB_URL="https://dev.account.bcregistry.gov.bc.ca/"
-VUE_APP_BUSINESSES_URL="https://dev.account.bcregistry.gov.bc.ca/"
 VUE_APP_CORPORATE_ONLINE_URL="https://www.corporateonline.gov.bc.ca"
 VUE_APP_BUSINESS_DASH_URL="https://dev.business-dashboard.bcregistry.gov.bc.ca/"
 VUE_APP_BUSINESS_REGISTRY_URL="https://dev.business-registry-dashboard.bcregistry.gov.bc.ca/"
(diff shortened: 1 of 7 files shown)


app/.env.example
```diff
@@ -3,7 +3,6 @@ VUE_APP_PATH="/"
 #vaults web-url
 VUE_APP_REGISTRY_HOME_URL="https://dev.bcregistry.gov.bc.ca/"
 VUE_APP_AUTH_WEB_URL="https://dev.account.bcregistry.gov.bc.ca/"
-VUE_APP_BUSINESSES_URL="https://dev.account.bcregistry.gov.bc.ca/"
```
severinbeauvais, line 6: This is the old key for the MBR (which is now BRD and is no longer in Auth Web).

app/src/plugins/getConfig.ts
```diff
@@ -6,17 +6,14 @@ export async function getConfig (): Promise<EnvConfigI> {
   // get config from environment
   const windowLocationOrigin = window.location.origin
 
-  if (!{ appBaseURL } || !windowLocationOrigin) {
```
severinbeauvais, line 9: Fixes this:

image

The new code should be correct if the intent is check whether the string is falsy (empty/null).

image

app/src/resources/breadcrumb-resources.ts
```diff
@@ -23,9 +23,9 @@ export function getRegistryHomeBreadcrumb (): BreadcrumbIF {
 }
 
 export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
-  const businessUrl = sessionStorage.getItem('BUSINESSES_URL')
+  const businessUrl = sessionStorage.getItem('AUTH_WEB_URL')
   return {
     text: 'Staff Dashboard',
-    href: `${businessUrl}staff/${getParams()}`
+    href: `${businessUrl}staff/dashboard/active${getParams()}`
```
severinbeauvais, line 29: This more directly redirects staff to the Staff Dashboard.

eve-git (approved):

PR #872, opened 2026-05-13
Fix/conflict check namex solr api

Oluwapizzy:
Calls /possible-conflict-names on namex-solr-api
replaces broken endpoints on legacy namex


app/src/store/actions.ts
stevenc987, line 1: Can you remove the commented codes instead?

stevenc987: I saw the latest version is 5.9.1 on main branch in package.json. Please confirm that is the latest version you got.

